PR #18, opened 2015-04-16
Provide timestamps with timezone information in ISO format for sessions ...

josschne:
...and acquisitions


gsfr:
I just condensed your code a bit and added some more projection updates.

I generally have a rather dense coding style, and we roughly follow PEP8. Imports are primarily ordered by length, then alphabetically. They are grouped as external, internal, local.

The first line of commit messages should only be something like 50 characters long. Second line blank. Then itemized details, but you won't see many details from me. Vim's syntax highlighting guides you.

Feel free to rebase these commits to update your message. We are generally open to rewriting history on branches, but never on master. At least, we try not to do that.

josschne:
Ok, I've got it rebased.

PR #19, opened 2015-04-21
Fix name-related crash when running in demo mode

josschne:
This code path does not guarantee that identity is set. I was hitting a case where identity was undefined. This is a quick-fix since it only affects demo mode.

### DIFF
--- a/base.py
+++ b/base.py
@@ -72,8 +72,8 @@ class RequestHandler(webapp2.RequestHandler):
                         '_id': self.uid,
                         'email': self.uid,
                         'email_hash': hashlib.md5(self.uid).hexdigest(),
-                        'firstname': identity.get('given_name', 'Firstname'),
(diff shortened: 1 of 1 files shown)


gsfr:
I can't actually see the control flow that would lead to this being a problem, but there are a lot of possible flows. Assuming it is an actual problem, why not just set identity to an empty dict at the top of the function, rather than loosing the ability to pull in the user's real name?

josschne:
The control flow that causes this issue is connecting to API without an Authorization header (demo mode) using a user query-string parameter to specify a user that does not yet exist in the system.
I like the suggestion of trying to retain the first/last names for those who do connect with an Authorization header. I'll shoot another commit up.

gsfr:
Squashed, merged, and reworded to fit into 50 char limit.

PR #20, opened 2015-06-23
Hbm merge attempt #1

kofalt:
First attempt at merging in divergence from `master` and `demo`. Several functions had changes in both branches, sometimes mostly copies of the same code, so this merge desires careful review by @gsfr, @andynemzek, and maybe @josschne, in particular file handling code.

Result appears to work, with two exceptions 

* `/sessions` 404s out. This is expected as that's the current behaviour on master; that bug should be fixed separately.
* I'm getting `Upload failed: Server responded with 0 code` on file upload from FDM, despite the `?fdm=1` flag being present. Not sure how to approach that.

Suggestions welcome, up to and including that this merge should be abandoned & retried by another.

### DIFF
--- a/collections_.py
+++ b/collections_.py
@@ -301,14 +301,16 @@ class CollectionSessions(sessions.Sessions):
                 {'$group': {'_id': '$session'}},
                 ])
         query = {'_id': {'$in': [ar['_id'] for ar in agg_res]}}
-        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1}
-        projection['permissions'] = {'$elemMatch': {'_id': self.uid, 'site': self.source_site}}
-        sessions = list(self.dbc.find(query, projection))
+        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'project': 1, 'group': 1, 'timestamp': 1, 'timezone': 1}
+        sessions = self._get(query, projection, self.request.get('admin').lower() in ('1', 'true'))
         for sess in sessions:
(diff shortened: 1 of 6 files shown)


collections_.py
```diff
@@ -301,14 +301,16 @@ def get(self, cid):
                 {'$group': {'_id': '$session'}},
                 ])
         query = {'_id': {'$in': [ar['_id'] for ar in agg_res]}}
-        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1}
-        projection['permissions'] = {'$elemMatch': {'_id': self.uid, 'site': self.source_site}}
-        sessions = list(self.dbc.find(query, projection))
+        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'project': 1, 'group': 1, 'timestamp': 1, 'timezone': 1}
+        sessions = self._get(query, projection, self.request.get('admin').lower() in ('1', 'true'))
```
gsfr: We actually don't want to use ```ContainerList._get()``` in order to avoid the permissions check, as the user does not actually need permissions on the session in order to get to the acquisitions. We also don't want to expose the project, as done below. This will need to be addressed in FDM.
TODO: specify permission semantics, especially wrt collections.

gsfr:
Thanks, @kofalt. After careful review, I think we should simply drop this PR and the demo branch. The only actually useful thing from here was the reset functionality, which I have just hand-ported to master. Next, I will email examples of how to use the new upload functionality.

kofalt:
Sounds good; closing.

PR #21, opened 2015-07-06
Generate .tar instead of .zip for downloads

josschne:


core.py
```diff
@@ -357,18 +358,23 @@ def append_targets(targets, container, prefix, total_size, total_cnt):
...
-        return z, length
+        s = cStringIO.StringIO()
+        z = tarfile.open(mode="w|", fileobj=s)
+        for filepath, arcpath, size in targets:
+            z.add(filepath, arcname=arcpath, recursive=False)
+            yield s.getvalue()
+            s.truncate(0)
+            s.seek(0)
+        z.close()
+        yield s.getvalue()
```
gsfr: I'm trying to understand this code. Are this line and the next one actually necessary?

josschne: Closing the tarfile generates several empty blocks that are part of the tar file spec. Closing adds them to the stream, and the yield ensures they are written out before we wrap up.

core.py
```diff
@@ -357,18 +358,23 @@ def append_targets(targets, container, prefix, total_size, total_cnt):
...
 
     def _archivestream(self, ticket):
-        length = None # FIXME compute actual length
-        z = zipstream.ZipFile(allowZip64=True)
         targets = ticket['target']
-        for filepath, acrpath, _ in targets:
-            z.write(filepath, acrpath)
-        return z, length
+        s = cStringIO.StringIO()
+        z = tarfile.open(mode="w|", fileobj=s)
```
gsfr: To improve code clarity, you might consider calling "z" something like "archive" and "s" something like "stream".

Also, using the the ```with``` context manager would be nice and eliminate the need for ```z.close()```:
```with tarfile.open(mode="w|", fileobj=s) as archive:```

gsfr: Come to think of it, not sure the context manager works in the iterator scenario.

josschne: I'll try it out. Thanks for the comments.

core.py
gsfr: Of course, that makes sense. After all, tar is meant for tape and you need to get to the end of the tape sector or whatever that's called.

josschne:
This change now requires scitran/scitran#32

kofalt:
What happens without the buffer changes?

josschne:
Without the buffer changes you get this semi-regularly in the logs, along with a partial .tar download:

uwsgi_response_write_body_do(): Connection reset by peer [core/writer.c line 419] during GET /api/download?ticket=cd39d88d-853b-4d1e-8656-2822637e1dea&attach=1 (10.0.2.2)
[03] IOError: write error

kofalt:
Gotcha. Deferring to @gsfr for review, assuming latest commits finish the task.

PR #24, opened 2015-07-16
Multipart uploads

kofalt:
This will allow an API consumer to upload a multipart form to send files.
That is, this will now work:

```html
<form action="" method="post" enctype="multipart/form-data">
  <input type="file" name="file" />
</form>
```

Limitation: due to a limitation in the various python libraries I could find, when receiving a form the API will first buffer the stream to disk, rather than parsing as received. This will result in slower processing, since the file is hashed at the end of the upload.


kofalt:
Update: I tested the tip with a 2 GB and 7 GB file. For both I saw no increased RAM usage in uwsgi.

gsfr:
Needs re-testing for memory consumption and inter-site comms.

core.py
```diff
@@ -285,7 +288,7 @@ def store_file(fd, filename, md5, arcpath, arcname):
             self.abort(404, 'no such ticket')
         arcpath = os.path.join(self.app.config['upload_path'], ticket_id + '.tar')
 
-        if self.request.get('complete').lower() not in ['1', 'true']:
+        if self.request.GET.get('complete', '').lower() not in ('1', 'true'):
```
kofalt: I don't understand this change - is it logically identical? Was it broken before?

kofalt: To clarify, I meant only the `[ ]` -> `( )` change, which was added recently

kofalt:
Confirmed with a 7 GB file that the last two commits do not cause a memory usage explosion, though I imagine depending on the submitted form field order, that may cause extra seeking. Either way, it still works.

Not sure how this could affect inter-site features, since the only user interface that supports inter-site will not be using form uploads. I would expect that if something broke (eg, the GET semantics) it would affect single-site usage equivalently.

core.py
gsfr: In this case. ```[]``` and ```()``` are functionally identical. In Python, ```[]``` denotes a mutable array and ```()``` denotes an immutable tuple. The change simply harmonizes the code with the rest of the repo.

kofalt:
Can you confirm what remains before merging? Ref comments above.

gsfr:
I'll test inter-site asap. It's slightly complicated to set up.

kofalt:
Cool, thanks!

gsfr:
Inter-site tests completed with @lmperry's help. Worked fine. Remote uploads are still using excessive memory, not unique to this branch, though. Will address separately.
@kofalt Ready when you are.

kofalt:
Sounds good; merging.

PR #25, opened 2015-07-16
Jobs API

kofalt:
Currently a first pass at the jobs API. Ping @gsfr (and @josschne if desired).
The mongo docs were a bit confusing to me, so lots of learning here.

Points of interest / requests for assistance:

1. Authentication

1. Thoughts on placement of job generator? Function sig?

1. I assume javadoc is not kosher... how shall I doc fn params?

1. Our spec has us removing `util.create_job` in this change; it's called from `bootstrap.py` and `core.py`, thoughts on migration?

1. Something's fishy with lookups: `ValueError: invalid literal for int() with base 10'` when I try to invoke `Job.put`. Maybe I no longer need `int()`, which was copied from older code? Do we do something funky with IDs here?

1. The HTTP return is mongoish: `"_id": {  "$oid": "55a8171890deaf65ad62ea09" }`, `"modified": {        "$date": 1437061320658 }`. How to fix?

1. My state machine impl is hand-rolled and silly; open to changes.


kofalt:
Among other things, I found that some usages of pymongo are [deprecated](http://api.mongodb.org/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_and_modify), so in addition to writing new jobs code I've tried to use the newer & correct pymongo functions.

gsfr:
I have a local branch that upgrades all deprecated pymongo calls, but at the time we still had many instances running an older and incompatible version of pymongo. Might be time to revisit.

kofalt:
They add still work fine in our scitran-pinned version; I figured the best path forward would be to use the new calls in any new code we write. 

I'd like to tackle a full update separately - IMO, this would include upgrading from mongo 2.6.x to 3.x.

gsfr:
Agreed. Already running 3.x locally, btw.

kofalt:
From our conversation:

1. Authentication - as [auth changes](https://github.com/scitran/api/pull/26) have landed, we can now add `self.superuser_request` checks to every call. This flag is set for drones.

1. Job generator sig - good for now, may simplify later.

1. Javadoc - there's a python docstring mannerism to use

1. Migrating from `util.create_job` - removing it is correct; we have to add logic to file save code that adds a `dirty` boolean set to false.

1. Lookups - Mongo's type system is a superset of JSON's, and defaults to awkward serialization. To pass in a string ID, use `bson.ObjectId(_id)`. To return one, use `str(obj._id)`.

1. Mongoish returns - similarly, there are `util.format_timestamp` and `util.parse_timestamp` functions to combat datestamps.

1. State machine impl - leave for now, possible future discussion.

kofalt:
Each concern above has been addressed by separate commits. 

I'd like to review in this PR only the jobs API features that have already been added, and start a different one for the cron job. As it stands, new files are marked as ready for job creation, and jobs are never enqueued, which I think is an appropriate scope for this branch.

Landing this will be sufficient to start building the engine integration.

gsfr:
We can do a separate PR for the cron job, but we need a fully operational engine before any of this can hit master. Maybe want an engine branch and have a number of PRs going that way.

util.py
```diff
@@ -91,13 +90,26 @@ def commit_file(dbc, _id, datainfo, filepath, data_path):
     if _id is None:
         _id = _update_db(dbc.database, datainfo)
     container_path = os.path.join(data_path, str(_id)[-3:] + '/' + str(_id))
+    target_filepath = container_path + '/' + fileinfo['filename']
     if not os.path.exists(container_path):
         os.makedirs(container_path)
-    r = dbc.update_one({'_id':_id, 'files.filename': fileinfo['filename']}, {'$set': {'files.$': fileinfo}})
-    #TODO figure out if file was actually updated and return that fact
```
kofalt: If I understand this change correctly, I think commit_file will still not return if the file was updated. Are you comfortable with that? I just wanted to check as this TODO was removed.

util.py
```diff
@@ -91,13 +90,26 @@ def commit_file(dbc, _id, datainfo, filepath, data_path):
...
-    if r.matched_count != 1:
-        dbc.update({'_id': _id}, {'$push': {'files': fileinfo}})
-    shutil.move(filepath, container_path + '/' + fileinfo['filename'])
+    container = dbc.find_one_and_update({'_id':_id, 'files.filename': fileinfo['filename']}, {'$set': {'files.$': fileinfo}})
+    if container: # file already exists
+        for f in container['files']:
+            if f['filename'] == fileinfo['filename']:
```
kofalt: Do we want to consider rejecting changes to a file of identical filename? Personally, I would be surprised by silent replacement behaviour (versus an OS file manager which would error).

If not, I feel like we should consider adding a ticket on the user interface warning a user about replacement. The file upload modal could check to see if a given filename already exists and ask the user.

util.py
```diff
@@ -91,13 +90,26 @@ def commit_file(dbc, _id, datainfo, filepath, data_path):
...
+                if identical_content(target_filepath, f['filehash'], filepath, fileinfo['filehash']): # existing file has identical content
+                    log.debug('Dropping    %s (identical)' % filename)
+                    os.remove(filepath)
+                else: # existing file has different content
+                    log.debug('Replacing   %s' % filename)
+                    shutil.move(filepath, target_filepath)
+                    dbc.update_one({'_id':_id, 'files.filename': fileinfo['filename']}, {'$set': {'files.$.dirty': True}})
```
kofalt: If the content of the file is different, shouldn't we mark it as dirty here also?

util.py
```diff
@@ -170,102 +182,22 @@ def _entity_metadata(dataset, properties, metadata={}, parent_key=''):
...
+            zf2_infolist = sorted(zf2.infolist(), key=lambda zi: zi.filename)
+            if zf1.comment != zf2.comment:
+                return False
+            if len(zf1_infolist) != len(zf2_infolist):
+                return False
+            for zii, zij in zip(zf1_infolist, zf2_infolist):
+                if zii.CRC != zij.CRC:
+                    return False
+            else:
+                return True
```
kofalt: I'm very worried about going down the rabbit hole of container-agnostic content comparison that digs into tars & zips to see what's inside.

Our execution library has nascent consistent-filesystem-hashing functionality that tries to formalize what this function attempts, and it turns out to be a deceptively-complex problem requiring a lot of thought, edge case handling, and test coverage.

I would much rather the api not open up filesystem containers until & unless we sit down and design + test that feature wholesale. This would have the advantage of the engine & the server agreeing on how filesystems are described.

kofalt:
Rebased onto master to solve trivial merge conflict.

util.py
kofalt: Notes from our conversation:

This check accommodates an edge case in the reaper, which may on occasion redundantly upload zip files of reaped data. You will add a flag to file uploads (`reaper=true`?) that whitelists this zip-check behaviour.

util.py
kofalt: Notes from our conversation: IIRC we agreed to leave the behaviour for now. We'll handle this on the client (Elisa had thoughts about how).

kofalt:
Two questions remain above, after which I think it's ready to land to a 'dev' or 'flywheel' branch as per our conversation earlier. If desired, feel free to add the zip-toggle flag to this PR also.

I think Github doesn't let PRs change their merge target, so when it's ready we can just close this & merge it to a branch of our choosing.

util.py
gsfr: I believe that to be what we are doing. See end of line 106.

gsfr:
Rebased.
I have added a return value and a force parameter to commit_file(). Reaper uploads always set force to true.

kofalt:
This looks great :+1: 

As per our earlier conversation, we'll keep this off master via a flywheel branch. 
Since github PRs cannot change their branch target, I'll close this & merge it manually.

PR #26, opened 2015-07-17
Discussion: Auth changes

kofalt:
From our notes regarding placeholder device authentication:

* Eliminating the drone port, rootCA keys, and drone keys.
* Automated authorization will be via a single shared secret stored in config.toml.
* Automated calls must set their user agent header to 'SciTran Drone NAME', where NAME can be anything (possibly ‘reaper’ or ‘engine’).
* Automated calls must send the shared secret in the X-SciTran-Auth header.
* Drone listing in the database will be completely ignored and have no effect.
* All drones will operate at the same permission level and have the same capabilities.

This adds the ability to authenticate with a shared secret via the above spec.

Does not use constant-time string compare and is thus vulnerable to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack).

Requires scitran/scitran#40.

### DIFF
--- a/api.py
+++ b/api.py
@@ -1,17 +1,3 @@
-#!/usr/bin/env python
-#
-# @author:  Gunnar Schaefer, Kevin S. Hahn
-
-import logging
-logging.basicConfig(
-        format='%(asctime)s %(name)16.16s:%(levelname)4.4s %(message)s',
-        datefmt='%Y-%m-%d %H:%M:%S',
-        level=logging.DEBUG,
-        )
-log = logging.getLogger('scitran.api')
-logging.getLogger('scitran.data').setLevel(logging.WARNING) # silence scitran.data logging
-logging.getLogger('MARKDOWN').setLevel(logging.WARNING) # silence Markdown library logging
-
 import os
 import json
 import webapp2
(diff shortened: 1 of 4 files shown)


gsfr:
@kofalt Sorry for getting the arg parsing stuff mixed into this. It ended up being a lot deeper than I thought it would. Still, I'm glad I finally got rid of this silly redundancy.

api.wsgi
```diff
@@ -24,85 +28,96 @@ ap.add_argument('--api_uri', help='api uri, with https:// prefix')
...
+elif not api.app.config['api_uri']:
+    centralclient_enabled = False
+    log.warning('api_uri not configured -> SciTran Central functionality disabled')
+if not api.app.config['central_uri']:
+    centralclient_enabled = False
+    log.warning('central_uri not configured -> SciTran Central functionality disabled')
+if not api.app.config['drone_secret']:
+    log.warning('drone_secret not configured -> Drone functionality disabled')
+if not api.app.config['apps_path']:
+    log.warning('apps_path is not defined -> App functionality disabled')
```
kofalt: Hmm, this set of checks makes me want a json schema for the loaded config map :)

Maybe we could define one and add it to `check-version.py` in the infra.
If that sounds like a good idea I'll add a ticket over there about it.

api.wsgi
```diff
@@ -24,85 +28,96 @@ ap.add_argument('--api_uri', help='api uri, with https:// prefix')
...
 ap.add_argument('--central_uri', help='scitran central api', default='https://sdmc.scitran.io/api')
 ap.add_argument('--log_level', help='log level [info]', default='info')
+ap.add_argument('--drone_secret', help='shared drone secret')
+
+if __name__ == '__main__':
+    import paste.httpserver
+    logging.getLogger('paste.httpserver').setLevel(logging.WARNING) # silence paste logging
+
+    ap.add_argument('--host', default='127.0.0.1', help='IP address to bind to [127.0.0.1]')
+    ap.add_argument('--port', default='8080', help='TCP port to listen on [8080]')
```
kofalt: Are these conditionally-added arguments only for launching the app without a wsgi server?

api.py
```diff
@@ -109,53 +95,3 @@ def dispatcher(router, request, response):
...
-        log.warning('apps_path is not defined -> apps functionality disabled')
-    elif not os.path.exists(app.config['apps_path']):
-        os.makedirs(app.config['apps_path'])
-
-    db_client = pymongo.MongoReplicaSetClient(args.db_uri) if 'replicaSet' in args.db_uri else pymongo.MongoClient(args.db_uri)
-    app.db = db_client.get_default_database()
-    app.db.sites.update({'_id': args.site_id}, {'_id': args.site_id, 'name': 'Local'}, upsert=True)
-
-    app.debug = True # send stack trace for uncaught exceptions to client
-    paste.httpserver.serve(app, host=args.host, port=args.port, ssl_pem=args.ssl_cert)
```
kofalt: Is this remove-only diff for `api.py` only remove parsing redundancy with `api.wsgi`? I assume this was because of folks using multiple ways to launch the app?

api.wsgi
```diff
@@ -24,85 +28,96 @@ ap.add_argument('--api_uri', help='api uri, with https:// prefix')
 ap.add_argument('--site_id', help='site ID for Scitran Central [local]', default='local')
 ap.add_argument('--site_name', help='site name', nargs='*', default=['Local'])  # hack for uwsgi --pyargv
 ap.add_argument('--oauth2_id_endpoint', help='OAuth2 provider ID endpoint', default='https://www.googleapis.com/plus/v1/people/me/openIdConnect')
-ap.add_argument('--demo', help='enable automatic user creation', action='store_true', default=False)
 ap.add_argument('--insecure', help='allow user info as urlencoded param', action='store_true', default=False)
 ap.add_argument('--central_uri', help='scitran central api', default='https://sdmc.scitran.io/api')
 ap.add_argument('--log_level', help='log level [info]', default='info')
+ap.add_argument('--drone_secret', help='shared drone secret')
```
kofalt: This change is invalid without a corresponding change in the infrastructure. I've added a commit to the infra PR to update the template & launch correctly.

kofalt:
I rather would have separated the base and not-base changes into separate branches, but since master is pretty quiet it's not worth teasing apart the commit.

I'm ready to merge this if you are, pending a few questions above.

api.wsgi
gsfr: Yup.

api.py
gsfr: Yup.

PR #29, opened 2015-08-14
Proper group and user schema

gsfr:
@mrDarcyMurphy Is this good to go?

@kofalt @linavo @josschne This prevents empty string ```_id``` fields.

Note the new ```/schema/group``` and ```/schema/user``` endpoints.
Both deliver their POST schema by default and can take an optional ```method=put``` query param to deliver the respective PUT schema.


mrDarcyMurphy:
@gsfr https://localhost:8443/api/schema/groups isn't loading for me. Should it?

gsfr:
Try https://localhost:8443/api/schema/group.

gsfr:
I went with singular, since it's really the schema for a single group/user.

mrDarcyMurphy:
:facepalm: that works.

Next: The `detail` in the response for a failed request has poor formatting and can't be used directly. It'd be difficult to parse as well because it's got a bunch of `u`'s in it.

```
{
  "code":400,
  "uid":"darcymurphy@invenshure.com",
  "detail":"u'z' is too short\n\nFailed validating u'minLength' in schema[u'properties'][u'name']:\n    {u'maxLength': 32,\n     u'minLength': 2,\n     u'pattern': u'^[0-9A-Za-z .@_-]*$',\n     u'title': u'Name',\n     u'type': u'string'}\n\nOn instance[u'name']:\n    u'z'"
}
```

gsfr:
I just pushed an attempt at better error messages. Let me know if you'd prefer anything different or additional. [Here](https://python-jsonschema.readthedocs.org/en/latest/errors/) is all the info I could get you.

mrDarcyMurphy:
@gsfr Cool. Here's how it looks to me:

``` javascript
{
  "code":400,
  "uid":"darcymurphy@invenshure.com",
  "detail":{
    "validator_value":2,
    "instance":"a",
    "validator":"minLength",
    "relative_path":["name"]
  }
}
```

From there I know what property failed, what it's value was, and why. That should do it. :+1:

gsfr:
The latest commit requires two simple DB migrations:
```
db.users.update_many({}, {'$set': {'created': datetime.datetime.utcfromtimestamp(0), 'modified': datetime.datetime.utcnow()}})
db.groups.update_many({}, {'$set': {'created': datetime.datetime.utcfromtimestamp(0), 'modified': datetime.datetime.utcnow()}})
```

kofalt:
This looks great. A forward-compatible way to upgrade objects to a new json schema pattern :+1:

PR #31, opened 2015-08-18
Zipfiles

gsfr:
This upgrades our on-disk storage format for DICOMs from .tgz to .zip
@kofalt @lmperry Could you please give this a test run?
scitran/testdata already has zip files on master.
scitran/data has a matching PR and needs to be on the zip_parse branch to test this PR.


gsfr:
So, this didn't actually get merged, but rather screwed up by yours truly.

Issue #27, opened 2015-07-20
Upgrade mongo usage


kofalt:
As mentioned in #25, we're using some pymongo functions which are deprecated; see diff there for more details. Eventually it might be wise to upgrade those.

PR #33, opened 2015-08-24
WIP: Fix pymongo deprecations

Eric89GXL:
Currently the uwsgi log is populated with a few of these warnings:
```
./persistent/logs/uwsgi.log:./util.py:140: DeprecationWarning: find_and_modify is deprecated, use find_one_and_delete, f
ind_one_and_replace, or find_one_and_update instead
```
I don't know much about `pymongo` / MongoDB -- does this seem like the right fix for those?

It seems to work okay with the sample data and have the same behavior when adding MEEG data as `master`, but it would be good if someone with more `pymongo` knowledge could confirm it's correct.

### DIFF
--- a/util.py
+++ b/util.py
@@ -15,6 +15,7 @@ import datetime
 import mimetypes
 import dateutil.parser
 import tempdir as tempfile
+from pymongo.collection import ReturnDocument
 
 import scitran.data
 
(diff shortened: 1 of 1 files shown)


kofalt:
Do you think this will satisfy #27?

Diff LGTM but I'd like @gsfr or someone to confirm sanity check.

Eric89GXL:
Sounds like @gsfr had a local branch with similar changes a month or two ago, maybe we could use that instead of this, too.

kofalt:
¯\_(ツ)_/¯ whatever works!

gsfr:
This is far from complete wrt #27. I have a local stash from May(!) that I'll try to apply in the next couple of days and create a competing :smile: PR.

kofalt:
Should we merge / close this? Is there another imminent PR that will do more than this?

gsfr:
I'll open a more comprehensive PR as soon as #36 lands.

PR #35, opened 2015-08-25
Optional NewRelic support

josschne:
New Relic is conditionally brought in based on whether it is found in import, and fails quietly if not found.

### DIFF
--- a/api.py
+++ b/api.py
@@ -115,6 +115,10 @@ def dispatcher(router, request, response):
         response.write(json.dumps(rv, default=bson.json_util.default))
         response.headers['Content-Type'] = 'application/json; charset=utf-8'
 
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -115,6 +115,10 @@ def dispatcher(router, request, response):
         response.write(json.dumps(rv, default=bson.json_util.default))
         response.headers['Content-Type'] = 'application/json; charset=utf-8'
 
+try:
+    from newrelic import agent
```
gsfr: The ```import newrelic.agent``` style is generally preferable over the ```from newrelic import agent``` style for code readability and traceability reasons.

kofalt:
LGTM

PR #36, opened 2015-08-25
ENH: Refactor module

### LINKED ISSUE #34: Q: module organization
I'm looking through the module, and it looks like the organization has some non-standard conventions, which is making it a little more challenging than usual for me to understand:

* imports are absolute (`import apps`) instead of relative (`from . import apps`) so it's tough to tell what is from the local package vs built-in or external imports
* scripts like `bootstrap.py` are in the same location as source modules -- typically these live in a `bin` directory or so, to make it clear what is part of the module, and what is a callable script

Do these need to be that way? I'm happy to work on reorganizing these aspects of package if it makes sense to do so at some point. I think it would make it easier to maintain and for other devs to understand the layout, and in the meantime would teach me a bit about the module. But if this seems overkill, it doesn't need to be done.

Eric89GXL:
Closes #34.

This refactors the module as discussed in the above issue.


api/util.py
```diff
@@ -80,6 +83,17 @@ def commit_file(dbc, _id, datainfo, filepath, data_path, force=False):
...
     fileinfo = datainfo['fileinfo']
+    log_path = os.path.join(os.path.split(data_path)[0], 'logs')
+    # XXX Eventually it would be good to find a more principled/constant
+    # way to do this
+    if os.path.isdir(log_path):
+        hdlr = logging.FileHandler(os.path.join(log_path, 'commit_file.log'))
+        fmt = logging.Formatter(fmt=log_fmt, datefmt=datefmt)
+        hdlr.setFormatter(fmt)
+        hdlr.setLevel(logging.DEBUG)
+        log.addHandler(hdlr)
```
Eric89GXL: This is the thing that made debugging much easier, namely logging the `commit_file` calls to `persistent/logs/commit_file.log`. It's a bit hackishly coded right now, I suspect it will be easier to get this right with a single-repo setup.

Eric89GXL:
Requires https://github.com/scitran/scitran/pull/64 for `scitran` to work properly

gsfr:
Thanks, Eric. Very cool.

A couple of things:
- Shouldn't we create the ```scitran.api``` namespace?
- ```bin/centralclient.py``` should be executable, not ```api/centralclient.py```.
- Isn't a ```data``` directory for a single file a bit excessive? I, for one, find it plain annoying.
- Please resolve the merge conflicts.

Eric89GXL:
> Shouldn't we create the scitran.api namespace?

Yeah, but unfortunately I don't think we can right now. IIRC there are a couple of things that need to import from both `scitran.api` and `scitran.data`, but they aren't actually both nested under a single `scitran` folder currently. So anything that tries to import both will only find one of them, and not the other (which it finds would be dependent upon `sys.path` order). But it shouldn't be difficult to nest `api` under `scitran` in the eventual "big merge" situation once this is in place.

> bin/centralclient.py should be executable, not api/centralclient.py.

Whoops! Yeah I `git mv`'ed the latter, and created the former without changing perms. Fixed with latest commit.

> Isn't a data directory for a single file a bit excessive? I, for one, find it plain annoying.

It's overkill for now. I usually do that in case the number of data files grows -- then it's convenient to provide some structure. But for now it's overkill and it can always be created later, so I've removed it.

> Please resolve the merge conflicts.

Done.

I deleted my `persistent` folder and re-reaped some MEEG data, and it still works. So I don't think I broke anything with the changes :)

gsfr:
Cool. Makes sense about the namespace. Unfortunate, though.

If you guys are interested (@Eric89GXL, @kofalt), I'd like to do a Hangout next week to discuss the bigger refactoring plans.

I will be traveling to MSP tomorrow, but should have time to test this from scratch at the airport.

@kofalt Is ```apps.py``` still relevant?

@Eric89GXL Can we now safely rename ```collections_.py``` to ```collections.py```?

Eric89GXL:
Yeah I can do the renaming tomorrow morning.

Feel free to pick some times that work for you all for hangout. My schedule
is pretty flexible, but I'm on eastern time right now so before 2pm pacific
is better for me but not required.

api/apps.py
```diff
@@ -107,8 +104,11 @@ def post(self):
                 jsonschema.validate(app_meta, APP_SCHEMA)
             except (ValueError, jsonschema.ValidationError) as e:
                 self.abort(400, str(e))
-            util.insert_app(self.app.db, app_temp, apps_path, app_meta=app_meta)  # pass meta info, prevent re-reading
+            insert_app(self.app.db, app_temp, apps_path, app_meta=app_meta)  # pass meta info, prevent re-reading
             log.debug('Recieved App: %s' % app_meta.get('_id'))
+        """
+        # XXX util.insert_app doesn't exist...?
+        raise NotImplementedError
```
Eric89GXL: BTW just wanted to point this out -- this method made use of a nonexistent `util.insert_app` function, so I don't think it is ever actually called. Should I just delete it? Right now I hackishly "commented" it out by making the code a big string.

Eric89GXL:
`collections_` moved.

@gsfr don't forget you need scitran/scitran#64 simultaneously when you go to test.

gsfr:
Works!

- Not sure why we now have a Makefile to clean stuff.
- Wondering if ```api.wsgi``` should be in ```bin```. It's basically an executable python file. Is the currently layout typical for WSGI apps?
- I don't think we actually have a need for ```bin/centralclient.py```. We always use UWSGI.
- @kofalt Is ```apps.py``` still relevant? (Restating for completeness, not pressure.)

Eric89GXL:
> Not sure why we now have a Makefile to clean stuff.

The `Makefile` is there because it's nice to be able to do `make clean` to ensure all the `.pyc` files have been cleared out, or their Python3 equivalents if eventually this moves to that platform. I can remove it if it seems unnecessary. Usually other commands end up in there, like `make test` or `make flake` (I guess `make lint` in this case probably), but maybe this will end up being redundant with `live.sh` somehow.

> Wondering if api.wsgi should be in bin. It's basically an executable python file. Is the currently layout typical for WSGI apps?

I have never worked with it, so it doesn't really matter to me. If it's primarily used as an executable, which it sounds like it is, it's probably worth moving.

> I don't think we actually have a need for bin/centralclient.py. We always use UWSGI.

I wasn't sure what that was used for, so I'm happy to remove it. It's just copy-pasted from some code that was already in there, perhaps it was outdated.

gsfr:
>> Not sure why we now have a Makefile to clean stuff.

> The Makefile is there because it's nice to be able to do make clean to ensure all the .pyc files have been cleared out, or their Python3 equivalents if eventually this moves to that platform. I can remove it if it seems unnecessary. Usually other commands end up in there, like make test or make flake (I guess make lint in this case probably), but maybe this will end up being redundant with live.sh somehow.

I'd say we keep it lean until a real need arises.

>> Wondering if api.wsgi should be in bin. It's basically an executable python file. Is the currently layout typical for WSGI apps?

> I have never worked with it, so it doesn't really matter to me. If it's primarily used as an executable, which it sounds like it is, it's probably worth moving.

Agreed. This will very likely need a matching change in scitran/scitran#64.

>> I don't think we actually have a need for bin/centralclient.py. We always use UWSGI.

> I wasn't sure what that was used for, so I'm happy to remove it. It's just copy-pasted from some code that was already in there, perhaps it was outdated.

Yeah, it's legacy. Let's ditch it.

Eric89GXL:
@gsfr comments addressed with changes here and in https://github.com/scitran/scitran/pull/64

Eric89GXL:
I re-tested locally and things still appear to be okay

api/util.py
gsfr: I don't like this at all. Let's get rid of it. How is this better than a ```tail -f | grep``` of the UWSGI log, anyway?

Also, the logging module has ways of selectively using handlers for certain messages. Not that I'm saying we should do that, though.

gsfr:
I also just retested. Everything still works. I'll be adding a few line comments shortly.

api/util.py
```diff
@@ -15,9 +12,15 @@
...
 import tempdir as tempfile
+import logging
 
 import scitran.data
 
+log_fmt = '%(asctime)s %(message)s'
+datefmt = '%Y-%m-%d %H:%M:%S'
+log = logging.getLogger('scitran.api')
+logging.basicConfig(format=log_fmt, datefmt=datefmt, level=logging.INFO)
+
```
gsfr: Ideally, we'd only have one definition of our logging style. We currently have this one and one in ```api.wsgi```. ```util.py``` is probably the right place, but I prefer the definition we have in ```api.wsgi```.

api/util.py
```diff
@@ -150,6 +168,12 @@ def _update_db(db, datainfo):
             new=True,
             projection=PROJECTION_FIELDS,
             )
+    try:
+        session_label = session['label']
+    except KeyError:  # this can happen if nims_metadata_status == None
+        session_label = '(unknown)'
+    log.info('Using group_id="%s", project_name="%s", and session_label="%s"'
+             % (project['group'], project['name'], session_label))
```
gsfr: Let's try to keep the overall number of log lines per sorted file to a minimum. I'm currently seeing 3 lines per file at ```info``` level, plus one for parsing. One line at ```info``` and one more at ```debug``` seems like a good target. I see the issue of having access to the relevant info at the right time without returning a bunch of basically useless info.

I would play with it myself, but I won't because I can't commit into your PR. If you're going to continue contributing at this rate, it might make sense to make you a commiter for our repos. Thoughts?

Please also try to follow the layout of existing log entires. Note the spacing.

Also, this should be a single line.

api/util.py
Eric89GXL: I agree it's a hack and I don't like it either, but none of these logging
messages got saved currently as far as I could find. The first thing I
checked was uwsgi and didn't see anything, and I think I actually grep
-R'ed to check and didn't find the messages anywhere. I think it's better
to have the hack in place, working but ugly, than not have the output at
all, but I'll get rid of it if you want. Ideally we could do it in a more
principled way, but I don't get how the pieces for together well enough yet
to see what that is...

api/util.py
Eric89GXL: You can always open a PR into my branch with changes, it might work better
than silently committing. I could also give you write permissions to my
fork if you'd prefer to avoid opening PRs.

api/util.py
```diff
@@ -80,6 +85,17 @@ def commit_file(dbc, _id, datainfo, filepath, data_path, force=False):
...
+    # way to do this
+    if os.path.isdir(log_path):
+        hdlr = logging.FileHandler(os.path.join(log_path, 'commit_file.log'))
+        fmt = logging.Formatter(fmt=log_fmt, datefmt=datefmt)
+        hdlr.setFormatter(fmt)
+        hdlr.setLevel(logging.DEBUG)
+        log.addHandler(hdlr)
+    else:
+        hdlr = None
     log.info('Sorting     %s' % filename)
```
gsfr: ```bootstrap.py:87``` is redundant with this. This maybe useful for reaper uploads, though, and we might need an extra line in ```reaper()``` in ```core.py```.

api/util.py
gsfr: During bootstrapping, messages only go to ```stdout```, which you can probably redirect if you really want it in a file. Reaper uploads should surely be logged to ```uwsgi.log```. If they're not, that needs to be fixed. Might be as simple as adjusting the log level down.

api/util.py
gsfr: I'm worried about things getting too confusing with PRs into PRs.

Eric89GXL:
@gsfr comments addressed. You were right about the log, just setting `log.setLevel(logging.INFO)` made the messages appear in the `uwsgi` log.

kofalt:
On review this and scitran/scitran#64 make sense to me, what's the status here?
@gsfr Do we want to do a hangout discussing plans?

I'm ready to sanity check + merge these two PRs.

Eric89GXL:
Ready from my end, but I don't think @gsfr has had a chance to look at the changes I made to address his comments.

gsfr:
Thanks for addressing these final things, @Eric89GXL. I'm good to go on this, bar the one line comment I just made.

@kofalt A final sanity check from scratch would be great.

@kofalt Verdict on relevance of ```apps.py```?

Eric89GXL:
@gsfr done

kofalt:
Nuke apps.py. But I wouldn't make that a requirement of this PR. Eric feel free to delete :)

Eric89GXL:
@kofalt @gsfr done, it required removing the corresponding entry from the routes in `api.py`, I hope that's what you had in mind. If not, I can remove the last commit and it can be done properly in a subsequent PR.

@kofalt did the final sanity check from scratch work for you? It runs fine over here.

gsfr:
I did another test from scratch this morning, including scitran/data#28 and scitran/scitran#64. Everything seems to have worked.

Unfortunately, we have a merge conflict now. Please rebase or merge and then we are finally ready to go.

Eric89GXL:
@gsfr done, I'll re-run it again just to make sure it still works

kofalt:
I'll look at this within the hour :)

Eric89GXL:
Sample data loading and MEEG reaping both appear to work fine over here

kofalt:
Sanity checks passed; merging both PRs.

PR #37, opened 2015-08-25
Fix arbitrary path access

kofalt:
User input was not sanitized, so it was possible to request read or write of any file on the system. I was able to request a specific file be written on a completely separate path from the data folder.

This PR should drop any paths on a filename parameter when up/downloading.


kofalt:
Merging based on discussion with @gsfr.

PR #38, opened 2015-08-28
Expanded upon the Group's regex patterns.

mrDarcyMurphy:
Both `_id` and `name` now require a letter or number at the start and end, and enforce the minimum and maximum length.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -6,14 +6,14 @@
             "_id": {
                 "maxLength": 32,
                 "minLength": 2,
-                "pattern": "^[0-9a-z]{1}[0-9a-z.@_-]{0,30}[0-9a-z]{1}$",
+                "pattern": "^[0-9a-z][0-9a-z.@_-]{0,30}[0-9a-z]$",
                 "title": "ID",
                 "type": "string"
             },
             "name": {
                 "maxLength": 32,
                 "minLength": 2,
-                "pattern": "^[0-9A-Za-z]{1}[0-9A-Za-z .@_-]{0,30}[0-9A-Za-z]{1}$",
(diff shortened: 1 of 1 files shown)


kofalt:
Seems reasonable. Pinging @gsfr for sanity check.

gsfr:
It's all fine, but a couple of points. I'm a big fan of WTO (write things once).
- If I want to change the min or max length, I now have to do it in multiple places and even do math to get it right.
- The mentions of ```{1}``` are redundant.

Hence, I would remove all explicit length declarations from the regex.

kofalt:
@mrDarcyMurphy is this ready for re-review? I see another commit.

mrDarcyMurphy:
@gsfr I can see your reasoning, and I'm a fan of only writing things once as well. In this case it bothers me that the regex is technically incorrect without the length quantifiers. 

I've removed the redundant `{1}`s at least, not sure I'm willing to fight to keep the others. If you still want them gone then let me know and I'll remove them.

mrDarcyMurphy:
@kofalt Yeah. Just waiting on @gsfr's feedback.

gsfr:
My view is that we use the regex for the pattern and we separately impose the length constraints. That way we can be sure that we favor human-readable error messages, such as "not long enough" over the more cryptic "does not meet expected pattern". That said, the python validator I'm using seems to favor length constraints anyway, but I'm guessing that's just an implementation detail.

Anyway, please update or merge as desired.

mrDarcyMurphy:
I'll merge this in with the caveat that if the length in the regex becomes a hassle it should be removed.

PR #40, opened 2015-09-03
Save user's Google avatar if they have one

kofalt:
This PR will upgrade the user's avatar from our default (gravitar) to their Google avatar if that provider is currently in use, and if they have a Google avatar to display.

Eventually we might want to expand this out into a more general support of multiple oAuth providers; this is the bare minimum to get things working today.

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -45,6 +45,36 @@ class RequestHandler(webapp2.RequestHandler):
                     if not self.uid:
                         self.abort(400, 'OAuth2 token resolution did not return email address')
                     self.app.db.authtokens.replace_one({'_id': access_token}, {'uid': self.uid, 'timestamp': datetime.datetime.utcnow()}, upsert=True)
+
+                    # Opportunistically set user's avatar based on their auth provider
+                    # TODO: after api starts reading toml config, switch on
+                    # auth.provider rather than manually comparing endpoint URL.
+                    if self.app.config['oauth2_id_endpoint'] == 'https://www.googleapis.com/plus/v1/people/me/openIdConnect' and identity.get('picture') != None:
+                        avatarURL = identity.get('picture')
+
+                        # NOTE: Google URLs have a size attached. This code removes that parameter.
+                        # One could also set the size explicitly.
+                        # from urllib import urlencode
+                        # from urlparse import urlparse, urlunparse, parse_qs
+                        # u = urlparse(avatarURL)
+                        # query = parse_qs(u.query)
+                        # query.pop('sz', None)
+                        # u = u._replace(query=urlencode(query, True))
+                        # avatarURL = urlunparse(u)
+
+                        result = self.app.db.users.find_one_and_update(
+                            {
+                                'email': self.uid
+                            },
+                            { '$set': {
+                                'avatar': avatarURL,
+                                'modified': datetime.datetime.utcnow()
+                            }}
+                        )
+
+                        if result == None:
+                            log.debug('Could not find user record to update avatar')
+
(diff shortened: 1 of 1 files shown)


gsfr:
I really don't like to have this code in the path of every single request that hits the API.

Also, what if the user wants to keep their gravatar and not have it swapped out underneath them?

I say we drop this until we have proper UI/UX for an editable avatar, which is part of our upcoming design discussion.

kofalt:
Well, this will only run on token refresh, not every request. I actually think this is the exact desired frequency - whenever we refresh to the login provider, we should know if any user info changed.

One optimization would be to only modify the user document if the avatar URL differs from what's stored - as written, the "modified" timestamp will reflect the most recent token refresh. I think that would mean adding `{'avatar': {'$ne': avatarURL}` to the query and removing the None results check.

As to choosing an avatar - yeah, I don't have an answer for that. Since we don't have configurable user profile support, the real question is which should we default to for now - and I'd argue that gravitar is the inferior option. For sites using Google authentication I really wouldn't expect my profile picture to differ from my Google login... and I certainly wouldn't want to sign up for WordPress to get an avatar to render.

We could make this configurable at the site level by adding another configuration key? Dunno, I guess it depends on if any existing since know / care about gravitars. As a practical matter, if the API isn't storing avatar URLs our web app will likely switch to loading the avatar from the hello.js plugin. I'd rather avoid that if possible because the split-brain nature feels wrong to me.

gsfr:
Yes, sorry. Of course, you're right about the code path. GitHub code folding got the better of me.

Editable user profiles are one of the few loose ends that should have been in *hana*. In fact, since we don't have a pressing need for a release right now, I'm hoping they might be in before we actually cut *hana*.

I could imagine a button the user profile that lets the user set the avatar to the one coming from the auth provider. But I could also imagine letting the user paste in any URL they want.

I don't disagree with what you are saying about the Google avatar being expected when using Google auth, although I can't even set an avatar on my crippled Stanford Google account.

Is there a way to pre-generate the Google avatar URL as we do for gravatar right now?

kofalt:
Oops. Crap. Merged this by accident. Stand by.

kofalt:
Force-pushed away my accident; let's continue the conversation here. 

There is an API call we could make to get a user's avatar:

`https://www.googleapis.com/plus/v1/people/{G_USER_ID}?fields=image&key={OUR_API_KEY}`

which returns

```json
{ "image": {
 "url": "https://lh3.googleusercontent.com/...."
}}
```

What I don't have is a way to convert an arbitrary email into a google user ID. I found at least one stack overflow who responded that there's no way to do that. This doesn't surprise me too much, since it'd allow one to collect a full set of google users.

Unless I misunderstand the (complicated) set of google APIs, I think we have to wait until we're actually handed a valid token from a user. This approach also covers the user avatar URL changing - they're very clearly not permalinks.

If we later add profile configuration, it's trivial to re-generate a gravitar URL, and it's trivial to pass a Google token when we have one. Adding feature could be simply saving the preference to the user doc, and updating this code here to check the preference.

gsfr:
Ok, fair enough. Let's get it in.

The mongo query should use ```update_one()```, though, not ```find_one_and_update()```, and we should be querying on ```_id```, not ```email```, as the latter is technically user-defined.

Will push commit momentarily.

gsfr:
Pushed ac8061e2336eaaed7cfecd8623e2e75b371b971e, but GitHub won't update this merged PR anymore.

kofalt:
Commit looks good. I'll open a new PR momentarily with one small change.

PR #42, opened 2015-09-03
Bootstrap

gsfr:
This moves all required bootstrapping to the app startup and makes bootstrapping users, groups, and data optional.

@kofalt Sanity check, please. Matching PR for scitran/scitran about to hit.

### DIFF
--- a/api/util.py
+++ b/api/util.py
@@ -36,7 +36,7 @@ for mt in MIMETYPES:
 
 valid_timezones = pytz.all_timezones
 
-PROJECTION_FIELDS = ['group', 'name', 'timestamp', 'permissions', 'public']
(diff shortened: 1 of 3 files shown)


kofalt:
Looks reasonable. I take it `create_index` is idempotent?

gsfr:
Yup. The DB is unavailable, though, while the indexes are being built.

kofalt:
Hmm, then I guess we'll eventually need some `await_index` equivalent for when we add new indexes. But this works for now.

gsfr:
You can build them in the background, but that doesn't make much sense at application start, unless it really takes an inordinate amount of time.

kofalt:
LGTM and sister PR is ready; merging.

PR #44, opened 2015-09-08
Add retries to oauth resolution

josschne:
Use exponential backoff as described in https://developers.google.com/drive/web/handle-errors to gracefully handle rate limiting.

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -2,6 +2,8 @@
 
 import copy
 import json
+import time
+import random
 import logging
 import webapp2
 import datetime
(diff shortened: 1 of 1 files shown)


gsfr:
@josschne and I had a chat about this today. We discussed necessity as well as pros and cons. I was worried that very long backoffs may actually lead to more token resolution requests, due to users frantically reloading our seemingly unresponsive page.

Holding this back for now. Next step is to ensure that the app hits ```/users/self``` *before* making any further requests. This will get very close to ensuring that only ever a single remote token resolution is needed per app load. It may even eliminate the need for scitran/api#43.

kofalt:
This won't work with the just-landed refactor. Closing pending a revival.

PR #45, opened 2015-09-09
optionally include public data in authenticated requests

gsfr:
This change was requsted by our friends at OpenfMRI.

@rentzso @kofalt Sanity check please.

### DIFF
--- a/api/containers.py
+++ b/api/containers.py
@@ -134,6 +134,8 @@ class ContainerList(base.RequestHandler):
                     query['permissions'] = {'$elemMatch': {'_id': uid or self.uid, 'site': self.source_site, 'access': 'admin'}}
                 else:
                     query['permissions'] = {'$elemMatch': {'_id': uid or self.uid, 'site': self.source_site}}
+            if self.request.GET.get('public', '').lower() in ('1', 'true'):
+                query['$or'] = [{'public': True}, {'permissions': query.pop('permissions')}]
             projection['permissions'] = {'$elemMatch': {'_id': uid or self.uid, 'site': self.source_site}}
         containers = list(self.dbc.find(query, projection))
         for container in containers:
@@ -163,11 +165,13 @@ class Container(base.RequestHandler):
                 self.abort(403, 'this ' + dbc_name + ' is not public')
             del container['permissions']
         elif not self.superuser_request:
-            if not user_perm:
+            if not container.get('public') and not user_perm:
                 self.abort(403, self.uid + ' does not have permissions on this ' + dbc_name)
             if min_role and users.INTEGER_ROLES[user_perm['access']] < users.INTEGER_ROLES[min_role]:
                 self.abort(403, self.uid + ' does not have at least ' + min_role + ' permissions on this ' + dbc_name)
-            if user_perm['access'] != 'admin': # if not admin, mask permissions of other users
+            if not user_perm:
+                container['permissions'] = []
+            elif user_perm['access'] != 'admin': # if not admin, mask permissions of other users
                 container['permissions'] = [user_perm]
         if self.request.GET.get('paths', '').lower() in ('1', 'true'):
             for fileinfo in container['files']:
--- a/api/containers.py
+++ b/api/containers.py
@@ -122,7 +122,7 @@ class ContainerList(base.RequestHandler):
         return json_body
 
     def _get(self, query, projection, admin_only=False, uid=None):
-        projection = {p: 1 for p in projection + ['files']}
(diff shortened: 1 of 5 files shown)


kofalt:
A bigger question: why would public data *not* be included in authenticated requests? I would consider that surprising behaviour.

gsfr:
I assert that users care more about their own data than about random data that some other user happened to have made public.

kofalt:
Review-wise, the logic is getting a bit dense for me. I think I get the first part...
Pinging @josschne or @andynemzek, can one of you take a whack at it?

I'm wondering if (probably after this PR) we could have the permissions layer draft a query postfix that A) filters objects the current context doesn't have perms to read, and B) fixes up the permissions array the way this code seems to do after the query. Then we could have the complex view logic in one place. At least, that's how I'd tackle it with RDB but I'm not sure the semantics are valid in mongo.

api/projects.py
```diff
@@ -119,7 +119,7 @@ def get(self, uid=None, gid=None):
             if not group:
                 self.abort(400, 'invalid group id')
         query = {'group': gid} if gid else {}
-        projection = ['group', 'name', 'notes', 'timestamp', 'timezone']
+        projection = ['group', 'name', 'notes', 'timestamp', 'timezone', 'public']
```
kofalt: Is there a reason only the projection for projects is changing, and not Container or any other object?

gsfr:
@rentzso is working on streamlining permissions logic as part of a larger refactoring to incorporate the new data model as well as race-less ways to update notes, permissions, etc. Let's try to have a call with him next week.

kofalt:
> I assert that users care more about their own data than about random data that some other user happened to have made public.

Ah. That sounds like it would apply to ContainerList._get, which I now see is different than the Container._get changes. It initially looked to me that a consumer would always need to pass the public flag to know that accessing a specific `_id` will succeed.

api/projects.py
gsfr: I was going for the minimal change set to achieve the requested behavior. For consistency, we may indeed want to update the projections of all containers.

kofalt: What was the requested behaviour change? Specifically a request regarding projects?

gsfr: Yes. The request was to be able to retrieve *permitted* and *public* projects in a single request and, more importantly, be able to access public data, despite sending an auth token. The first part is arguable, but the second part was just plain wrong before.

api/containers.py
```diff
@@ -134,6 +134,8 @@ def _get(self, query, projection, admin_only=False, uid=None):
                     query['permissions'] = {'$elemMatch': {'_id': uid or self.uid, 'site': self.source_site, 'access': 'admin'}}
                 else:
                     query['permissions'] = {'$elemMatch': {'_id': uid or self.uid, 'site': self.source_site}}
+            if self.request.GET.get('public', '').lower() in ('1', 'true'):
+                query['$or'] = [{'public': True}, {'permissions': query.pop('permissions')}]
```
josschne: I don't understand the {'permissions': query.pop('permissions')} part of this line.

gsfr: Yes, this one is a bit tricky. We move the 'permissions' field into the '$or' field.
```{'permissions': {...}}``` -> ```{'$or': [{'public': True}, {'permissions': {...}}]}```

rentzso:
@gsfr Just did the sanity check. Everything looks good.

@kofalt if you want to take a look at the code I am developing, on the branch refactoring there is a version to update container sublists like tags, group roles, container permissions, etc
https://github.com/scitran/api/tree/refactoring
It is WIP and it is not commented yet, but it starts to have a pretty consistent shape.

The idea is to follow this steps on each operation:
1) get the container 
2) check if it is a superuser request. If it is, perform the operation
3) if not, check the user permissions on the container
4) perform the operation if the user has the correct permissions

The permissions logic is defined in the decorator methods of this new module:
https://github.com/scitran/api/blob/refactoring/api/permchecker.py

Decorators may be  complex to wrap your head around them, but they allow to extract and isolate cleanly the permission logic. 

I haven't addressed the public request issue, but I believe it could be handled in a similar way to superuser requests.

gsfr:
The public field is now part of the projection for all container lists.

rentzso:
Latest commit looks good. OK to merge.

kofalt:
@rentzso Thanks, I'll definitely give that a read.

rentzso:
Thanks @kofalt I have created a (fake) pull request with the current version of the branch.
https://github.com/scitran/api/pull/47

PR #46, opened 2015-09-16
Multiple avatars

gsfr:


kofalt:
Should I interpret the assign as a request for review?

api/base.py
```diff
@@ -39,30 +42,25 @@ def __init__(self, request=None, response=None):
...
             else:
                 r = requests.get(self.app.config['oauth2_id_endpoint'], headers={'Authorization': 'Bearer ' + access_token})
-                if r.status_code == 200:
+                if r.ok:
                     identity = json.loads(r.content)
                     self.uid = identity.get('email')
                     if not self.uid:
                         self.abort(400, 'OAuth2 token resolution did not return email address')
                     self.app.db.authtokens.replace_one({'_id': access_token}, {'uid': self.uid, 'timestamp': request_start}, upsert=True)
+                    self.app.db.users.update_one({'_id': self.uid, 'firstseen': {'$exists': False}}, {'$set': {'firstseen': request_start}})
```
kofalt: This will only set `.firstseen` if the field has never been set?

api/base.py
```diff
@@ -39,30 +42,25 @@ def __init__(self, request=None, response=None):
...
-                        # u = urlparse(avatar)
-                        # query = parse_qs(u.query)
-                        # query.pop('sz', None)
-                        # u = u._replace(query=urlencode(query, True))
-                        # avatar = urlunparse(u)
-                        if avatar:
-                            r = self.app.db.users.update_one({'_id': self.uid, 'avatar': {'$ne': avatar}}, {'$set':{'avatar': avatar, 'modified': request_start}})
+                        provider_avatar = identity.get('picture')
+                        # Remove attached size param from URL.
+                        u = urlparse.urlparse(provider_avatar)
```
kofalt: You'll still want an `if provider_avatar` here to protect against crashing if the response does not contain an avatar.

api/base.py
```diff
@@ -96,9 +94,12 @@ def __init__(self, request=None, response=None):
         elif drone_request:
             self.superuser_request = True
         else:
-            user = self.app.db.users.find_one({'_id': self.uid}, ['root', 'wheel'])
+            user = self.app.db.users.find_one_and_update({'_id': self.uid}, {'$set': {'lastseen': request_start}}, ['root', 'wheel'])
```
kofalt: Are you sure you want to trigger a DB write on *every request*?
Or maybe I'm misreading this.

api/users.py
```diff
@@ -48,7 +49,11 @@ def post(self):
             json_body['created'] = json_body['modified'] = datetime.datetime.utcnow()
             json_body.setdefault('email', json_body['_id'])
             json_body.setdefault('preferences', {})
-            json_body.setdefault('avatar', 'https://gravatar.com/avatar/' + hashlib.md5(json_body['email']).hexdigest() + '?s=512&d=mm')
+            gravatar = 'https://gravatar.com/avatar/' + hashlib.md5(u['email']).hexdigest() + '?s=512'
+            if requests.head(gravatar, params={'d': '404'}):
+                u.setdefault('avatar', gravatar)
```
kofalt: I take it `requests.head` will evaluate as truthy IFF response was good?

kofalt:
Overall looks good. After each comment is resolved it's ready for merge.

Eric89GXL:
BTW I currently (on master) no longer have an avatar image by default, whereas before I had an image (from Google maybe?). From what I remember of discussions this was desired behavior, but I wanted to double check.

kofalt:
Yeah, my thought process is that I want to display the provider's avatar by default, since that is far more likely to be set than gravitar. OTOH, @gsfr has informed me about dilapidated Google orgs who can't set an avatar on their profile.

Here, we're working on storing both the provider's and gravitar's URL, and (soon) a user-provider URL. The server will only switch from gravitar to the provider if the gravitar was not set for that user.

Any API consumer will be able to modify their user document and set what's in use.

kofalt:
To express in JSON:

```js

// choices the user has
avatars: {
	"provider": "https://google.com/...",
	"gravitar": "https://gravitar.com/...",
	"custom":   "https://example.com/..."
}

// what avatar is in use
"avatar": "https://google.com/..."
```

kofalt:
This will enable

1. Optimal default behaviour for the average user (who will never configure their settings)
2. A user-interface that lets you view & choose your avatar

kofalt:
It's complicated by the fact that it's not possible to get a Google avatar for a user until they actually login. Thus the logic in this PR which will set the default avatar on login IFF it has never been set before.

Thoughts / comments?

Eric89GXL:
That sounds reasonable, and running this branch from a clean install my image shows up again.

api/base.py
gsfr: Yes. Updating to check for ```None```, which checks for existence as well as null value.

api/base.py
gsfr: Addressed by adding a empty string default to ```.get('picture')```.

api/base.py
gsfr: That's what that does right now. Open to discussion. I was thinking we'd try it and see how it performs, but on second thought, I will add a separate query to the remote token block.

Renaming ```firstsseen``` and ```lastseen``` to more appropriate ```firstlogin``` and ```lastlogin```, respectively.

api/users.py
gsfr: Exactly.

gsfr:
Commit added. All comments addressed. Thanks, @kofalt.

kofalt:
LGTM

PR #47, opened 2015-09-17
Refactoring

rentzso:
I am removing the WIP warning on the title of this Pull request. 
Except for some cleaning the code is ready to be reviewed. Also the method parse_file in util.py should be changed using the new schema for files.

Below I am keeping the original comment for reference.

#### Original comment - September 17

This is still a work in progress.

I am creating this pull request to have a preliminary code review.

This code  will handle updates to container sublists like tags, group roles, container permissions, etc

The idea is to follow this steps on each operation:
1) get the container 
2) check if it is a superuser request. If it is, perform the operation
3) check if it is a public request. If it is, perform only GET operation on public containers
4) if it is not a superuser or a public request, check the user permissions on the container
5) perform the operation if the user has the correct permissions

The permissions logic is defined in the decorator methods of this new module:
https://github.com/scitran/api/blob/refactoring/api/permchecker.py

Decorators may be complex to wrap your head around them, but they allow to extract and isolate cleanly the permission logic.


api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Specific behaviors (permissions checking logic for authenticated and not superuser users, storage interaction)
+    are specified in the routes defined in api.py
+    """
+
+    def __init__(self, request=None, response=None):
+        super(ListHandler, self).__init__(request, response)
+        self._initialized = None
+
+    def get(self, *args, **kwargs):
```
rentzso: remove permcheck and storage from kwargs and indicate them explicitly
remove *args from the function signature

api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+        container, perm_checker, storage = self._initialize_request(kwargs)
+        _id = container["_id"]
```
rentzso: single quotes

api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+        2) the storage class that will handle the database actions
+        3) the permission checker decorator that will be used
+        """
+        if self._initialized:
+            return self._initialized
+        perm_checker = kwargs.pop('permchecker')
+        storage = kwargs.pop('storage')
+        _id = kwargs.pop('cid', None) or kwargs.pop('_id')
+
+        storage.load_collection(self.app.db)
```
rentzso: replace with:
storage.dbc = self.app.db[storage.coll_name]

api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+            self.abort(404, 'Element not removed from list {} in collection {} {}'.format(storage.list_name, storage.coll_name, _id))
+
+    def _initialize_request(self, kwargs):
+        """
+        This method loads:
+        1) the container that will be modified
+        2) the storage class that will handle the database actions
+        3) the permission checker decorator that will be used
+        """
+        if self._initialized:
```
rentzso: is this caching useful?

api/permchecker.py
```diff
@@ -0,0 +1,82 @@
...
+
+def default_sublist(handler, container):
+    """
+    This is the default permissions checker generator.
+    The resulting permissions checker modifies the apply_change method by checking the user permissions
+    on the container before actually executing this method.
+    """
+    access = _get_access(handler.uid, container)
+    def g(apply_change):
+        def f(method, _id, elem_match = None, payload = None):
```
rentzso: elem_match should be called query

api/permchecker.py
```diff
@@ -0,0 +1,82 @@
...
+            if method == 'GET' and container.get('public', False):
+                min_access = -1
+            elif method == 'GET':
+                min_access = INTEGER_ROLES['ro']
+            elif method in Set(['POST', 'PUT', 'DELETE']):
+                min_access = INTEGER_ROLES['rw']
+            else:
+                min_access = float('inf')
```
rentzso: use MAX_INT

api/permchecker.py
```diff
@@ -0,0 +1,82 @@
...
+def group_roles_sublist(handler, container):
+    """
+    This is the customized permissions checker for group roles operations.
+    """
+    access = _get_access(handler.uid, container)
+    def g(apply_change):
+        def f(method, _id, elem_match = None, payload = None):
+            if method == 'GET' and elem_match.get('_id') == handler.uid:
+                return apply_change(method, _id, elem_match, payload)
+            elif method == 'PUT' and elem_match.get('_id') == handler.uid:
```
rentzso: delete 62 and 63

api/permchecker.py
```diff
@@ -0,0 +1,82 @@
...
+    return g
+
+def group_roles_sublist(handler, container):
+    """
+    This is the customized permissions checker for group roles operations.
+    """
+    access = _get_access(handler.uid, container)
+    def g(apply_change):
+        def f(method, _id, elem_match = None, payload = None):
+            if method == 'GET' and elem_match.get('_id') == handler.uid:
```
rentzso: add also possibility for a user to DELETE her own permissions

api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+        super(ListHandler, self).__init__(request, response)
+        self._initialized = None
+
+    def get(self, *args, **kwargs):
+        container, perm_checker, storage = self._initialize_request(kwargs)
+        _id = container["_id"]
+
+        result = perm_checker(storage.apply_change)('GET', _id, elem_match=kwargs)
+
+        if result is None or result.get(list_name) is None or len(result[list_name]) == 0:
```
rentzso: should be simplified to:
if result.get(list_name):

api/listhandler.py
```diff
@@ -0,0 +1,239 @@
...
+            self.abort(404, 'Element not found in list {} of collection {} {}'.format(storage.list_name, storage.coll_name, _id))
+        return result[list_name][0]
```
rentzso: probably this logic should go in the storage class

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+        self.coll_name = coll_name
+        self.list_name = list_name
+        self.use_oid = use_oid
+        # the collection is not loaded when the class is instantiated
+        # this allows to instantiate the class when the db is not available
+        # dbc is initialized using the load_collection method
+        self.dbc = None
+
+
+    def load_collection(self, db):
```
rentzso: remove this method as for comment in ListHandler class

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+
+        For simplicity we load its full content.
+        """
+        if self.dbc is None:
+            raise RuntimeError('collection not initialized before calling get_container')
+        if self.use_oid:
+            _id = bson.objectid.ObjectId(_id)
+        query = {'_id': _id}
+        log.debug('query {}'.format(query))
+        self.container = self.dbc.find_one(query)
```
rentzso: do not assign to self.container

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+            return self._delete_el(_id, elem_match)
+        if action == 'PUT':
+            return self._update_el(_id, elem_match, payload)
+        if action == 'POST':
+            return self._create_el(_id, payload)
+        raise ValueError('action should be one of GET, POST, PUT, DELETE')
+
+    def _create_el(self, _id, payload):
+        log.debug('payload {}'.format(payload))
+        if isinstance(payload, dict):
```
rentzso: to handle strings override this method in the StringListStorage class

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+            if payload.get('_id') is None:
```
rentzso: the logic for unique items should change, allowing different unique criteria:

e.g. site + uid for permissions, filename for files, no checks for notes

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+        log.debug('elem_match {}'.format(elem_match))
+        if isinstance(elem_match, str):
+            query_params = elem_match
+        else:
+            query_params = {'$elemMatch': elem_match}
+        query = {'_id': _id, self.list_name: query_params}
+        projection = {self.list_name + '.$': 1}
+        log.debug('query {}'.format(query))
+        log.debug('projection {}'.format(projection))
+        return self.dbc.find_one(query, projection)
```
rentzso: pass only the element in the list and not the container

api/dao/storage.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class StringListStorage(ListStorage):
+
+    def __init__(self, coll_name, list_name, use_oid, key_name):
```
rentzso: key_name should be always 'value'
the parameter gets removed

api/api.py
```diff
@@ -9,99 +9,204 @@
...
-        webapp2.Route(r'/<:[0-9a-f]{24}>/sessions',                 collections.CollectionSessions, name='coll_sessions'),
-        webapp2.Route(r'/<:[0-9a-f]{24}>/acquisitions',             collections.CollectionAcquisitions, name='coll_acquisitions'),
+        webapp2.Route(
+            r'/self',
+            userhandler.UserHandler, handler_method='self', methods=['GET']
+        ),
+        webapp2.Route(
+            r'/roles',
+            userhandler.UserHandler, handler_method='roles', methods=['GET']
+        ),
```
rentzso: this route should be removed.

api/api.py
```diff
@@ -9,99 +9,204 @@
...
+from handlers import listhandler
+from handlers import userhandler
+from handlers import grouphandler
+from handlers import containerhandler
+from handlers import collectionshandler
 
+#regexes used in routing table:
+routing_regexes = {
+    'group_id_re': '[0-9a-z][0-9a-z.@_-]{0,30}[0-9a-z]',
+    'cid_re': '[0-9a-f]{24}',
```
josschne: Suggestion to either make this variable name more readable, or add doc strings here.

api/api.py
```diff
@@ -9,99 +9,204 @@
...
-        webapp2.Route(r'/download',                                 core.Core, handler_method='download', methods=['GET', 'POST'], name='download'),
-        webapp2.Route(r'/upload',                                   core.Core, handler_method='upload', methods=['POST']),
-        webapp2.Route(r'/reaper',                                   core.Core, handler_method='reaper', methods=['POST']),
-        webapp2.Route(r'/sites',                                    core.Core, handler_method='sites', methods=['GET']),
-        webapp2.Route(r'/search',                                   core.Core, handler_method='search', methods=['GET', 'POST']),
-    ]),
-    webapp2_extras.routes.PathPrefixRoute(r'/api/schema', [
-        webapp2.Route(r'/group',                                    users.Group, handler_method='schema', methods=['GET']),
-        webapp2.Route(r'/user',                                     users.User, handler_method='schema', methods=['GET']),
+        webapp2.Route(
```
josschne: Suggestion to format these routes as before, despite the long line length. It is less readable now.

api/api.py
```diff
@@ -9,99 +9,204 @@
...
+    webapp2.Route(
+        _format(r'/api/groups/<_id:{group_id_re}>'),
+        grouphandler.GroupHandler, name='group'
+    ),
+    webapp2.Route(
+        _format(r'/api/<cont_name:groups>/<cid:{group_id_re}>/<list_name:roles>'),
+        listhandler.ListHandler, name='group'
+    ),
+    webapp2.Route(
+        _format(r'/api/<cont_name:groups>/<cid:{group_id_re}>/<list_name:roles>/<site:{site_re}>/<_id:{user_id_re}>'),
```
josschne: Suggestion to add comments on these routes. The new routes remove duplication (which is great!) but they take a bit more time to read and understand. Comment could help here.

api/api.py
```diff
@@ -9,99 +9,204 @@
...
+        _format(r'/api/<cont_name:{cont_name_re}>'),
+        containerhandler.ContainerHandler, name='cont_list', handler_method='get_all', methods=['GET']
+    ),
+    webapp2.Route(
+        _format(r'/api/<cont_name:{cont_name_re}>'),
+        containerhandler.ContainerHandler, methods=['POST']
+    ),
+    webapp2.Route(
+        _format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>'),
+        containerhandler.ContainerHandler, name='cont_details', methods=['GET','PUT','DELETE']
```
josschne: Do we use the "name=" param in the code for every route? If not, might be able to eliminate them to reduce visual noise.

api/core.py
```diff
@@ -370,7 +202,7 @@ def _archivestream(self, ticket):
         stream.close()
 
     def download(self):
```
josschne: Suggest rename to bulk_download to distinguish between downloading a tar and downloading an individual file

api/dao/liststorage.py
```diff
@@ -0,0 +1,173 @@
...
+
+from .. import mongo
+from . import APIStorageException
+
+log = logging.getLogger('scitran.api')
+
+
+class ListStorage(object):
+    """
+    This class provides access to sublists of a mongodb collections elements (called containers).
```
josschne: Unsure what the distinction is between ContainerStorage and ListStorage.

api/dao/liststorage.py
```diff
@@ -0,0 +1,173 @@
...
+
+        For simplicity we load its full content.
+        """
+        if self.use_oid:
+            _id = bson.objectid.ObjectId(_id)
+        query = {'_id': _id}
+        projection = None
+        if query_params:
+            query[self.list_name] = {
+                '$elemMatch': query_params
```
josschne: Question - where are we sanitizing the query_params?

api/dao/liststorage.py
```diff
@@ -0,0 +1,173 @@
...
+        query = {'_id': _id, self.list_name: {'$elemMatch': query_params}}
+        projection = {self.list_name + '.$': 1}
+        log.debug('query {}'.format(query))
+        log.debug('projection {}'.format(projection))
+        result = self.dbc.find_one(query, projection)
+        if result and result.get(self.list_name):
+            return result.get(self.list_name)[0]
+
+
+class StringListStorage(ListStorage):
```
josschne: Suggestion to put a doc explaining what a StringList is.

api/files.py
```diff
@@ -0,0 +1,140 @@
...
+
+    1) initialize the request
+    2) save a temporary file
+    3) check identical
+    4) move the temporary file to its destination
+
+    The operations could be safely interleaved with other actions like permission checks or database updates.
+    """
+
+    def __init__(self, client_addr, filename, body, received_md5, metadata, tags):
```
josschne: Is this really an MD5 still? If so, I thought we had moved to SHA384

api/files.py
```diff
@@ -0,0 +1,140 @@
...
+            for chunk in iter(lambda: self.body.read(2**20), ''):
+                md5.update(chunk)
+                sha384.update(chunk)
+                filesize += len(chunk)
+                fd.write(chunk)
+        self.filesize = filesize
+        self.md5 = md5.hexdigest()
+        self.sha384 = sha384.hexdigest()
+        duration = datetime.datetime.utcnow() - start_time
+        success = (self.md5 == self.received_md5) if self.received_md5 is not None else True
```
josschne: Why isn't SHA384 compared here? Why do we use two hashes?

api/files.py
```diff
@@ -0,0 +1,140 @@
...
+        self.metadata = metadata
+        self.tags = tags
+        self.mimetype = util.guess_mimetype(filename)
+        self.filetype = util.guess_filetype(filename, self.mimetype)
+
+    def save_temp_file(self, tempdir_path):
+        self.tempdir_path = tempdir_path
+        success, duration = self._save_temp_file(self.tempdir_path)
+        if not success:
+            self.abort(400, 'Content-MD5 mismatch.')
```
josschne: Do we want to explicitly name MD5 in this error string?

api/handlers/containerhandler.py
```diff
@@ -0,0 +1,352 @@
...
+    4) augment the payload when appropriate
+    5) exec the request (using the mongo validator and the permissions checker)
+    6) check the result
+    7) augment the result when needed
+    8) return the result
+
+    Specific behaviors (permissions checking logic for authenticated and not superuser users, storage interaction)
+    are specified in the container_handler_configurations
+    """
+    use_oid = {
```
josschne: What is oid?

api/handlers/containerhandler.py
```diff
@@ -0,0 +1,352 @@
...
+    # the permissions checker and the json schema validators used to handle a request.
+    #
+    # "children_cont" represents the children container.
+    # "list projection" is used to filter data in mongo.
+    container_handler_configurations = {
+        'projects': {
+            'storage': containerstorage.ContainerStorage('projects', use_oid=use_oid['projects']),
+            'permchecker': containerauth.default_container,
+            'parent_storage': containerstorage.ContainerStorage('groups', use_oid=use_oid['groups']),
+            'mongo_schema_file': 'mongo/project.json',
```
josschne: Suggest renaming to storage_schema_file

api/handlers/containerhandler.py
```diff
@@ -0,0 +1,352 @@
...
+        'acquisitions': True
+    }
+    default_list_projection = ['files', 'notes', 'timestamp', 'timezone', 'public']
+
+    # This configurations are used by the ContainerHandler class to load the storage,
+    # the permissions checker and the json schema validators used to handle a request.
+    #
+    # "children_cont" represents the children container.
+    # "list projection" is used to filter data in mongo.
+    container_handler_configurations = {
```
josschne: Could this be simplified to a couple of constructor arguments - parent, child, me? It appears all of the various containers have the same values except for those three variables.

api/handlers/containerhandler.py
```diff
@@ -0,0 +1,352 @@
...
+        # In this case, we will check permissions on it.
+        target_parent_container, parent_id_property = self._get_parent_container(payload)
+        if target_parent_container:
+            if cont_name == 'sessions':
+                payload['group'] = target_parent_container['group']
+            payload['permissions'] = target_parent_container.get('roles')
+            if payload['permissions'] is None:
+                payload['permissions'] = target_parent_container['permissions']
+
+        permchecker = self._get_permchecker(container, target_parent_container)
```
josschne: Suggestion to move permchecker closer to its use so its lifetime is minimized

api/api.py
kofalt: Every single one of these regexes needs a comment explaining:

1. What the regex is for
1. How the regex is put together & how it operates

The group ID regex is a great example of a regex that needs verbose justification.

api/api.py
kofalt: Just to add a bit more - I could successfully understand about 1 in 3 of these routes before an extensive review of the classes in question, and about 2 in 3 afterwards.

Please add a comment per route. The routing table needs to be excessively clear, IMO

api/handlers/listhandler.py
```diff
@@ -0,0 +1,416 @@
...
+        if os.path.exists(filepath):
+            os.remove(filepath)
+            log.info('removed file ' + filepath)
+            result['removed'] = 1
+        else:
+            log.warning(filepath + ' does not exist')
+            result['removed'] = 0
+        return result
+
+    def put(self, cont_name, list_name, **kwargs):
```
josschne: Remove function

api/util.py
```diff
@@ -115,8 +113,11 @@ def commit_file(dbc, _id, datainfo, filepath, data_path, force=False):
                 else: # existing file has different content
                     log.debug('Replacing   %s' % filename)
                     shutil.move(filepath, target_filepath)
+                    update_set = {'files.$.dirty': True, 'files.$.modified': datetime.datetime.utcnow()}
+                    for k,v in fileinfo.iteritems():
+                        update_set['files.$.' + k] = v
                     dbc.update_one({'_id':_id, 'files.filename': fileinfo['filename']},
-                            {'$set': {'files.$.dirty': True, 'files.$.modified': datetime.datetime.utcnow()}})
+                            {'$set': update_set})
```
josschne: Looking for more explanation on what update_set does and what this change is about.

api/validators.py
```diff
@@ -0,0 +1,152 @@
...
+    'tag.json'
+])
+mongo_schemas = set()
+input_schemas = set()
+# validate and cache schemas at start time
+for schema_file in os.listdir(schema_path + '/schemas/mongo/'):
+    mongo_schemas.add(schema_file)
+    resolver.resolve('mongo/' + schema_file)
+
+assert mongo_schemas == expected_mongo_schemas, '{} is different from {}'.format(mongo_schemas, expected_schemas)
```
josschne: Why is it important to compare schemas in the directory to a known set?

api/validators.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            _schema = schema
+        if method in ['POST', 'PUT']:
+            try:
+                _validate_json(payload, _schema)
+            except jsonschema.ValidationError as e:
+                handler.abort(400, str(e))
+    return g
+
+def key_check(handler, schema_file):
```
josschne: What does this function do?

bin/api.wsgi
```diff
@@ -55,75 +38,64 @@ args.site_name = ' '.join(args.site_name).strip('"\'')
...
 args.upload_path = os.path.join(args.data_path, 'upload')
 
-app.config = vars(args)
+from api import mongo
+mongo.configure_db(args.db_uri, args.site_id, args.site_name, args.api_uri)
+
+from api.util import log
+from api import api
+from api import centralclient, jobs
+from api import jobs
```
josschne: Is there a reason to have these imports lower in the file?

api/mongo.py
```diff
@@ -0,0 +1,26 @@
+import pymongo
+import datetime
+
+db = None
+
+
+def configure_db(db_uri, site_id, site_name, api_uri):
+    global db
+    db = pymongo.MongoClient(db_uri).get_default_database()
```
josschne: This no longer has a sleep loop. How can we launch this robustly?

bin/api.wsgi
```diff
@@ -55,75 +38,64 @@ args.site_name = ' '.join(args.site_name).strip('"\'')
...
-    @uwsgidecorators.cron(0, -1, -1, -1, -1)  # top of every hour
-    def upload_storage_cleaning(signum):
-        upload_path = application.config['upload_path']
-        for f in os.listdir(upload_path):
-            fp = os.path.join(upload_path, f)
-            timestamp = datetime.datetime.utcfromtimestamp(int(os.stat(fp).st_mtime))
-            if timestamp < (datetime.datetime.utcnow() - datetime.timedelta(hours=1)):
-                log.debug('upload %s was last modified %s' % (fp, str(timestamp)))
-                os.remove(fp)
-
```
josschne: Why is this no longer needed?

bin/bootstrap.py
```diff
@@ -76,7 +78,7 @@ def sort(args):
     log.info('found %d files to sort (ignoring symlinks and dotfiles)' % file_cnt)
     for i, filepath in enumerate(files):
         log.info('sorting     %s [%s] (%d/%d)' % (os.path.basename(filepath), util.hrsize(os.path.getsize(filepath)), i+1, file_cnt))
-        hash_ = hashlib.sha1()
+        hash_ = hashlib.sha384()
```
josschne: Why do we hash here? Why was this changed?

test_integ/test_collection.py
```diff
@@ -0,0 +1,106 @@
...
+    r = requests.delete(base_url + '/acquisitions/' + test_data.aid + '?user=admin@user.com&root=true', verify=False)
+    assert r.ok
+    r = requests.delete(base_url + '/sessions/' + test_data.sid + '?user=admin@user.com&root=true', verify=False)
+    assert r.ok
+    r = requests.delete(base_url + '/projects/' + test_data.pid + '?user=admin@user.com&root=true', verify=False)
+    assert r.ok
+
+
+@with_setup(setup_db, teardown_db)
+def test_sequence():
```
josschne: Can we name this test to describe what is being tested?

josschne:
Speaking for myself, @andynemzek, and @kofalt - would like to see responses to the existing comments, but preliminary LGTUs

rentzso:
Thanks guys for the code review!
I will start working on your comments tomorrow.

kofalt:
I don't think you planned to do this, but please explicitly do **not** force push this branch.

The diff is extensive and I plan to review fc72b56 against any further changes made.

api/api.py
rentzso: we use the name param for the routes that are reused (eg when creating the debug infos for a request)

api/dao/liststorage.py
rentzso: I will add a better comment and an example to each class docstring

api/files.py
rentzso: MD5 is used for transport hashing
SHA384 for storage hashing (eg an algorithm would use this hash to check if it has been run before on the same data).

api/files.py
rentzso: See comment above.
Content-MD5 is the standard hashing for upload over HTTP.

api/files.py
kofalt: Let's use the same algorithm for both purposes. MD5 is obsolete for a variety of reasons

api/files.py
rentzso: Good catch.
self.abort is not available here.

success is false only if there is a Content-MD5 mismatch, so I prefer to be specific in the message.

api/files.py
rentzso: We discussed this at the time with Kevin, and we decided to use MD5 as some cloud services (eg Amazon S3) supports it and it is not very computationally expensive specially on the browser.

Now there is a discussion going on with @gsfr and @josschne about not supporting any hashing on file transport at all.
Maybe we can continue it on a github issue. Let me know what you think about.

api/handlers/containerhandler.py
rentzso: it means object id.
groups use simple strings for ids,
projects, sessions and acquisitions instead use them.

I will add a comment ;)

api/handlers/containerhandler.py
rentzso: The intent here is to give maximum flexibility by decoupling all the different component of the handlers.

If you remove them from the configuration, you are coupling and you won't be able to quickly change them in the future.

api/validators.py
rentzso: This allow to check at API start time that all the needed schemas (and only these) are present.

bin/api.wsgi
rentzso: Yes, we need to do these imports after mongo e the db handle has been initialized.

api/mongo.py
rentzso: good point.
I will add a try catch block similar to the one that existed before.

bin/api.wsgi
rentzso: the new API won't support sorted upload (you may see that the method isn't in core.py anymore).
This method was cleaning the files created only by that method, as uploads on the */files route automatically removes the temporary files afterwards.

bin/bootstrap.py
rentzso: we decided to change the hash algorithm for storage hashing to reduce the collision rate.

api/validators.py
rentzso: for sublists of mongo container there is no automatic key check when creating, updating or deleting an object. We are adding a custom array field to the json schemas ("key_fields").
The uniqueness is checked on the combination of all the array "key_fields".
For an example you can check https://github.com/scitran/api/blob/refactoring/api/schemas/input/permission.json

So this method ensures that:
1. after a POST and PUT request we don't have two items with the same values for the key set
2. a GET will retrieve a single item
3. a DELETE (most importantly) will delete a single item

I will add a comment with the description.

api/util.py
rentzso: in this branch of the code, we are overriding an existing file.
```update_set``` allows to update all the file info like size, hash, etc.

This comes from this bugfix for the BIDS guys: https://github.com/scitran/api/commit/dc198a9b6010b8562d5f4c9170b5ab966ebab682

api/dao/liststorage.py
rentzso: Good question:
actually the variable name is kind of misleading and needs to be changed.
the query_params are not the request query_params
but essentially they are the variable extracted in the route using the regexes.
As they are mapped from the **kwargs in the handler methods.
https://webapp-improved.appspot.com/guide/routing.html#the-url-template

So that is were the validation happens: in the routing table.

rentzso:
Just finished reviewing your comments. Please let me know if the answers are satisfactory enough or if some of them need a follow up.
The comments without an answer are implicitly acknowledged. 
I will push a new version of the code with appropriate modifications by today/tomorrow morning.

api/api.py
rentzso: @kofalt  could you give me an example comment for a route (maybe one that you found particularly difficult to understand)?

I definitely want the routing table to be excessively clear, but I would like to avoid too verbose or redundant content.

rentzso: I have created an issue https://github.com/scitran/api/issues/63 so that we can move forward with the merge.

rentzso:
@josschne @kofalt @andynemzek I just pushed 3 new commits to the refactoring branch.
This is mainly where I address your comments: https://github.com/scitran/api/commit/7d7a9acddad021c08fc6843dad0c68c636c48aab
Let me know if you think I left out something.

We are planning to merge it later today/Monday morning.

Thanks again for the hard work on the review!

rentzso:
addressed @andynemzek comments in 91f6be1badae716d4635fbad9daa2a185131c0cd

api/validators.py
josschne: I'll leave this to your judgement.

bin/api.wsgi
josschne: Thanks!

PR #48, opened 2015-09-18
Add session counts to projects and collections

josschne:
Add a session count to the projects and collections endpoints

### DIFF
--- a/api/collections.py
+++ b/api/collections.py
@@ -219,9 +219,9 @@ class Collections(containers.ContainerList):
             {'$group': {'_id': "$collections", 'sessions': {'$addToSet': "$session"}}}
             ])
         # Convert to dict for easy lookup
-        session_counts = {str(coll['_id']): len(coll['sessions']) for coll in session_counts}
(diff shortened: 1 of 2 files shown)


gsfr:
Thanks, @josschne.

I'm concerned about the possible performance impact of the these additions on large databases. ```aggregate``` operations, in particular, are not cheap.

Looks like those changes are needed to close current FDM tickets, so I'm open to merging them now and revisiting later.

If counts turn out to become always-needed information, we may want to reverse the data structure in the database. Instead of having sessions point to their projects, projects will hold a list of their sessions and thus immediately know how many they contain.

mrDarcyMurphy:
Tried this out for a separate ticket and it worked great. :+1:

gsfr:
@josschne Feel free to merge. The only thing I would change is to use the ```ObjectId```s themselves as dictionary keys, instead of their string representations. That cuts out some clutter and seemingly useless conversions.

josschne:
Thanks all. I'll make the suggested update and merge.

PR #49, opened 2015-09-21
Job api fixes

kofalt:
Some small fixes:

* In `spawn_jobs`:
  * do not attempt to enqueue non-dicom data
  * correct parameter type passed to `queue_job`
* In `queue_job`:
  * Correct some docs
  * Correct a duplicate / erroneous job map entry
* In `Job.put`:
  * Correct query parameter
  * Check query result & inform client
* In `api.wsgi`:
  * Process job creation queue (`job_creation`) on launch before setting cron timer
  * Easier to debug, prevents any initial rush

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -62,6 +62,9 @@ def spawn_jobs(db, containers, file):
         File object that is used to spawn 0 or more jobs.
     """
 
+    if file['filetype'] != 'dicom':
+        return
+
     # File information
     filename = file['filename']
     filehash = file['filehash']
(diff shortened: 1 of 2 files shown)


gsfr:
LGTM.

There is at least one use of camel case (```jobQuery```) that should be converted to underscore.

Also, it is important that the initial job creation invocation communicates well with the operator as it may considerably delay startup of the entire API. In that line, I have been longing for certain info to be passed through by ```live.sh``` or ```flywheel.sh```.

kofalt:
Style corrected & initial logging statement added.

Agreed on logging comment. We need an integrated health-check. This would also replace the sleep intervals in bootstrapping. Not sure what would be best; I'd be open to a ticket discussion on sci/sci.

PR #50, opened 2015-09-21
Retry failed jobs in every instance

kofalt:
There is a cron job that will check for jobs that have not been heartbeated in awhile, declare them as failed, and retry them based on a max-attempts limit.

However, when a client reports a job as failed, this retry logic was not called.

This PR moves the retry logic to a central location, and calls it in both scenarios.

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -166,6 +166,19 @@ def queue_job(db, algorithm_id, container_type, container_id, filename, filehash
     log.info('Running %s as job %s to process %s %s' % (algorithm_id, str(_id), container_type, container_id))
     return _id
 
+def retry_job(db, j):
+    """
+    Given a failed job, either retry the job or fail it permanently, based on the attempt number.
+    TODO: make max attempts configurable
+    """
+
+    if j['attempt'] < 3:
+        job_id = queue_job(db, j['algorithm_id'], j['container_type'], j['container_id'], j['filename'], j['filehash'], j['attempt']+1, j['_id'])
+        log.info('respawned job %s as %s (attempt %d)' % (j['_id'], job_id, j['attempt']+1))
+    else:
+        log.info('permanently failed job %s (after %d attempts)' % (j['_id'], j['attempt']))
+
+
 class Jobs(base.RequestHandler):
 
     """Provide /jobs API routes."""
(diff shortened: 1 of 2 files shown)


kofalt:
This was tested by triggering random job failures.

Before: Total job count does not increase, including failed
After: Total job count increases

kofalt:
This was tested by triggering random job failures.

Before: Total job count does not increase, including failed.
Steady state was initial job count != complete job count.

After: Total job count increases when a job fails.
Steady state was initial job count == complete job count.

api/jobs.py
```diff
@@ -267,3 +280,8 @@ def put(self, _id):
         result = self.app.db.jobs.update_one(job_query, {'$set': mutation})
         if result.modified_count != 1:
             self.abort(500, 'Job modification not saved')
+
+        # If the job did not succeed, check to see if job should be retried.
+        if 'state' in mutation and mutation['state'] == 'failed':
```
gsfr: I typically find it cleaner to write this as ```if mutation.get('state') == 'failed':```, but either way works.

gsfr:
LGTM.

Any reason we should use distinct states for timed-out jobs vs failed+retried jobs vs permanently failed jobs?

api/jobs.py
kofalt: Noted. I keep forgetting about that `get` syntax, is that added by a library or standard for all maps?

kofalt:
Two reasons I can think of that we might:

1. We may want to know the cause of observed job failures
1. We may want to keep all job processing in a cron job.

I disagree with number 2 outright, as whenever possible we should minimize processing latency. Regarding number 1, in my opinion understanding job success / failure is a whole other topic that I'd like to expand upon - we don't save exits codes, let alone more detailed fault analysis.

When we tackle that discussion we might decide to add more states to the machine, but I suspect it'll be more likely we start having a dedicated results object with useful stuff on it.

kofalt:
Or at least, that's my current thought process with the state machine holding a single attempt. If we linked attempts together, say having a state machine that encapsulates retries, that opinion could change.

api/jobs.py
gsfr: Standard for all [maps](https://docs.python.org/2/library/stdtypes.html#mapping-types-dict).

gsfr:
Fair.

PR #51, opened 2015-09-24
Correct abandoned jobs query

kofalt:
The dead-jobs query was using an invalid key.
Because of this, abandoned jobs would never be reclaimed.

### DIFF
--- a/bin/api.wsgi
+++ b/bin/api.wsgi
@@ -171,7 +171,7 @@ else:
             j = application.db.jobs.find_one_and_update(
                 {
                     'state': 'running',
-                    'heartbeat': {'$lt': datetime.datetime.utcnow() - datetime.timedelta(seconds=100)},
+                    'modified': {'$lt': datetime.datetime.utcnow() - datetime.timedelta(seconds=100)},
                 },
                 {
                     '$set': {


gsfr:
Somehow I'm thinking that ```modified``` and ```heartbeat``` are two different things. ```modified``` might indicate the last state change, for instance.

kofalt:
We could do that, but if we did, the following (at minimum) would need to change:

1. A job should update a `heartbeat` key in addition to `modified` when transitioning to `running`
1. Heartbeating a job should update the `heartbeat` key rather than `modified`
1. Heartbeating a job should be  `/heartbeat` or similar, rather than a generic `PUT`

Thus, we should either use the current behaviour, or use `heartbeat` and make the above changes.

I like your idea of recording how long the job has been in a given state, but the `modified` key won't serve that purpose right now. *Any* modification updates that key, by design - so if anyone updates the doc for any reason, or if we later start recording progress on a job object, that information would be lost. To serve that purpose we'd need a `lastStateChange` key that only the server is allowed to mutate, or more completely a `stateChange` map that records when the job last changed to each state.

Currently, I'm fine with the `modified` timestamp pulling double-duty, but that might change if we start recording progress or something on the job object. Another choice is if updating `heartbeat` should also update `modified` - my opinion is yes, but open to discussion.

In summary, using a `heartbeat` key isn't how the jobs API is currently designed to work. But it could be! My proposal is that we merge this bugfix first, and if we're thinking `heartbeat` keys are a good idea let's discuss in a new ticket / PR.

gsfr:
I agree with all that. Merging now. For the record, the only thing that does not make sense to me is updating ```modified``` when updating ```heartbeat```. I see no benefit in having a ```heartbeat``` field in that case. Other than that, I like your idea of keeping a record of all state changes.

PR #52, opened 2015-09-25
Generate job information on-the-fly

kofalt:
Currently a blob of JSON is saved into a job struct when the job is enqueued. There's no particular reason to do this, since it effectively serializes logic into the database before it's used.

This PR:

1. Moves non-formula keys required to generate a formula into an `intent` object
1. Adds a `generate_formula` which takes an intent
1. Upgrades `/next` to save & return the formula at the moment of launch.

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -248,7 +248,7 @@ class Jobs(base.RequestHandler):
             return_document=pymongo.collection.ReturnDocument.AFTER
         )
 
-        if result == None:
+        if result is None:
             self.abort(400, 'No jobs to process')
 
         # Second, update document to store formula request.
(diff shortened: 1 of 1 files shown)


api/jobs.py
```diff
@@ -231,6 +251,20 @@ def next(self):
...
+            {
+                '_id': result['_id']
+            },
+            { '$set': {
+                'request': generate_formula(result['intent'])}
+            },
+            return_document=pymongo.collection.ReturnDocument.AFTER
+        )
+
+        if result == None:
```
gsfr: While this yields the correct result, the common pythonism for equality to ```None``` is ```result is None```. Also in line 251 above.

I often use the more casual ```if not result```, although that evaluates to ```true``` in many more cases, such as ```{}, [], '', 0```.

gsfr:
LGTM.

api/jobs.py
kofalt: Type comparison makes sense. I want the explicit nil check, so I'll change to `is None` and merge.

gsfr: Sounds good.

PR #53, opened 2015-09-29
allow avatars.* to be set on a user

gsfr:


gsfr:
@mrDarcyMurphy The first commit allows ```avatars.*``` to be set, but your code currently sends in the complete ```avatars``` object, which is allowed by the second commit. Either one works for me.

mrDarcyMurphy:
@gsfr Just tested it, worked for me. Thanks.

gsfr:
@mrDarcyMurphy Great.

Do you want to keep your code as is and completely overwrite ```avatars``` with each update or refactor to use my first commit and only update ```avatars.custom```?

mrDarcyMurphy:
Good question. Users can update their gravatar/g+avatar outside of flywheel. Does the API handle that at all right now? I'll double check how the UI handles it in the mean time.

gsfr:
The Google (any provider in the future) avatar is updated each time the API resolves an uncached token. Gravatar works on hashed email addresses. Image changes are automatic, but email changes are not handled right now.

mrDarcyMurphy:
@gsfr That's fine. The email should stay the same because it's used for both the G+ account and to resolve the gravatar. 

I've also confirmed that they both update after logging out and in again. Thought there can be a lag of a few minutes for the changes to take effect. 

Knowing that, it's probably safe to only update `profile.avatars.custom` and `profile.avatar`. Though I'd like to be able to wipe out `profile.avatars.custom`, otherwise we're hot-linking someone else's content every time the profile page is rendered.

mrDarcyMurphy:
@gsfr @kofalt I think we're good to go here. I've left my client code compatible with this endpoint style for now.

gsfr:
Thanks, @mrDarcyMurphy.

I had more discussion on this with @rentzso last week. We concluded that accepting nested data structures is indeed the way to go. After all, the API should consume data in the same format as it delivers it. Ambiguous semantics need to be addressed, but we came up with a simple solution that fits our needs. Will present in person for discussion end of week.

mrDarcyMurphy:
@gsfr Sounds good.

PR #55, opened 2015-10-16
Advanced filtering

gsfr:
@rentzso LGTM. Are you ok to merge?

### DIFF
--- a/api/sessions.py
+++ b/api/sessions.py
@@ -115,11 +115,11 @@ class Sessions(containers.ContainerList):
             query = {'group': gid}
         else:
             query = {}
-        projection = ['label', 'subject_code', 'subject.code', 'project', 'group']
+        projection = ['label', 'subject_code', 'subject.code', 'project', 'group', 'subject.age', 'subject.sex']
         sessions = self._get(query, projection, self.request.GET.get('admin', '').lower() in ('1', 'true'))
         for sess in sessions:
             if 'subject_code' not in sess:
-                sess['subject_code'] = sess.pop('subject', {}).get('code', '') # FIXME when subject is pulled out of session
+                sess['subject_code'] = sess.get('subject', {}).get('code', '') # FIXME when subject is pulled out of session
         if self.debug:
             for sess in sessions:
                 sid = str(sess['_id'])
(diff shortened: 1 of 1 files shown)


api/sessions.py
```diff
@@ -115,11 +115,19 @@ def get(self, pid=None, gid=None):
...
+            session_measurements = self.app.db.acquisitions.aggregate([
+                {'$match': {'session': {'$in': [sess['_id'] for sess in sessions]}}},
+                {'$group': {'_id': '$session', 'measurements': {'$addToSet': '$datatype'}}}
+                ])
+            session_measurements = {sess['_id']: sess['measurements'] for sess in session_measurements}
         for sess in sessions:
+            sess['measurements'] = session_measurements.get(sess['_id'], None)
             if 'subject_code' not in sess:
-                sess['subject_code'] = sess.pop('subject', {}).get('code', '') # FIXME when subject is pulled out of session
+                sess['subject_code'] = sess.get('subject', {}).get('code', '') # FIXME when subject is pulled out of session
```
kofalt: Was `pop` the correct approach before this change?

api/sessions.py
```diff
@@ -115,11 +115,19 @@ def get(self, pid=None, gid=None):
...
         sessions = self._get(query, projection, self.request.GET.get('admin', '').lower() in ('1', 'true'))
+        session_measurements = {}
+        if self.request.GET.get('measurements', '').lower() in ('1', 'true'):
+            session_measurements = self.app.db.acquisitions.aggregate([
+                {'$match': {'session': {'$in': [sess['_id'] for sess in sessions]}}},
+                {'$group': {'_id': '$session', 'measurements': {'$addToSet': '$datatype'}}}
+                ])
+            session_measurements = {sess['_id']: sess['measurements'] for sess in session_measurements}
         for sess in sessions:
+            sess['measurements'] = session_measurements.get(sess['_id'], None)
```
kofalt: To me, it looks like this line will overwrite any pre-existing `measurements` key on the object if `measurements` was not passed as a URL param. Is there any case where that might be a problem?

rentzso: 'measurements' isn't a key on session objects.

api/sessions.py
rentzso: get is probably better :)

api/sessions.py
kofalt: Cool, just checking.

api/sessions.py
kofalt: Haha, okay :)

PR #56, opened 2015-10-21
Simplify root & wheel fields

gsfr:


rentzso:
LGTM

kofalt:
Context: this PR removes the `wheel` user property, which previously indicated the *ability* of a user to statefully (server-side) act as a superuser. The `root` user property previously held that state - whether any request from that user was a superuser request - now serves the purpose old `wheel` field had.

Upshot: You no longer toggle a global superuser mode on the server - instead, if your user property `root` is set to true, you can send `root=1` as a parameter on any request and be treated as a superuser. This makes it much easier to selectively make superuser requests as desired without toggling a global switch.

kofalt:
LGTM

PR #57, opened 2015-10-21
add measurements on CollectionSessions

rentzso:


rentzso:
tested by @andynemzek

api/collections.py
```diff
@@ -315,11 +315,19 @@ def get(self, cid):
...
-        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1}
+        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1, 'subject.age': 1, 'subject.sex': 1}
         projection['permissions'] = {'$elemMatch': {'_id': self.uid, 'site': self.source_site}}
         sessions = list(self.dbc.find(query, projection)) # avoid permissions checking by not using ContainerList._get()
+        session_measurements = {}
+        if self.request.GET.get('measurements', '').lower() in ('1', 'true'):
+            session_measurements = self.app.db.acquisitions.aggregate([
+                {'$match': {'session': {'$in': [sess['_id'] for sess in sessions]}}},
+                {'$group': {'_id': '$session', 'measurements': {'$addToSet': '$datatype'}}}
+                ])
```
kofalt: Not sure I understand this aggregation, could you walk me through it?

rentzso:
Sure!
There are two parts in this aggregation:
1) $match -> select all acquisitions with session in the list "sessions"
2) $group -> group by _id ($session that is a reference to the session field of each matched document) and use the mongodb "accumulator" $addToSet gathering the datatypes of all the acquisitions contained in a given session

for a detailed explanation of $group this is the mongodb page: https://docs.mongodb.org/manual/reference/operator/aggregation/group/ 

But if something is still not clear, let me know :)

kofalt:
That makes sense, thanks!

Is the result of that query a unique set, or just a set of every match?
Or to put it another way - if there are duplicate `datatype` matches, are duplicates removed?

rentzso:
yep duplicates are removed by $addToSet

PR #61, opened 2015-11-06
Bids download

rentzso:
This merge will allow downloads in BIDS format.

It also fixes two bugs.


gsfr:
LGTM.

@rentzso Please make sure we also get all this functionality into master after refactoring lands. Please create an appropriate ticket, if applicable.

kofalt:
@rentzso when merging PRs, don't forget to then hit the "Delete branch" button if appropriate.
I haven't done so here in case you explicitly wanted to leave this branch around.

rentzso:
@kofalt thanks for the advice but in fact we need to keep this branch a little more for the BIDS guys.

PR #62, opened 2015-11-12
Enable j option on pymongo.MongoClient() connections.

ryansanford:
This option requires writes are committed to the journal before we get a response form Mongo.

### DIFF
--- a/bin/api.wsgi
+++ b/bin/api.wsgi
@@ -81,7 +81,8 @@ if not os.path.exists(app.config['upload_path']):
 
 for x in range(10):
     try:
-        app.db = pymongo.MongoClient(args.db_uri).get_default_database()
+        # j option specifies writes must commit to journal before response is received.
+        app.db = pymongo.MongoClient(host=args.db_uri, j=True).get_default_database()
     except:
         time.sleep(6)
     else:


kofalt:
LGTM; ping @gsfr for sanity check.

kofalt:
This makes sense so just merging for now.

gsfr:
Thanks guys.

Adding this for future reference:
screen shot 2015-11-12 at 9 38 04 pm

kofalt:
Nice. @ryansanford based on the above table, should we also add w=1?

ryansanford:
The 'w=1' option would be redundant when using the 'j' option, but that's
not a bad thing. Also, w=1 is the default since Mongo 2.2. Specifying it
would be verbose only with the landscape today and would be practicing
Defensive Coding, at the expense of potential break down the line if
mongodb or pymongo change.

I lean more toward verbose with descriptive comments, but think it's more
personal choice rather than there being an absolute best choice. Nice thing
with being verbose is that 4 years down the line, if some one new is
copy/pasting code, it's a little hint to them to learn about the option and
why it may be important.





On Fri, Nov 13, 2015 at 7:54 AM, Nathaniel Kofalt <notifications@github.com>
wrote:

> Nice. @ryansanford <https://github.com/ryansanford> based on the above
> table, should we also add w=1?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scitran/api/pull/62#issuecomment-156436055>.
>



-- 

*Ryan Sanford* | *Dev Ops Engineer*





O: 612.223.7359 X2167

ryansanford@flywheel.io

www.flywheel.io

PR #65, opened 2015-11-17
Serve API with PasteScript

andynemzek:
Install PasteScript into your python environment and then execute a command like this:
```
/path/to/venv/bin/paster serve /path/to/api/dev.ini --reload
```
This will launch the API with PasteScript and activate automatic reloading.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 *.egg-info
 /persistent
 /runtime
+bootstrap.json
--- a/api/wsgi.py
+++ b/api/wsgi.py
@@ -1,35 +1,38 @@
 # vim: filetype=python
 
 import os
-import logging
+import sys
 
 os.environ['PYTHON_EGG_CACHE'] = '/tmp/python_egg_cache'
 os.umask(0o022)
 
 def app_factory(global_config, **config):
-    from util import log
-    logging.getLogger('scitran.data').setLevel(logging.WARNING) # silence scitran.data logging
-    log_level = config['log_level'].upper() if 'log_level' in config else 'INFO'
-    log.setLevel(getattr(logging, log_level))
-
-    config = process_config(config, log)
-
+    args = config
     from . import mongo
-    mongo.configure_db(config['db_uri'], config['site_id'], config['site_name'], config['api_uri'])
+    mongo.configure_db(args['db_uri'])
 
     # imports delayed after mongo has been fully initialized
     from . import api
+    from . import config
 
-    api.app.config = config
+    log = config.log
+    args = process_config(args, log)
+    config.set_log_level(log, args['log_level'])
 
-    try:
-        import newrelic.agent
-        newrelic.agent.initialize('../../newrelic.ini')
-        log.info('New Relic detected and loaded. Monitoring enabled.')
-    except ImportError:
-        log.info('New Relic not detected. Monitoring disabled.')
-    except newrelic.api.exceptions.ConfigurationError:
-        log.warn('New Relic detected but configuration was not valid. Please ensure newrelic.ini is present. Monitoring disabled.')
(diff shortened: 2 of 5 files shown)


andynemzek:
NOTE: if we move forward with PasteScript, we should probably do some extra work to consolidate the old bin/api.wsgi with the new wsgi.py

kofalt:
Is this a purely additive change?
Is it purely for development?
Will it influence / replace uwsgi?

andynemzek:
Yes, yes and maybe.  This is not for prod use.  We may want to consolidate how we're launching uwsgi and paster at some point.  Otherwise, the goal would be for uwsgi to be unaffected.

andynemzek:
Also of note here, you can use a debugger with paster.  For example, you can put this in your code:
```
import pdb; pdb.set_trace()
```
And you will be dropped into the debugger in the terminal that is running paster

gsfr:
Very cool, @andynemzek!

@rentzso and I will exercise reloading and debugging functionality tomorrow. We'll also look into consolidating with the existing files.

gsfr:
I'm sorry, but I have been swamped and won't get to this until next week.

andynemzek:
Ok, just a reminder, these are purely additive changes, so they can technically be merged whenever.

kofalt:
@josschne wanna take a look?

gsfr:
Got `https` working with @andynemzek yesterday, simply by adding `ssl_pem` param under `[app:main]`.

andynemzek:
Hi All, where are we at with this?  I will be needing this shortly (or something like it) to proceed with building the dev env.

kofalt:
I've got no complaints.

gsfr:
@andynemzek: Per our offline discussion, please rebase onto master and expand with bash wrapper. I'll help with converting the API's config system to environment variables. Will seek buy-in from @kofalt later today.

andynemzek:
@gsfr I created a minimal bash script to run the API and mongo.  I think this script is relatively solid.  I also modified the wsgi.py file to work against the changes in master...this however, needs more work.  Again, it's additive, so we could merge and work on it incrementally if necessary.  Let me know what you think.

gsfr:
Very cool!

I've made a few changes to the bash script, mainly created variables for 'runtime' and 'persistent' and made them optionally settable as command line arguments.

We are still missing the ability to optionally bootstrap users and/or data. I'm not sure about the exact desired behavior here, with respect to idempotence, or instance.

For reference:
```
bin/bootstrap.py users mongodb://localhost:9001/scitran bootstrap.json
bin/bootstrap.py data --copy mongodb://localhost:9001/scitran ../testdata persistent/data
```

andynemzek:
Thanks for making those changes.  One thing to note about variable directory names...we obviously can't change the .gitignore so the dirs would appear at a `git status` as files to add to the repo.  If they indicate a directory outside the repo I guess that's fine.  

I will be checking out the bootstrapping shortly.

andynemzek:
@gsfr see what you think about this.  Users are bootstrapped every time cause it's a fast operation.  Data is bootstrapped only if the persistent/data dir does not exist.

andynemzek:
@gsfr can we pull this in to master?  It's needed for the dev env work.

kofalt:
This LGTM!

It's a purely additive change, so let's just take initiative and merge this.
Open to further proposals in future PRs!

gsfr:
Sorry for being so slow. The only thing I would have changed is the unconditional user and group bootstrapping. While user bootstrapping is pretty idempotent, group permissions simply get clobbered, which can be annoying. Also, it's not exactly fast for longer user and group lists. I'd only bootstrap if `mongo.lock` does not exist. Will follow up with another PR.

PR #67, opened 2015-11-23
Refactoring uploads

rentzso:
- fix the reaper route that now uses the new data model.
- fix bugs on non privileged uploads
- define new config module
- remove unused code


.gitignore
```diff
@@ -2,3 +2,5 @@
 .*sw[op]
 *.py[co]
 *.egg-info
+persistent
+runtime
```
kofalt: Please prefix any rules possible with a `/` prefix.
That will require they match in the top directory (relative to the gitgnore file).

It's a good way to make sure you don't accidentally match something else later.

test/runtests.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-../../../live.sh cmd PYTHONPATH=code/api:code/data nosetests -vv --exe --collect-only code/api/test/$1
\ No newline at end of file
+../../../live.sh cmd PYTHONPATH=code/api:code/data nosetests -vv --exe code/api/test/$1
```
josschne: I just caught this. :) Glad you saw it too. Tests weren't actually running!

rentzso:
merged after testing with @gsfr

PR #68, opened 2015-11-25
Tweak bash scripts

kofalt:
A few changes:

* Use a compatible interpreter reference. Among other things, `sh` is symlinked to `bash` on OSX, which generally leads to irritation on other platforms.
* Always run relative to script path.
* Set -e mode: not very necessary ATM, but causes a script to exit on error (rather than blindly continue).


gsfr:
LGTM. Feel free to merge.

PR #70, opened 2015-11-25
Bids symlink

rentzso:
These changes allow the creation of a tarball with relative symlinks to files instead of the actual content.

### DIFF
--- a/api/core.py
+++ b/api/core.py
@@ -364,6 +364,8 @@ class Core(base.RequestHandler):
         targets = []
         # FIXME: check permissions of everything
         projects = []
+        if len(req_spec['nodes']) != 1:
+            self.abort(400, 'bids downloads are limited to single dataset downloads')
         for item in req_spec['nodes']:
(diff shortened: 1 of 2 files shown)


kofalt:
Can you explain the motivation for this change?
I'm not sure I get exactly what this does or why.

josschne:
Ditto.

gsfr:
[BIDS](http://bids.neuroimaging.io) is an alternative data hierarchy we support for [OpenfMRI](https://openfmri.org).

The motivation for this enhancement is to avoid very large downloads for data processing in shared filesystem HPC scenarios by allowing the download of relative-path symlinks instead.

OpenfMRI is currently tailing `aloha`, but planning to switch to `master` at some point, hence #71. (Looks like that's only the symlinks and not yet BIDS-style downloads, though.)

If need be, this PR can be held back until we cut `aloha-2`, but I don't currently see it interfering.

PR #71, opened 2015-11-26
Downloads

rentzso:
reduce number of queries in project downloads
add symlinks download
fix batch download after datamodel refactoring

### DIFF
--- a/api/core.py
+++ b/api/core.py
@@ -11,7 +11,6 @@ import markdown
 import cStringIO
 import validators
 
-from .dao import reaperutil
 from . import base
 from . import files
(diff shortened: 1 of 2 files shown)


gsfr:
@rentzso: Do you want to get this merged now or do you want to add the BIDS stuff first?

Also, there are some repeated imports in `core.py`.

rentzso:
this commit is independent from bids support. I would like to merge this before and then when it is ready, I will open a PR for BIDS support. 

I will now check the repeated imports in `core.py`

gsfr:
:+1:

PR #72, opened 2015-11-30
Allow for custom newrelic.ini path

gsfr:
This consolidates all new relic code into one place and removes the hardcoded path to the ini file.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -129,10 +129,6 @@ def dispatcher(router, request, response):
         response.headers['Content-Type'] = 'application/json; charset=utf-8'
 
 
-try:
-    import newrelic.agent
-    app = newrelic.agent.WSGIApplicationWrapper(webapp2.WSGIApplication(routes))
-except ImportError:
-    app = webapp2.WSGIApplication(routes)
+app = webapp2.WSGIApplication(routes)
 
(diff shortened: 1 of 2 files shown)


kofalt:
I like that this now only tries to import when requested.

What's our plan for compatibility as infra changes?
I think merging this would further the breakage I'm experiencing in the other PR.

After that's figured out, LGTM.

gsfr:
Are you ok with considering sci/sci aloha-only?

kofalt:
I would like to maintain it until we have the new infra in place - among other things, the ruleset work is based off master. Without that, I don't have a coherent way to do things like cycle instance state & test the engine, etc.

Keeping sci/sci in sync seems less work than putting something else together. Alternatively, if you're comfortable with no new work progressing on master (including ruleset or engine changes) until the new infra is ready, then we can merge this.

gsfr:
Obviously, we can't have you blocked. Can you please help me with a matching sci/sci PR?

kofalt:
Sure, I think both PRs are pretty straightforward (this and #73).
I'll see if I can whip up compatible sci/sci PRs for each.

kofalt:
PRs ready and master compatible; merging.

PR #73, opened 2015-11-30
Discussion: better config

kofalt:
1. Adds the ability to load a config file, replacing most flags (ref [sci/sci branch of same name](https://github.com/scitran/scitran/compare/config-improvements))

1. Adds `api/config` and `api/config.js` endpoints to ship config (one JSON, one JS snippet)

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -35,9 +35,8 @@ class RequestHandler(webapp2.RequestHandler):
         drone_secret = self.request.headers.get('X-SciTran-Auth', None)
 
         site_id = config.site_id()
-        # FIXME: another crash on master
-        # if site_id is None:
-        #    self.abort(503, 'Database not initialized')
+        if site_id is None:
+           self.abort(503, 'Database not initialized')
 
         # User (oAuth) authentication
         if access_token and self.app.config['oauth2_id_endpoint']:
--- a/api/base.py
+++ b/api/base.py
@@ -36,7 +36,7 @@ class RequestHandler(webapp2.RequestHandler):
 
         site_id = config.site_id()
         if site_id is None:
-           self.abort(503, 'Database not initialized')
+            self.abort(503, 'Database not initialized')
 
         # User (oAuth) authentication
         if access_token and self.app.config['oauth2_id_endpoint']:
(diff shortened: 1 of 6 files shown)


gsfr:
I like where you are going with this. I added two commits:

The first one cleans up the config endpoints a bit and exposes the new db config document.

The second one updates `bootstrap.py`: renames the `dbinit` target to `configure` and adds a mandatory `client_id` param. This is probably not how we want to do it, but it immediately let's us expose the `client_id` via the config endpoint.

The next step could be to switch all config handling over to the toml-like dot notation. Let's discuss tomorrow.

It would be great to have the option for a config file plus command line arguments.

In any case, we need a way to ensure that all required config params are present at runtime.

We also need a decision of what to store in the DB and what to keep in memory only. Ideally, that split would align with what's provided to `bootstrap.py configure` and what's provided at runtime. In case a config file is used, the same file should work for both invocations.

kofalt:
Your changes look good; I've got a small tweak I'd like to add later this morning.

bin/api.wsgi
```diff
@@ -67,6 +97,8 @@ if not api.app.config['drone_secret']:
 if not os.path.exists(api.app.config['data_path']):
     os.makedirs(api.app.config['data_path'])
 
+# FIXME hard-coded off to fix crash at @gsfr's direction
+centralclient_enabled = False
```
kofalt: I don't currently know how to avoid this crash; if you like we can screen share later to debug.
Alternatively, let me know if you're fine with this merging.

api/base.py
```diff
@@ -35,8 +35,9 @@ def __init__(self, request=None, response=None):
         drone_secret = self.request.headers.get('X-SciTran-Auth', None)
 
         site_id = config.site_id()
-        if site_id is None:
-            self.abort(503, 'Database not initialized')
+        # FIXME: another crash on master
+        # if site_id is None:
+        #    self.abort(503, 'Database not initialized')
```
kofalt: Same here, I think.

bin/api.wsgi
```diff
@@ -21,6 +22,8 @@ ap.add_argument('--central_uri', help='scitran central api', default='https://sd
 ap.add_argument('--log_level', help='log level [info]', default='info')
 ap.add_argument('--drone_secret', help='shared drone secret')
 
+# REVIEW: remove most above flags?
+ap.add_argument('--config', help='path to config file')
```
kofalt: Based on our discussion I think we're leaving all / most current flags.
If that's the case I'll remove this comment.

gsfr:
Looks good to me now. Ready to merge.

PR #84, opened 2015-12-07
Job rules

kofalt:
Partial rework to make two changes:

1. A simple rule system that allows for configurable job triggers on file upload
1. *Prepare* jobs module for the possibility of multiple scientific inputs to a single job.

This change is now sufficiently large to land.

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -40,45 +40,26 @@ ALGORITHMS = [
 
 # TODO: json schema
 
-
-def create_jobs(db, container, container_type, file):
-
-    # CHANGES:
-    # This no longer takes an array, should take string container type and container object
-    # This calls rules.check_rules
-    # Some of this lookup logic moves to rules.check_rules
-    # Rethink what exact data is being stored on a job
-    # generate_formula gets smarted
-    #
-    # OLD DOCS + LOGIC FOLLOW
-
+def create_job(db, container, container_type, file_, alg_name):
     """
-    Spawn some number of queued jobs to process a file.
+    Spawn a job to process a file.
 
     Parameters
     ----------
     db: pymongo.database.Database
         Reference to the database instance
-    containers: [ tuple(string, scitran.Container) ]
-        An array of tuples, each containing a container type name, and a container object.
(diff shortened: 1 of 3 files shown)


api/rules.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    Check all rules that apply to this file, and enqueue the jobs that should be run.
+    Returns the algorithm names that were queued.
+    """
+
+    job_list = []
+
+    # Get configured rules for this project
+    project = get_project_for_container(db, container)
+    rules = project. get('rules', [])
```
josschne: Is there an extra space here? project.get

api/jobs.py
```diff
@@ -37,38 +43,39 @@ def valid_transition(from_state, to_state):
     "dcm2nii"
 ]
 
+# Holds all details of a scitran file needed to use that as an input a formula.
```
josschne: Please make this a sentence. :)

api/jobs.py
```diff
@@ -2,14 +2,20 @@
 API request handlers for process-job-handling.
 """
 
+# We shadow the standard library; this is a workaround.
+from __future__ import absolute_import
```
josschne: I don't understand the need for absolute_import here - even with the comment :)

api/jobs.py
kofalt: Fixed, subjectively; let me know.

api/rules.py
kofalt: Fixed

kofalt:
Fixed the first two comments & did a rebase.

josschne:
LGTM

api/jobs.py
kofalt: We have a python package that is unambiguously named relative to the standard library - stdlib has `collections`, we have `collections.py`. Apparently in python 3 this is less painful.

That snippet is a backport of smarter import behaviour, changing the file to actually do what we ask it to `from collections import namedtuple` rather than interpreting that as `from ./collections.py import namedtuple`.

PR #85, opened 2015-12-07
Fix keys

### LINKED ISSUE #83: Uploading data doesn't result in jobs getting created in master branch
Looks like this was caused by renaming some elements of the file document inside the acquisitions collection within mongodb.

dirty->unprocessed
filename->name
filehash->hash
filetype->type

One example of the mismatch
api.wsgi Line 136
            for c in application.db[c_type].find({'files.dirty': True}, ['files']):

kofalt:
Closes #83. Ref @rentzso @ryansanford

### DIFF
--- a/api/dao/reaperutil.py
+++ b/api/dao/reaperutil.py
@@ -27,14 +27,14 @@ class ReapedAcquisition(object):
         return None
 
     def update_file(self, fileinfo):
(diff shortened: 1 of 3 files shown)


rentzso:
LGTM

ryansanford:
I added changes that I made as well this morning.

kofalt:
Additions LGTM; let's land this for now.

PR #86, opened 2015-12-07
Sanitize filepaths

kofalt:
After the refactor, it looks like the API will accept any path sent to it & try to write.
`os.path.basename` is safe to call multiple times & this seemed like the best entry points.

Thoughts?


rentzso:
LGTM

josschne:
Can the filename have a trailing slash?

kofalt:
Dunno. @rentzso?

kofalt:
Merged for now; open to further changes regarding trailing slashes

rentzso:
there shouldn't be any trailing slashes

PR #90, opened 2015-12-08
Paste script improvements

gsfr:


run.sh
```diff
@@ -69,11 +65,17 @@ else
...
+    BOOTSTRAP_USERS=0
 else
-    echo "Creating persistence store exists at $PERSITENT_DIR/db"
+    echo "Creating database location at $PERSITENT_DIR/db"
     mkdir -p $PERSITENT_DIR/db
+    if ! [ -f "bootstrap.json" ]; then
+        echo "Cannot bootstrap users. Please create bootstrap.json from bootstrap.json.sample."
+        exit 1
+    fi
+    BOOTSTRAP_USERS=1
```
andynemzek: We might want to keep the bootstrap file check near the top...I placed it up top because it's one of the few conditions that will actually exit the script.  Notifying immediately saves the user time versus waiting for most of the script to complete and then requiring a rerun.

run.sh
```diff
@@ -85,25 +87,27 @@ pip install -U -r requirements.txt
...
 # Set python path so scripts can work
 export PYTHONPATH=.
 
 # Configure api
-bin/bootstrap.py configure mongodb://localhost/scitran local Local https://localhost:8080/api oauth_client_id
+bin/bootstrap.py configure mongodb://localhost:9001/scitran local Local https://localhost:8080/api oauth_client_id
 
 # Boostrap users
-bin/bootstrap.py users mongodb://localhost/scitran bootstrap.json
+if [ "$BOOTSTRAP_USERS" -eq "1" ]; then
```
andynemzek: Don't need quotes around either OP here.

run.sh
gsfr: Agreed. I realized the tradeoff when I made the change, but I didn't want to require a file that may not even get used. Now, I realize that I can just move the entire db check to the top and we're fine.

run.sh
gsfr: Done.

gsfr:
@andynemzek Commit landed.

andynemzek:
LGTM!

PR #69, opened 2015-11-25
Discussion: Qa hack

kofalt:
This is a quick hack to add QA as a hardcoded algorithm.
Ping @gsfr and @josschne. 

~~Currently two spots (new lines 72 & 213) where QA is intentionally disabled.
Will remove in another commit soon once I get the alg working.~~

PR #91, opened 2015-12-08
More job fixes

kofalt:
Fixes more key errors (!), ports #69 to master, and enables a QA alg rule.

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -205,7 +205,7 @@ class Jobs(base.RequestHandler):
         List all jobs. Not used by engine.
         """
         if not self.superuser_request:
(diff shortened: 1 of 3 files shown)


gsfr:
I was just able to test this. Jobs are getting created and are looking good. Don't have engine to test end-to-end.

Had to rebase to master to get superuser requests working. Was painless. Happy to force-push if desired.

Side note: HTTP status code 401 is reserved for reporting invalid oauth tokens. Please use 403 instead.

kofalt:
Rebased & ready to merge. 

Threw in 401 -> 403. We should stop putting integers everywhere and use references instead.

PR #94, opened 2015-12-08
Cache testdata download

andynemzek:
Caching test data now to speed up repeated bootstrap


kofalt:
LGTM, note that this won't expire the cache if testdata is updated.
In that regard we're at the level of functionality the old scripting provided.

PR #95, opened 2015-12-09
fix uploads of small files

### LINKED ISSUE #88: Error thrown when uploading .txt files
Uploading a text file throws this error:

```
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 1535, in __call__
    rv = self.handle_exception(request, response, e)
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 1529, in __call__
    rv = self.router.dispatch(request, response)
  File "./api/api.py", line 126, in dispatcher
    rv = router.default_dispatcher(request, response)
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 1278, in default_dispatcher
    return route.handler_adapter(request, response)
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 1102, in __call__
    return handler.dispatch()
  File "./api/base.py", line 125, in dispatch
    return super(RequestHandler, self).dispatch()
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 572, in dispatch
    return self.handle_exception(e, self.app.debug)
  File "/scitran/persistent/venv/local/lib/python2.7/site-packages/webapp2.py", line 570, in dispatch
  File "./api/handlers/listhandler.py", line 367, in post
    file_store = files.FileStore(self.request, tempdir_path, filename=kwargs.get('name'))
  File "./api/files.py", line 77, in __init__
    self.hash = self.received_file.get_hash()
AttributeError: 'cStringIO.StringO' object has no attribute 'get_hash'
```
Image (.png) files work just fine.

rentzso:
cgi.FieldStorage by default doesn't use make_file.
Only if the file is bigger than 1000 bytes the method is called.
Overriding the private method __write allows to always call make_file
independently of file size.

closes #88 

@josschne please take a look to see if this is the correct way and if there are less hacky ways of fixing this.
you can check the code of the original __write method here: https://hg.python.org/cpython/file/2.7/Lib/cgi.py#l672

### DIFF
--- a/api/files.py
+++ b/api/files.py
@@ -34,8 +34,6 @@ def getHashingFieldStorage(upload_dir, hash_alg):
     class HashingFieldStorage(cgi.FieldStorage):
         bufsize = 2**20
         def make_file(self, binary=None):
-            if not self.filename:
-                return self.file
             self.open_file = HashingFile(os.path.join(upload_dir, os.path.basename(self.filename)), hash_alg)
             return self.open_file
 
@@ -43,8 +41,11 @@ def getHashingFieldStorage(upload_dir, hash_alg):
         # _FieldStorage__file is the private variable __file of the same class
         def _FieldStorage__write(self, line):
             if self._FieldStorage__file is not None:
-                self.file = self.make_file('')
-                self.file.write(self._FieldStorage__file.getvalue())
+                # use the make_file method only if the form includes a filename
+                # e.g. do not create a file and a hash for the form metadata.
+                if self.filename:
+                    self.file = self.make_file('')
+                    self.file.write(self._FieldStorage__file.getvalue())
                 self._FieldStorage__file = None
             self.file.write(line)
 
--- a/api/files.py
+++ b/api/files.py
@@ -34,8 +34,19 @@ def getHashingFieldStorage(upload_dir, hash_alg):
     class HashingFieldStorage(cgi.FieldStorage):
         bufsize = 2**20
         def make_file(self, binary=None):
+            if not self.filename:
+                return self.file
             self.open_file = HashingFile(os.path.join(upload_dir, os.path.basename(self.filename)), hash_alg)
             return self.open_file
+
+        # override private method __write of superclass FieldStorage
+        # _FieldStorage__file is the private variable __file of the same class
+        def _FieldStorage__write(self, line):
+            if self._FieldStorage__file is not None:
+                self.file = self.make_file('')
+                self.file.write(self._FieldStorage__file.getvalue())
+                self._FieldStorage__file = None
+            self.file.write(line)
 
         def get_hash(self):
             return self.open_file.get_hash()


api/files.py
```diff
@@ -34,9 +34,20 @@ def getHashingFieldStorage(upload_dir, hash_alg):
...
         def make_file(self, binary=None):
+            if not self.filename:
+                return self.file
             self.open_file = HashingFile(os.path.join(upload_dir, os.path.basename(self.filename)), hash_alg)
             return self.open_file
 
+        # override private method __write of superclass FieldStorage
+        # _FieldStorage__file is the private variable __file of the same class
+        def _FieldStorage__write(self, line):
+            if self._FieldStorage__file is not None:
```
kofalt: Two (possibly dumb) questions:

1. Why override the _superclass_ method? Why not just create our own `__write` method? Wouldn't that have the same effect? Presumably it's called by `self.__write` which would work, yes?
1. Similarly, why wouldn't `self.__file` work? I don't think superclass members are isolated from subclasses, are they?

api/files.py
```diff
@@ -34,9 +34,20 @@ def getHashingFieldStorage(upload_dir, hash_alg):
...
+                self.file = self.make_file('')
+                self.file.write(self._FieldStorage__file.getvalue())
```
kofalt: Since `.getvalue()` returns the entire contents of the file, are you certain this branch of logic will only trigger on inconsequentially small files? How could we validate that?

api/files.py
```diff
@@ -34,9 +34,20 @@ def getHashingFieldStorage(upload_dir, hash_alg):
     class HashingFieldStorage(cgi.FieldStorage):
         bufsize = 2**20
         def make_file(self, binary=None):
+            if not self.filename:
+                return self.file
```
kofalt: What conditions trigger this logic without a filename? Maybe add a comment explaining?

api/files.py
kofalt: Also, this branch of logic doesn't reference the `line` variable, is that intended behaviour?

kofalt:
Sorry about more review comments than code :)

The change looks reasonable, but I want to make sure I understand it.
Also interested in @josschne's thoughts.

api/files.py
rentzso: I have just removed the logic on file size from the original method (actually line 47 could be removed).
the logic in lines 45-48 just ensures to call the method make_file only once

api/files.py
rentzso: 1. `self.__write` is private method and would be translated internally by Python to `self._HashingFieldStorage__write` and wouldn't have any effect
2. `self.__file` has the same behavior
3. This was my question! However check the original `__write` method and you will see that the override is safe, but maybe there is a better place for this logic to go.

api/files.py
rentzso: Only if the form field has a filename we treat that as a file.
e.g. we are not creating a file and a hash for the form metadata.

kofalt:
Cool, review comments LGTM! Pending @josschne

rentzso:
Just pushed a commit to move the `self.filename` logic in a better place and added a comment to it.

kofalt:
:+1:

josschne:
Other than comment above, LGTM

kofalt:
@josschne Do you think we should approach that idea, or just merge this for now?

rentzso:
Merging for now. Please open a new PR for further improvements.

PR #97, opened 2015-12-09
use hash storage to save/retrieve files

rentzso:


gsfr:
LGTM. Interested in feedback from @ryansanford on our CAS implementation.

Also, should we add storing the hash alg as part of the hash field in the DB.

api/files.py
```diff
@@ -71,11 +70,12 @@ def __init__(self, request, dest_path, filename=None, hash_alg='sha384'):
             self.payload = request.POST.mixed()
             self.filename = filename or self.payload.get('filename')
             self._save_body_file(dest_path, filename, hash_alg)
+        self.path = dest_path + '/' + self.filename
```
josschne: Do we want os.path.join here?

api/files.py
```diff
@@ -71,11 +70,12 @@ def __init__(self, request, dest_path, filename=None, hash_alg='sha384'):
...
         self.duration = datetime.datetime.utcnow() - start_time
         self.mimetype = util.guess_mimetype(self.filename)
         self.filetype = util.guess_filetype(self.filename, self.mimetype)
         self.hash = self.received_file.get_hash()
-        self.size = os.path.getsize(os.path.join(self.path, self.filename))
+        self.size = os.path.getsize(os.path.join(self.path))
```
josschne: I don't think os.path.join is necessary here.

bin/bootstrap.py
```diff
@@ -129,12 +129,13 @@ def data(_, args):
         with open(filepath, 'rb') as fd:
             for chunk in iter(lambda: fd.read(2**20), ''):
                 hash_.update(chunk)
-        destpath = os.path.join(args.storage_path, container.path)
+        computed_hash = hash_.hexdigest()
+        destpath = os.path.join(args.storage_path, util.path_from_hash(computed_hash))
+        dir_destpath = os.path.dirname(destpath)
```
josschne: Minor comment - is destdir more readable?

api/util.py
```diff
@@ -86,3 +86,8 @@ def guess_filetype(filepath, mimetype):
         return 'text'
     else:
         return subtype
+
+def path_from_hash(hash_):
```
josschne: Clever. Might be worth a comment.

ryansanford:
I'm still getting up to speed. Is this just changing the file naming scheme, or is this for an alternate storage path?

Either way, my comments are probably outside the scope of this change.
- aligning file names with DB index keys is essential when you start having millions of files. This way support folks can quickly find related db records from only the file name and/or path.
- encapsulating some context into the folder hierarchy is very nice. What to include depends on what your needs might be in the case the database is unavailable for an indefinite amount of time. For example, acquisition date/time might be good for us, since in a recovery scenario, we might want to recover the most recent 6 months of acquisitions first.

rentzso:
merging after addressing joe's comments

Issue #100, opened 2015-12-11
Bring back cron-like functionality


kofalt:
At the moment, [job creation](https://github.com/scitran/api/blob/3d84ddf47a4c1a05bf2fd2bc7454ed8944904f94/bin/api.wsgi#L135-L189) only runs via uwsgi cron. Without uwsgi, these tasks will not run.

#93 will reduce but not eliminate our need for cron-like functionality. The current best proposal is to use [APScheduler](https://apscheduler.readthedocs.org/en/latest/), which has several advantages over our old approach and in general looks pretty great.

gsfr:
@andynemzek is working on cron container in fly/fly.

Issue #60, opened 2015-11-06
Dev infrastructure


gsfr:
We are looking to create a simple, yet effective way to launch the scitran API for developers. Besides the API itself, the only other launch requirement is ```mongod```. A single entry point, such as ```./scitran.sh run``` and a single config file are desired. Ideally, no further dependencies are introduced. Implementation candidates are bash and python. Options for app launching and live reloading are [Reflex](https://github.com/cespare/reflex) and possibly [Paste Script](http://pythonpaste.org/script/).

Below is a starting point for discussion, intended to be refined in place before commencing work.

Requirements:
- [ ] works on
  - [x] Mac
  - [ ] Linux
  - [ ] Linux in Vagrant on Mac
- [x] live reloading of the API
- [x] API's stdout and stderr are printed
- [x] bootstrapping of users and data from arbitrary locations

Targets:
- [ ] ```prepare``` - recent bash may be all we need
- [ ] ```run``` - ```setup``` + ```bootstrap``` + ```launch```
- [ ] ```setup``` - install reflex, mongo, venv, etc.
- [ ] ```launch``` - launch API and mongo
- [ ] ```update``` - update this repo
- [ ] ```cmd``` - run arbitrary commands in the application context

Convenience targets (likely wrappers around ```cmd```):
- [ ] ```bootstrap``` - bootstrap users or data
- [ ] ```reset``` - reset database (via ```boostrap.py clean```)
- [ ] ```mongo-log``` - tail the ```mongod``` log
- [ ] ```python``` - ipython shell with ```db``` variable set
- [ ] ```lint``` - run a linter

Directory structure:
- ```scitran.sh``` or ```scitran.py``` - entry point
- ```api``` - API source code
- ```bin``` - ```api.wsgi``` and ```bootstrap.py```
- ```test``` - unit and integration tests
- ```runtime``` - disposable environment for python dependencies and mongo binaries
- ```templates``` - ```bootstrap.json```, ```config.toml```
- ```persistent``` - ```data```, ```db```

Eric89GXL:
Is this related to https://github.com/scitran/scitran/issues/2, or is it specific to the `api` repo?

kofalt:
This should probably moved to scitran/scitran.

gsfr:
This is specific to the API and should thus stay here. We may still rename this repo to ```core``` if that makes more sense.

The API is about to shed its scitran/data dependency, at which point it will be datatype-agnostic and no longer have dependencies on other scitran components.

[...]

PR #101, opened 2015-12-12
Overhaul config system

### LINKED ISSUE #96: Script organization
Determine new location/home for /run.sh and /bin/* files. /scripts could be a place to do this.

gsfr:
This PR converts the app to using only environment variables for configuration. Sensible defaults are provided, so configuration is not required for development.

All configuration has been centralized into the `config` module. The app now self-configures opportunistically when the database becomes available. The `bootstrap.py configure` step has been removed entirely.

uwsgi still works, in addition to paster:
`PYTHONPATH=. uwsgi --http :8443 --virtualenv /tmp/runtime --master --wsgi-file bin/api.wsgi`
needs `pip install uwsgi uwsgidecorators`

Open issues:
- cron functionality needs to be ported into an app server independent framework, ref #100
- when running with uwsgi, we get a mongo warning from the `job_creation` cron job: "MongoClient opened before fork. Create MongoClient"

Achieves most of the desires of #60 and closes #96.


bin/run.sh
```diff
@@ -0,0 +1,189 @@
...
+fi
+
+TESTDATA_VERSION=$(curl -sLI https://github.com/scitran/testdata/archive/master.tar.gz | grep ETag | tail -n 1 | cut -f 2 -d '"')
+if [ ! -d "$SCITRAN_PERSISTENT_PATH/testdata" ]; then
+    echo "Downloading testdata to $SCITRAN_PERSISTENT_PATH/testdata"
+    mkdir "$SCITRAN_PERSISTENT_PATH/testdata"
+    curl -L https://github.com/scitran/testdata/archive/master.tar.gz | tar xz -C "$SCITRAN_PERSISTENT_PATH/testdata" --strip-components 1
+else
+    if [ "$TESTDATA_VERSION" != "$(cat $SCITRAN_PERSISTENT_PATH/.testdata_version)" ]; then
+        echo "Testdata out of data; downloading"
```
kofalt: s/data/date

gsfr: Fixed.

ryansanford:
I'll be testing this branch, using only the env vars, and no config.toml with the docker dev env shortly.

ryansanford:
Is the double underscore for  "SCITRAN_SITE__ID" in sample.config intentional?

I did a code search for single and double underscore, and got no hits for either.

ryansanford:
Endpoint http://docker.local.flywheel.io:8080/api/config

This shouldn't expose "drone_secret" I think?

andynemzek:
@gsfr really appreciate the changes here...I think normalizing configuration to the config module is a good move.  I did notice that the paster dev.ini was removed...it doesn't look like it's versioned anywhere else.  Was that intentional?

api/config.py
```diff
@@ -13,25 +16,107 @@
 
 logging.getLogger('MARKDOWN').setLevel(logging.WARNING) # silence Markdown library
 logging.getLogger('requests').setLevel(logging.WARNING) # silence Requests library
+logging.getLogger('paste.httpserver').setLevel(logging.WARNING) # silence Paste library
```
kofalt: Is this a good idea? In dev it might be useful to see individual request logs.
Are there entirely superfluous?

bin/run.sh
```diff
@@ -0,0 +1,190 @@
...
+fi
+if [ -z "$SCITRAN_PERSISTENT_DB_PATH" ]; then
+    SCITRAN_PERSISTENT_DB_PATH="$SCITRAN_PERSISTENT_PATH/db"
+fi
+if [ -z "$SCITRAN_PERSISTENT_DB_PORT" ]; then
+    SCITRAN_PERSISTENT_DB_PORT="9001"
+fi
+if [ -z "$SCITRAN_PERSISTENT_DB_URI" ]; then
+    SCITRAN_PERSISTENT_DB_URI="mongodb://localhost:$SCITRAN_PERSISTENT_DB_PORT/scitran"
+fi
```
kofalt: These can be reduced to one-liners:

```bash
SCITRAN_RUNTIME_HOST=${SCITRAN_RUNTIME_HOST:-"127.0.0.1"}
SCITRAN_RUNTIME_PORT=${SCITRAN_RUNTIME_PORT:-"8080"}
SCITRAN_RUNTIME_PATH=${SCITRAN_RUNTIME_PATH:-"./runtime"}
SCITRAN_RUNTIME_SSL_PEM=${SCITRAN_RUNTIME_SSL_PEM:-""}
SCITRAN_RUNTIME_BOOTSTRAP=${SCITRAN_RUNTIME_BOOTSTRAP:-"bootstrap.json"}
SCITRAN_PERSISTENT_PATH=${SCITRAN_PERSISTENT_PATH:-"./persistent"}
SCITRAN_PERSISTENT_DATA_PATH=${SCITRAN_PERSISTENT_DATA_PATH:-"$SCITRAN_PERSISTENT_PATH/data"}
SCITRAN_PERSISTENT_DB_PATH=${SCITRAN_PERSISTENT_DB_PATH:-"$SCITRAN_PERSISTENT_PATH/db"}
SCITRAN_PERSISTENT_DB_PORT=${SCITRAN_PERSISTENT_DB_PORT:-"9001"}
SCITRAN_PERSISTENT_DB_URI=${SCITRAN_PERSISTENT_DB_URI:-"mongodb://localhost:$SCITRAN_PERSISTENT_DB_PORT/scitran"}
```

Question: why both default env vars, **and** a default config map in python?
If that is necessary, add comments to both sections saying that each must be kept in sync with the other.

kofalt:
Overall LGTM. A large diff but reasonable conceptual change.

gsfr:
@ryansanford: Double underscore was intentional, but silly, so removed. `drone_secret` moved to `core` to avoid exposure. Thanks for the catch!

@andynemzek: `dev.ini` is now built on the fly by `run.sh`.

api/config.py
gsfr: The paste log seemed at `debug` seems redundant and spammy to me.

bin/run.sh
gsfr: Thanks for the one-liners. Much cleaner.

The different default variables serve different purposes. All runtime and most persistent variables only exists in `run.sh`. The two that are duplicated *do not* need to be kept in sync, but are required for the api to operate, if started without external configuration. Hence, their presence in the api. Generally, the api supports being started in multiple ways, with `run.sh` being one of them, without any hard configuration requirements.

bin/run.sh
```diff
@@ -0,0 +1,170 @@
...
+echo() { builtin echo "[SCITRAN] $@"; }
+
+
+set -o allexport
+
+if [ "$#" -eq 1 ]; then
+    EXISTING_ENV=$(env | grep "SCITRAN_" | cat)
+    source "$1"
+    eval "$EXISTING_ENV"
+fi
```
gsfr: @kofalt: Any suggestions on how to make this order of precedence fiddling cleaner? The desired order of precedence is *environment, config file, defaults*.

kofalt: Well, cat's not necessary in that pipe, you can remove it.

In that order you describe, does last win or first win?

gsfr: First one wins.

The cat is necessary in case the grep gets no match.

ryansanford:
@gsfr 
No more immediate concerns from me. :+1:

bin/run.sh
kofalt: In that case, I'd leave it like this. I'm not a huge fan of trying to coalesce environment sets, or saving config in that format, but if we're going with that this method LGTM.

kofalt:
Seconded on my end, no further concerns

PR #107, opened 2015-12-15
Better support mongo replicas and variable retry params

andynemzek:


andynemzek:
This must be merged before the analogous retrydb branch on fly/fly.  The changes here keep everything functionally the same.  It simply adds additional overridable configs so we can make use of them in our various environments.

api/config.py
```diff
@@ -44,6 +44,8 @@
     },
     'persistent': {
         'db_uri': 'mongodb://localhost:9001/scitran',
+        'db_connect_timeout': 2000,
```
kofalt: Recommend making this higher, say `5000`

kofalt:
LGTM after comment

andynemzek:
@kofalt done.  Thanks!

PR #114, opened 2016-01-04
Move features in deployed branch to aloha

rentzso:
This pull request adds features that are only present in the deployed branch:
* custom filters on file type in batch download
* search on group and project name
* batch download optimizations
* small modifications needed to render the sdm ui properly

These changes are needed to complete the aloha version of sdm.

### DIFF
--- a/api/collections.py
+++ b/api/collections.py
@@ -328,6 +328,7 @@ class CollectionSessions(sessions.Sessions):
         for sess in sessions:
             sess['measurements'] = session_measurements.get(sess['_id'], None)
             sess['subject_code'] = sess.get('subject', {}).get('code', '') # FIXME when subject is pulled out of session
+            sess['attachment_count'] = len([f for f in sess.get('files', []) if f.get('flavor') == 'attachment'])
         if self.debug:
             for sess in sessions:
                 sid = str(sess['_id'])
@@ -364,6 +365,7 @@ class CollectionAcquisitions(acquisitions.Acquisitions):
         acquisitions = list(self.dbc.find(query, projection))
         for acq in acquisitions:
             acq.setdefault('timestamp', datetime.datetime.utcnow())
+            acq['attachment_count'] = len([f for f in acq.get('files', []) if f.get('flavor') == 'attachment'])
         if self.debug:
             for acq in acquisitions:
                 aid = str(acq['_id'])
--- a/api/collections.py
+++ b/api/collections.py
@@ -315,7 +315,7 @@ class CollectionSessions(sessions.Sessions):
                 {'$group': {'_id': '$session'}},
                 ])
         query = {'_id': {'$in': [ar['_id'] for ar in agg_res]}}
-        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1, 'subject.age': 1, 'subject.sex': 1}
+        projection = {'label': 1, 'subject.code': 1, 'notes': 1, 'timestamp': 1, 'timezone': 1, 'subject.age': 1, 'subject.sex': 1, 'files': 1}
         projection['permissions'] = {'$elemMatch': {'_id': self.uid, 'site': self.source_site}}
         sessions = list(self.dbc.find(query, projection)) # avoid permissions checking by not using ContainerList._get()
         session_measurements = {}
--- a/api/core.py
+++ b/api/core.py
@@ -348,7 +348,7 @@ class Core(base.RequestHandler):
                 for acq in acquisitions:
                     session = session_dict[acq['session']]
                     acq_prefix = prefix + '/' + session.get('label', 'untitled') + '/' + acq.get('label', 'untitled')
-                    total_size, file_cnt = append_targets(targets, acq, acq_prefix, total_size, file_cnt, filter_)
+                    total_size, file_cnt = _append_targets(targets, acq, acq_prefix, total_size, file_cnt, req_spec['optional'], data_path, filter_)
             elif item['level'] == 'session':
                 session = self.app.db.sessions.find_one({'_id': item_id}, ['project', 'label', 'files'])
                 project = self.app.db.projects.find_one({'_id': session['project']}, ['group', 'name'])
--- a/api/core.py
+++ b/api/core.py
@@ -340,13 +340,15 @@ class Core(base.RequestHandler):
                 prefix = '/'.join([arc_prefix, project['group'], project['name']])
                 total_size, file_cnt = _append_targets(targets, project, prefix, total_size, file_cnt, req_spec['optional'], data_path, filter_)
                 sessions = self.app.db.sessions.find({'project': item_id}, ['label', 'files'])
-                for session in sessions:
+                session_dict = {session['_id']: session for session in sessions}
+                acquisitions = self.app.db.acquisitions.find({'session': {'$in': session_dict.keys()}}, ['label', 'files', 'session'])
+                for session in session_dict.itervalues():
                     session_prefix = prefix + '/' + session.get('label', 'untitled')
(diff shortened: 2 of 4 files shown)


api/collections.py
```diff
@@ -328,6 +328,7 @@ def get(self, cid):
         for sess in sessions:
             sess['measurements'] = session_measurements.get(sess['_id'], None)
             sess['subject_code'] = sess.get('subject', {}).get('code', '') # FIXME when subject is pulled out of session
+            sess['attachment_count'] = len([f for f in sess.get('files', []) if f.get('flavor') == 'attachment'])
```
kofalt: What is this counter used for? Mainly I'm wondering which user interface features need this specific count, since we could probably count a lot of different attributes...

For example, I think you could get this client-side with a simple `_.filter(array, function(x) { return x.flavor == "attachment" }).length` if you're using underscore.

api/collections.py
```diff
@@ -364,6 +365,7 @@ def get(self, cid):
         acquisitions = list(self.dbc.find(query, projection))
         for acq in acquisitions:
             acq.setdefault('timestamp', datetime.datetime.utcnow())
+            acq['attachment_count'] = len([f for f in acq.get('files', []) if f.get('flavor') == 'attachment'])
```
kofalt: (same here)

api/core.py
```diff
@@ -101,10 +104,19 @@
...
+            if not filter_['attachments'] and f['flavor'] == 'attachment':
+                continue
+            if not filter_['dicom'] and f['filetype'] == 'dicom':
+                continue
+            if not filter_['nifti'] and f['filetype'] == 'nifti':
+                continue
+            if not filter_['montage'] and f['filetype'] == 'montage':
+                continue
+            if not filter_['other'] and f['flavor'] != 'attachment' and f['filetype'] not in ['montage', 'nifti', 'dicom']:
+                continue
```
kofalt: Help me understand this new filter logic?

api/core.py
```diff
@@ -416,7 +430,11 @@ def _archivestream(self, ticket):
         stream = cStringIO.StringIO()
         with tarfile.open(mode='w|', fileobj=stream) as archive:
             for filepath, arcpath, _ in ticket['target']:
-                yield archive.gettarinfo(filepath, arcpath).tobuf()
+                try:
+                    yield archive.gettarinfo(filepath, arcpath).tobuf()
+                except Exception as e:
```
kofalt: Would it be wiser to be more specific about the type of Exception? I'm nervous blindly discarding *any* error that arises. Maybe a comment as to why also, couldn't this result in incomplete tarballs?

api/core.py
```diff
@@ -698,8 +726,7 @@ def _parse_query_string(query_string):
         elif date_from:
             acq_query.update({'timestamp': {'$gte': date_from}})
         if description:
-            # glob style matching, whole word must exist within description
-            pass
+            acq_query.update({'description': _parse_query_string(description)})
```
kofalt: What does this new feature enable?

kofalt:
Tentatively LGTM pending review comments :)

gsfr:
LGTM.

Most of @kofalt's comments are request for clarity, which is certainly important.

The comment about tarball exceptions is critical, though, and has implications beyond aloha. Let's be sure we always get valid tarballs out, even when files are missing. Looks like that's already the case, but let's be sure and comment appropriately.

api/core.py
rentzso: it enables search on acquisition's description

api/core.py
rentzso: filter_ is expected (if not null) to be  a json object with some boolean properties:
'attachments', 'dicom', 'nifti', 'montage', 'other'
Each of the if-conditions discards files if the corresponding filter properties is false.
e.g. if the 'attachments' property is false, discard the file if it is an attachment.
The last condition on 'other' excludes all the files that are not an attachment, a dicom, a nifti or a montage

api/collections.py
rentzso: We use it to render a paperclip for each session. The paperclip is present if there are attachments.
It's possible to move this client-side, but it is probably faster to do it here.

Note also that the attachment_count is already computed for projects, sessions and acquisitions:
https://github.com/scitran/api/blob/aloha/api/containers.py#L143

api/core.py
rentzso: It is not possible to have an incomplete tarball.
If there is an error on a file, it is fully discarded.

I understand your concern on discarding exceptions, but I thought that if there is an error on a single file for any reason it is still better to send the other files in any case. Should we make a distinction between a corrupted and a missing file in the user response?

Also, we are not completely blind to those errors, as we are logging a warn for all of them.

api/collections.py
kofalt: Sounds good. It makes sense to keep things consistent, then.

api/core.py
kofalt: Hmm, makes sense. Could you place that explanation into the code as a comment?

I'm not sure where best to place it, maybe up on line 86 as a comment next to its part of the json schema.

api/core.py
kofalt: I'm mainly trying to make sure that a web user (who won't see the log) won't be in a position to get a tarball that has fewer files than desired. I should clarify that by "incomplete tarballs" I didn't mean invalid ones, just ones with fewer files.

I can see how, if some files are corrupted, you might want to be able to retrieve any nondamaged files, but that strikes me as more of a disaster-recovery situation, and not a default desired behaviour.

I'm going to claim that we should abort entirely in that scenario, because I really don't want the user walking off thinking they have a complete set of data. I would want an entirely separate set of interactions that could check the availability of data, give a list of specific problems discovered, and selectively download the recoverable data.

api/collections.py
gsfr: This has been dropped globally in master, though, which is good because the api no longer has a strict notion of attachments. It's now up to the UI to make that interpretation, using tags and metadata.

api/core.py
gsfr: @rentzso has chosen a pragmatic solution for now. Imagine a scenario, where a user encounters a missing file after having received data for many hours and instead of letting them keep what they have, we throw it all away. That would surely piss me off. Maybe placing a README file into the tarball that lists any oddities about the download could be a good compromise.

kofalt: Are there any scenarios other than scitran database or filesystem corruption (or a sysadmin deleting files) where this error is likely to occur? Maybe I'm missing something?

I feel super confused right now, because I'd never imagine that when downloading something from a data host I'd have to extract `did-it-work-lol.txt` and check to see if I randomly didn't get the files I asked for. I want to make sure I'm not fundamentally misunderstanding what this change does.

gsfr: You're right. This would only occur if something bad has happened. We're likely to have a few bugs like #115, though, so this should remain until we are much more confident in our codebase. There is also a race condition between uploads and downloads. We should think about the order of operations for uploads. Proposal: deposit file first, then create DB entries. We're already planning a cron job that looks for non-referenced files.

api/core.py
rentzso: added a short comment to that

api/core.py
rentzso: Is there any action needed here? If not, it seems to me that the pull request is good to go.

kofalt: I firmly disagree with this change, and I don't think we can claim to be a platform that features or values data integrity with it enabled. That said, because the consensus disagrees, I'm fine with this landing on aloha (not master), since everything else LGTM.

gsfr: As stated before, I agree with your sentiment, @kofalt. This logic has however been in the code for a long time, in [one form or another](https://github.com/scitran/api/blob/master/api/core.py#L160). I opened a new issue (#119) for further discussion of this topic.

PR #117, opened 2016-01-06
Convert cron jobs to endpoints and remove unprocessed flag

gsfr:
@rentzso: Could you please take a look at this, especially the 4 calls of ```rules.create_jobs()```. Ideally, this function would only be called in a single place by the file handling code. Or maybe once for user uploads and once for reaper uploads. Anyway, I just hacked it in. Please fix/improve.

Also, I have only tested this for freshly bootstrapped files, but not yet for reaper or user uploads.

I was confused by the code in ```core.Core.reaper()```. I don't think this code should need to know whether it is adding or overwriting a file. There is also an inherent race condition in first checking if a file already exists and then trying to act accordingly. Wouldn't it be better to simply try to do something and then deal with the consequences?


rentzso:
I think that the ```rules.create_jobs()```  calls are fine where you placed them.
Adding them to the file handling code seems a little weird to me: for example we can call it in the ReapedAcquisition class in reaperutil.py (https://github.com/scitran/api/blob/master/api/dao/reaperutil.py#L13) but it would add a side effect (creating a jobs) to the methods add_file and update_file that were intended to do only those single things.

As for the logic in the reaper, it seems unavoidable for now. The two mongo operations are different, if a file already exists or not. However, we may have some simplifications if we find a way of avoiding checking for identical zip files.
Identity of files will be determined only by the hash and then implicitly checked on the filename. In this case a single atomic update, without previous logic on the existence of the file in the db, will be needed.

api/core.py
```diff
@@ -123,27 +124,27 @@ def reaper(self):
...
                 metadata=file_store.metadata
             )
             container = reaperutil.create_container_hierarchy(file_store.metadata)
             f = container.find(file_store.filename)
-            created = modified = datetime.datetime.utcnow()
             target_path = os.path.join(config.get_item('persistent', 'data_path'), util.path_from_hash(fileinfo['hash']))
             if not f:
                 file_store.move_file(target_path)
                 container.add_file(fileinfo)
+                rules.create_jobs(config.db, container.acquisition, 'acquisition', fileinfo)
```
rentzso: rules is not imported

api/handlers/listhandler.py
```diff
@@ -402,4 +401,5 @@ def post(self, cont_name, list_name, **kwargs):
             except IOError as e:
                 result = keycheck(storage.exec_op)('DELETE', _id, payload=payload)
                 raise e
+            rules.create_jobs(config.db, container, cont_name, file_properties)
```
rentzso: rules is not imported

kofalt:
Wanna rebase? Merge conflict

gsfr:
Did rebase. Still largely untested.

api/jobs.py
```diff
@@ -258,6 +259,25 @@ def next(self):
 
         return result
 
+    def reap_stale(self):
```
kofalt: Given that there's still a component called "reaper", probably a good idea to avoid use of the term elsewhere.

kofalt:
LGTM, pending testing

kofalt:
Works!

Issue #123, opened 2016-01-12
API hosted by uwsgi sometimes throws errors "No replica set members found yet"


ryansanford:
# Branch
master

# Description
Error is not consistent. About half the time I start uwsgi. Believe it's related to https://jira.mongodb.org/browse/PYTHON-986, as adding connect=False to pymongo.MongoClient() in config.py I'm unable to reproduce.  

I'll create a pull request for this change.

# Environment
Mongodb is in a standalone replicaset config

# Uwsgi config

```
[uwsgi]
wsgi-file = bin/api.wsgi
chdir=code/api
pythonpath=code/data
master = True
die-on-term = True
socket = [::]:9000
processes = 4
threads = 2
```

Actual error
```
2016-01-12T18:38:53.641920306Z [pid: 15|app: 0|req: 4/27] 192.168.3.177 () {44 vars in 809 bytes} [Tue Jan 12 18:38:53 2016] GET /api/sessions => generated 6120 bytes in 47 msecs (HTTP/1.1 200) 3 headers in 113 bytes (2 switches on core 0)
2016-01-12T18:38:53.651092720Z 2016-01-12 18:38:53      scitran.api                config.py    86:INFO Initializing database
2016-01-12T18:38:54.354535093Z 2016-01-12 18:38:54             root               webapp2.py  1552:ERRO No replica set members found yet
2016-01-12T18:38:54.354623330Z Traceback (most recent call last):
2016-01-12T18:38:54.354646766Z   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1535, in __call__
2016-01-12T18:38:54.354665146Z     rv = self.handle_exception(request, response, e)
2016-01-12T18:38:54.354681000Z   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1529, in __call__
2016-01-12T18:38:54.354697060Z     rv = self.router.dispatch(request, response)
2016-01-12T18:38:54.354712743Z   File "./api/api.py", line 120, in dispatcher
2016-01-12T18:38:54.354737566Z     rv = router.default_dispatcher(request, response)
2016-01-12T18:38:54.354793196Z   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1278, in default_dispatcher
2016-01-12T18:38:54.354811500Z     return route.handler_adapter(request, response)
2016-01-12T18:38:54.354826626Z   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1101, in __call__
[...]

PR #124, opened 2016-01-12
Resolve "No replica set members found yet" errors when using mongodb …

ryansanford:
…connections when API is hosted by API.

Issue #123 

Add connect=False to pymongo.MongoClient() for multithreaded api service hosts such as uwsgi.

@rentzso @gsfr @kofalt

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -77,7 +77,8 @@ db = pymongo.MongoClient(
     __config['persistent']['db_uri'],
     j=True,
     connectTimeoutMS=__config['persistent']['db_connect_timeout'],
-    serverSelectionTimeoutMS=__config['persistent']['db_server_selection_timeout']
+    serverSelectionTimeoutMS=__config['persistent']['db_server_selection_timeout'],
+    connect=False
 ).get_default_database()
 log.debug(str(db))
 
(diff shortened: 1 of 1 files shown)


kofalt:
Mind documenting that change?
I get allergic quickly to opaque stuff like database connection settings, eg `j=True`.

Manually cross-referencing flags with database docs can help, but it'd be good to capture the intent as well for the next time we need to fiddle with the connection settings.

api/config.py
```diff
@@ -77,7 +77,8 @@
     __config['persistent']['db_uri'],
     j=True,
     connectTimeoutMS=__config['persistent']['db_connect_timeout'],
-    serverSelectionTimeoutMS=__config['persistent']['db_server_selection_timeout']
+    serverSelectionTimeoutMS=__config['persistent']['db_server_selection_timeout'],
+    connect=False
```
gsfr: Ending the last line of such lists with a comma (,) allows for later additions to be one-line changes, unlike it was the case here.

gsfr:
I agree. End-of-line comments on at least ```j``` and ```connect``` would be great.

ryansanford:
@kofalt @gsfr  
Requested changes have been integrated. Any other comments before merging?

kofalt:
LGTM

PR #128, opened 2016-01-15
Add simple job stats query

kofalt:
This adds something I've been doing manually for awhile, so may as well make it an endpoint.

Notably, this modifies a mongo aggregate result, which by default looks like this:

```json
[
    {
        "_id": "pending", 
        "count": 37
    }
]
````

Which makes for infuriating parsing, so let's clean that up:

```json
{
    "complete": 37, 
    "failed": 0, 
    "pending": 35, 
    "running": 1
}
```

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -217,6 +217,20 @@ class Jobs(base.RequestHandler):
 
         return config.db.jobs.count()
 
+    def stats(self):
+        if not self.superuser_request:
+            self.abort(403, 'Request requires superuser')
+        result = config.db.jobs.aggregate([{"$group": {"_id": "$state", "count": {"$sum": 1}}}])
+
+        # Map mongo result to a useful object
+        states = {}
+        for r in result:
+            key = r['_id']
+            val = r['count']
+            states[key] = val
+
+        return states
+
     def next(self):
         """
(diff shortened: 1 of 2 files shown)


ryansanford:
LGTM after testing.

Let's consider opening this endpoint up so superuser not required?

kofalt:
Right now, every `/jobs` endpoint requires superuser.
An argument could be made that this information isn't very sensitive; that's just what we've done so far.

I'm going to merge for now, but I'm open to another ticket or PR that relaxes this for /stats, /count, etc.

api/jobs.py
```diff
@@ -217,6 +217,25 @@ def count(self):
...
+        states = {}
+
+        # Don't randomly omit keys that are zero
+        for s in JOB_STATES:
+            states[s] = 0
+
+        for r in result:
+            key = r['_id']
+            val = r['count']
+            states[key] = val
```
gsfr: Using dictionary comprehension, lines 226-235 could be replaced by this:
```
stats = {s: 0 for s in JOB_STATES}
stats.update({r['_id']: r['count'] for r in result})
```

kofalt: Nice!

PR #131, opened 2016-01-26
include algorithm in hash

### LINKED ISSUE #77: Include hash algorithm in hash
eg `\<hash\>` should become `sha384:\<hash\>`

rentzso:
closes #77 

The change is pretty small, but it will be needed to re-hash all the data.

@ryansanford @gsfr @kofalt let me know if this can go now or if you need some time to prepare the infrastructure.

### DIFF
--- a/api/util.py
+++ b/api/util.py
@@ -88,11 +88,12 @@ def path_from_hash(hash_):
     """
     create a filepath from a hash
     e.g.
-    hash_ = 01b395a1cbc0f218
+    hash_ = v0-sha384-01b395a1cbc0f218
     will return
-    0/1/b/3/9/5/a/1/01b395a1cbc0f218
+    0/1/b/3/9/5/a/1/v0-sha384-01b395a1cbc0f218
     """
-    path = [hash_[i] for i in xrange(8)]
+    hash_version, hash_alg, actual_hash = hash_.split('-')
+    path = [actual_hash[i] for i in xrange(8)]
     path.append(hash_)
     return os.path.join(*path)
 
--- a/api/util.py
+++ b/api/util.py
@@ -93,7 +93,7 @@ def path_from_hash(hash_):
     0/1/b/3/9/5/a/1/v0-sha384-01b395a1cbc0f218
     """
     hash_version, hash_alg, actual_hash = hash_.split('-')
-    path = [actual_hash[i] for i in xrange(8)]
+    path = [hash_version, hash_alg] + [actual_hash[2 * i : 2 * i + 2] for i in range(3)]
     path.append(hash_)
     return os.path.join(*path)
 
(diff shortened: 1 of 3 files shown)


api/util.py
```diff
@@ -88,11 +88,12 @@ def path_from_hash(hash_):
     """
     create a filepath from a hash
     e.g.
-    hash_ = 01b395a1cbc0f218
+    hash_ = v0-sha384-01b395a1cbc0f218
     will return
-    0/1/b/3/9/5/a/1/01b395a1cbc0f218
+    0/1/b/3/9/5/a/1/v0-sha384-01b395a1cbc0f218
```
gsfr: This does not look right.

kofalt:
> it will be needed to re-hash all the data

This is a pretty big warning sign, and is the sort of trick you can only get away with once.
Generally, the pattern has to be:

1. Add a new method of storage
1. Operate in a mode where old files are accessed old way, new files accessed new way
1. Write scripts that slowly move old to new, while the system is running, without overloading

On the presumption that we're not nuking all aloha & pre-aloha data sets, I would assert that a script that can do the transition (even if it has to be offline) needs to be part of this PR.

gsfr:
First off, the hash does not even change here, just the storage path.

Negative on the script. Ops issue. Aloha to bali was already a major ops issue before this PR. In fact, all dicom packages need to be read at the dicom level and repackaged to account for the api not parsing the data anymore. Aloha is not even CAS-based, so rehashing is the least of our problems.

Not to say that we don't need a script, but it does not have to be in this PR.

I say this is ready to land.

api/util.py
```diff
@@ -88,11 +88,12 @@ def path_from_hash(hash_):
...
     e.g.
-    hash_ = 01b395a1cbc0f218
+    hash_ = v0-sha384-01b395a1cbc0f218
     will return
-    0/1/b/3/9/5/a/1/01b395a1cbc0f218
+    v0/sha384/01/b3/v0-sha384-01b395a1cbc0f218
     """
-    path = [hash_[i] for i in xrange(8)]
+    hash_version, hash_alg, actual_hash = hash_.split('-')
+    path = [hash_version, hash_alg] + [actual_hash[2 * i : 2 * i + 2] for i in range(2)]
```
kofalt: Regarding `[actual_hash[2 * i : 2 * i + 2] for i in range(2)]`:

I really want us to get away from inline array building.
This instance requires mental reverse-engineering to understand it's selecting two sets of two characters.

How about `[actual_hash[0:2], actual_hash[3:5]]` instead?
Or even better:

```python
first_stanza = actual_hash[0:2]
second_stanza = actual_hash[3:5]
path = [hash_version, hash_alg, first_stanza, second_stanza]
```

gsfr: List comprehension is a key language feature of Python and we most-definitely won't stop using it.

I agree that this instance is totally pointless, though, and should be replaced by your first suggestion, but corrected:

```
[actual_hash[0:2], actual_hash[2:4]]
```

One more thing: We should be using tuples instead of arrays it indicate immutability when appropriate.

rentzso: Pushed a new commit on this.
Let me know how it looks.

kofalt:
Change LGTM! Merge as desired.

PR #135, opened 2016-01-28
Multiple files upload

rentzso:
This change enables a new route api/engine to handle uploads of multiple files with metadata.


api/core.py
```diff
@@ -354,7 +416,7 @@ def download(self):
                 config.db.projects.update_one({'_id': project_id}, {'$inc': {'counter': 1}})
         else:
             req_spec = self.request.json_body
-            validator = validators.payload_from_schema_file(self, 'input/download.json')
+            validator = validators.payload_from_schema_file(self, 'download.json')
```
kofalt: Was this incorrect before?

api/dao/reaperutil.py
```diff
@@ -144,4 +157,51 @@ def create_container_hierarchy(metadata):
...
+        )
+    session_obj = None
+    if session:
+        session['modified'] = now
+        session_obj = _update_container({'_id': acquisition_obj['session']}, session, 'sessions')
+    if project:
+        project['modified'] = now
+        if not session_obj:
+            session_obj = config.db.sessions.find_one({'_id': acquisition_obj['session']})
+        _update_container({'_id': session_obj['project']}, project, 'projects')
```
kofalt: It looks like this function is called several times without error handling.
If mongo throws an exception halfway through, will the result be a partial update? How bad is that?

kofalt:
LGTM pending a small question!

api/dao/reaperutil.py
rentzso: It's simply bad.
We would need a general "transaction" for multiple mongo operations as the reaper endpoint also applies multiple updates serially.

WIredTiger,  a new mongo driver (default in 3.2) seems to provide a transaction system, http://source.wiredtiger.com/2.4.1/transactions.html

However I think that this issue is not in the scope of this PR, but we should definitely address it sooner than later. We should explore if WiredTiger is the way to go or if a home baked system would be better in term of flexibility/optimization vs development/maintenance time.

kofalt: Fair. Let's leave it for now.

PR #136, opened 2016-01-28
Job tags

kofalt:
* Provide more structure around gears, types of gears, and how they're referenced
* Launch gears as specified in the (as yet unpublished) spec, with some temporarily workarounds
* Tag jobs with their name and type for future use
* Use new upload endpoint
* Add some new gears
* Change job rules to use new gears!

Of interest to @gsfr, @rentzso, and @ryansanford.

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -169,7 +169,7 @@ def retry_job(db, j, force=False):
     Can override the attempt limit by passing force=True.
     """
 
-    if j['attempt'] < 3 or Force:
+    if j['attempt'] < 3 or force:
         job_id = queue_job(db, j['algorithm_id'], convert_to_fileinput(j['input']), attempt_n=j['attempt']+1, previous_job_id=j['_id'])
         log.info('respawned job %s as %s (attempt %d)' % (j['_id'], job_id, j['attempt']+1))
     else:
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -6,13 +6,13 @@ API request handlers for process-job-handling.
 from __future__ import absolute_import
 
 import bson
-from enum import Enum, unique
 import pymongo
 import datetime
 from collections import namedtuple
 
 from . import base
 from . import config
+from . import util
 
 log = config.log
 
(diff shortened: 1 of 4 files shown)


api/jobs.py
```diff
@@ -37,11 +38,33 @@
...
+@unique
+class Category(Enum):
+    classifier  = 'classifier' # discover metadata
+    converter   = 'converter'  # translate between formats
+    qa          = 'qa'         # quality assurance
+    analytical  = 'analytical' # general purpose
+
+    # Enum strings are prefixed by their class: "Category.classifier".
+    # This overrides that behaviour and removes the prefix.
+    def __str__(self):
```
rentzso: Can we remove this method and use Category.classifier.value, Category.converter.value, etc in lines 57-59?
It seems cleaner to me. 
You avoid the string conversion  in line 142 also.

rentzso:
LGTM just a suggestion

api/jobs.py
kofalt: Hmm, not a bad idea. I guess it depends more generally on how we'd like to use enums in the code base. 

The downside of that approach is that it wouldn't work if the enums aren't string-based, where the current approach would (if I used `self.name` in that function rather than `self.value`).

My eventual intention was that if we use enums more, we'd make a small helper class that does it for you, and let it handle the setup. Let me try that once and see how it feels.

kofalt:
So it turned out, there was a lot better way to make enums :smile: 

As proposed, I changed `self.value` to `self.name` so it'll work with non-string enums, moved it to util, and made the Category enum in a much simpler way. I left the str operator since it makes sense that you're converting from an enum to a string. Let me know if that looks better.

rentzso:
It feels a little weird that we overrides `str` when we can use `.name` and `.value` more flexibly, but I don't want to block on this. So feel free to merge.

kofalt:
Cool. We can change that in the future for sure; to me it felt more explicit that you're converting an enum to a string. Open to further discussion.

rentzso:
My point is that converting an enum to a string doesn't seem like a standard operation, except for logging.
More proper seems to get the value of an enum or its name.
Then it's perfectly fine for me to do string conversion if needed.

kofalt:
Hmm, yeah I see what you're saying. I'm a little more used to type conversion, and more specifically I'm used to enums being types rather than complex objects.

And since they *are* complex objects, we may as well use their properties. I'll probably switch things around next time I touch this code or make a new enum.

kofalt:
Hmm, yeah I see what you're saying. I'm a little more used to type conversion, and more specifically I'm used to enums being types rather than complex objects.

And since they are complex objects, we may as well use their properties. I'll probably switch things around next time I touch this code or make a new enum.

PR #138, opened 2016-01-29
disallow key updates on container sublist

### LINKED ISSUE #103: Odd response to unchanged data
TL;DR: an update request is returning 404 when it maybe should be 304. 

When changing the permissions of a user on a project things work fine the first time. If I send the same request to change the permissions a second time, ie. trying to change `{access:'rw'}` to `{access:'rw'}`, then the response is `404 Element not updated...`.

Request:
- url: `http://localhost:9000/api/projects/abc123/permissions/local/foo@bar.com`
- body: `{"access":"rw"}`

1st Response:
- `200`
- `{"modified": 1}`

2nd Response:
- `404`
- `{"code":404,"uid":"foo@bar.com","detail":"Element not updated in list permissions of container projects abc123"}`

rentzso:
For PUT requests, disallow key updates on sublists like permissions, files, etc.
This closes #103 because now `result.modified_count == 0` is not a failure.
This is not true for StringListStorage. Its specific logic has been modified accordingly.

### DIFF
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -78,13 +78,7 @@ class ListStorage(object):
         for k,v in payload.items():
             mod_elem[self.list_name + '.$.' + k] = v
         query = {'_id': _id }
-        if exclude_params is None:
-            query[self.list_name] = {'$elemMatch': query_params}
-        else:
-            query['$and'] = [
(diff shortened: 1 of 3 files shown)


rentzso:
Still need to investigate. Most likely there's no need to disable key updates.

PR #142, opened 2016-02-02
create jobs for each file received by the engine

### LINKED ISSUE #141: Engine endpoint doesn't spawn jobs
Each entry for uploaded files needs to run the rule logic for spawning new jobs:
`rules.create_jobs(db, container, container_type, file_)`
The engine endpoint doesn't do this, meaning jobs won't trigger from uploaded results :(

I would have made this small change myself, but was halted by an impedance mismatch: the rules logic expects a file map-object with `name`, `type`, and ideally `measurements` keys. I noticed the engine endpoint returns json with different keys: `'filename': k`.

I didn't want to do anything here without knowing what was up.
Maybe we can resolve this inconsistency at the same time?

rentzso:
closes #141 

I have tested it on my local machine by force restarting all the infrastructure. @kofalt can you give it a try?

### DIFF
--- a/api/core.py
+++ b/api/core.py
@@ -1,6 +1,7 @@
 import os
 import re
 import bson
+import copy
 import json
 import pytz
 import tarfile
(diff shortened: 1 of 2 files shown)


kofalt:
Rebased (normally would ask the submitter to do this; hope you don't mind!)

Can confirm this branch launches QA jobs:
`db.getCollection('jobs').count({"tags": "qa"}) --> 44 `

Functionality SGTM; code I'm less clear on.
Would appreciate @gsfr reviewing this one.

api/core.py
```diff
@@ -216,31 +217,43 @@ def engine(self):
...
                 # if the file exists we don't need to create it
                 # skip update fileinfo for files that doesn't have a path
                 if not fileinfo.get('existing') and fileinfo.get('path'):
                     del fileinfo['path']
-                    reaperutil.add_fileinfo('acquisitions', acquisition_obj['_id'], fileinfo)
-            return [{'filename': k, 'hash': v['hash'], 'size': v['size']} for k, v in file_store.files.items()]
+                    acquisition_obj = reaperutil.add_fileinfo('acquisitions', acquisition_obj['_id'], fileinfo)
+
+            for f in acquisition_obj['files']:
+                if f['name'] in file_store.files.keys():
```
gsfr: `if f['name'] in file_store.files:`

gsfr:
Did in-person review with @rentzso. Merging...

PR #145, opened 2016-02-02
Added integration test for BIDS downloads

constellates:
@gsfr @rentzso Here's the integration test to cover the BIDS upload and download functionality. It tests that the file structure is preserved for both full file downloads and symlinked downloads.

Let me know if you have any questions, concerns or requests.

Thanks

### DIFF
--- a/test/integration_tests/test_bids_download.py
+++ b/test/integration_tests/test_bids_download.py
@@ -1,0 +1,253 @@
+import requests
+import json
+import time
+import logging
+import tarfile
+import os
+from nose.tools import with_setup
+
+log = logging.getLogger(__name__)
+sh = logging.StreamHandler()
+log.addHandler(sh)
+log.setLevel(logging.INFO)
+
+
+base_url = 'http://localhost:8080/api'
+test_data = type('',(object,),{'sessions': [], 'acquisitions': []})()
+
+session = None
+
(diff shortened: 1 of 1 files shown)


gsfr:
Thanks, @constellates. We're separately setting up automated integration test infra in #146.

@rentzso: Please make sure this PR is set up to work with our new infra.

test/integration_tests/test_bids_download.py
```diff
@@ -0,0 +1,253 @@
...
+    session.params = {
+        'user': 'test@user.com',
+        'root': True
+    }
+
+    # Convert file list to a tree
+    dataset = file_list_to_tree(file_list)
+
+    # Create a group
+    test_data.group_id = 'test_group_' + str(int(time.time()))
```
rentzso: please change this line to
`test_data.group_id = 'test_group_' + str(int(time.time()*1000))`
we then avoid to use the same group for two different test.

rentzso:
@constellates the test is compliant with our infra.
It is only needed to change how the group_id is generated

constellates:
@rentzso Thanks for taking a look. I've added the group ID update.

rentzso:
this has landed on https://github.com/poldracklab/bids-core

Closing

PR #146, opened 2016-02-03
Add travis-based integration tests

rentzso:
@josschne Creating this PR to track work on enabling integration testing in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,8 @@ install:
   - brew install python
 before_script:
   - bin/install.sh --ci
-  - bin/run.sh -d
 script:
   - bin/runtests.sh unit --ci
   - bin/runtests.sh integration --ci
 after_success:
   - coveralls
-after_script:
-  - bin/run.sh -s
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,16 @@
-#See https://github.com/travis-ci/travis-ci/issues/2312 for info on why we need
-# to install python manually
-# Python isn't supported OOTB, so pick a language that makes Travis happy
-language: objective-c
-os: osx
-sudo: false
-env:
-  - SCITRAN_RUNTIME_BOOTSTRAP=test/test_bootstrap.json
+#Travis OSX currently doesn't not support docker (via virtualbox).
+# "This computer doesn't have VT-X/AMD-v enabled."
+sudo: required
+dist: trusty
 install:
-  - brew update
-  - brew install python
+  - sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
+  - sudo sh -c "echo 'deb https://apt.dockerproject.org/repo ubuntu-trusty main' > /etc/apt/sources.list.d/docker.list"
+  - sudo apt-get update -qq
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y -q docker-engine
+  - sudo curl -o /usr/local/bin/docker-compose -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m`
+  - sudo chmod +x /usr/local/bin/docker-compose
 before_script:
-  - bin/install.sh --ci
+  - sudo bin/install.sh --ci
 script:
   - bin/runtests.sh unit --ci
   - bin/runtests.sh integration --ci
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,21 @@
-language: python
+#See https://github.com/travis-ci/travis-ci/issues/2312 for info on why we need
+# to install python manually
+# Python isn't supported OOTB, so pick a language that makes Travis happy
+language: objective-c
+os: osx
 sudo: false
-python:
-- '2.7'
+env:
+  - SCITRAN_RUNTIME_BOOTSTRAP=test/test_bootstrap.json
 install:
+  - brew update
+  - brew install python
+before_script:
   - bin/install.sh --ci
+  - bin/run.sh -d
 script:
   - bin/runtests.sh unit --ci
+  - bin/runtests.sh integration --ci
 after_success:
   - coveralls
+after_script:
+  - bin/run.sh -s
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -103,7 +103,12 @@ def with_a_group_and_a_project(api_as_admin, data_builder, request, bunch):
 def data_builder(api_as_admin):
     class DataBuilder:
 
-        # This madness allows us nicer reading code by auto-generating fucntions.
+        # This function is called whenever DataBuilder.create_X() or
+        # DataBuilder.delete_X() is called (those functions don't
+        # actually exist). It returns a callable object that does the right
+        # thing based on what X is.
+        # This madness allows significantly reduced copy/paste code and more
+        # readable tests.
         def __getattr__(self, name):
             parent_attribute = {
                 'group': '_id',
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -1,0 +1,152 @@
+import os
+import time
+import json
+import pytest
+import pymongo
+import requests
+
+
+@pytest.fixture(scope="session")
+def bunch():
+    class BunchFactory:
+        def create():
+            return type('Bunch', (), {})
+        create = staticmethod(create)
+
+    return BunchFactory()
+
+
+@pytest.fixture(scope="module")
+def db():
+    mongo_path = os.environ.get('MONGO_PATH', 'mongodb://localhost:9001/scitran')
+    return pymongo.MongoClient(mongo_path).get_default_database()
+
+
+@pytest.fixture(scope="module")
+def base_url():
+    return os.environ.get('BASE_URL', 'http://localhost:8080/api')
+
+
+class RequestsAccessor(object):
+    def __init__(self, base_url, default_params):
+        self.base_url = base_url
+        self.default_params = default_params
+
+    def _get_params(self, **kwargs):
+        params = self.default_params.copy()
+        if ("params" in kwargs):
+            params.update(kwargs["params"])
+        return params
+
+    def post(self, url_path, *args, **kwargs):
+        kwargs['params'] = self._get_params(**kwargs)
+        return requests.post(self.base_url + url_path, verify=False, *args, **kwargs)
+
+    def delete(self, url_path, *args, **kwargs):
+        kwargs['params'] = self._get_params(**kwargs)
+        return requests.delete(self.base_url + url_path, verify=False, *args, **kwargs)
+
+    def get(self, url_path, *args, **kwargs):
+        kwargs['params'] = self._get_params(**kwargs)
+        return requests.get(self.base_url + url_path, verify=False, *args, **kwargs)
+
+    def put(self, url_path, *args, **kwargs):
+        kwargs['params'] = self._get_params(**kwargs)
+        return requests.put(self.base_url + url_path, verify=False, *args, **kwargs)
+
+
+@pytest.fixture(scope="module")
+def api_as_admin(base_url):
+    accessor = RequestsAccessor(base_url, {"user": "admin@user.com", "root": "true"})
+    return accessor
+
+
+@pytest.fixture(scope="module")
+def api_as_user(base_url):
+    accessor = RequestsAccessor(base_url, {"user": "admin@user.com"})
+    return accessor
+
+
+@pytest.fixture(scope="module")
+def api_accessor(base_url):
+    class RequestsAccessorWithBaseUrl(RequestsAccessor):
+        def __init__(self, user):
+            super(self.__class__, self).__init__(base_url, {"user": user})
+
+    return RequestsAccessorWithBaseUrl
+
+
+@pytest.fixture()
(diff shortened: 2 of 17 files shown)


josschne:
Ping @gsfr / @rentzso . I think we're ready to go with this. Please take a look.

Note that coverage through paster for the integration tests is a bit tricky. If we want to get those numbers we probably will need to refactor the integration tests to call the WSGI handler directly, see an example here: http://stackoverflow.com/questions/6222528/unittesting-the-webapp-requesthandler-in-gae-python

bin/run.sh
```diff
@@ -7,21 +7,54 @@ cd "$( dirname "${BASH_SOURCE[0]}" )/.."
...
-    eval "$EXISTING_ENV"
-fi
-if [ "$#" -gt 1 ]; then
-    echo "Usage: $0 [config file]"
-    exit 1
-fi
-
-
-# Default config values
+SCITRAN_CORE_INSECURE=${SCITRAN_CORE_INSECURE:-"true"}
```
gsfr: This seems like a needlessly dangerous default. Couldn't we just have Travis set the environment variable?

bin/run.sh
```diff
@@ -32,16 +65,18 @@ SCITRAN_PERSISTENT_DATA_PATH=${SCITRAN_PERSISTENT_DATA_PATH:-"$SCITRAN_PERSISTEN
...
 SCITRAN_PERSISTENT_DB_URI=${SCITRAN_PERSISTENT_DB_URI:-"mongodb://localhost:$SCITRAN_PERSISTENT_DB_PORT/scitran"}
-
 set +o allexport
 
+#Non-exported local variables
+PASTER_INI_FILE=$SCITRAN_PERSISTENT_PATH/paster.ini
+
 
-if [ -f "$SCITRAN_PERSISTENT_DB_PATH/mongod.lock" ]; then
+if [[ -f "$SCITRAN_PERSISTENT_DB_PATH/mongod.lock" ]]; then
```
gsfr: What's the reason for converting all conditionals from single bracket to double bracket?

bin/run.sh
```diff
@@ -89,66 +124,70 @@ else
...
     touch "$SCITRAN_PERSISTENT_DATA_PATH/.bootstrapped"
-fi
+  fi
 
-
-# Serve API with PasteScript
-TEMP_INI_FILE=$(mktemp -t scitran_api)
-cat << EOF > $TEMP_INI_FILE
+  # Serve API with PasteScript
+  cat << EOF > $PASTER_INI_FILE
```
gsfr: Is there a problem with using `mktemp` in Travis?

bin/run.sh
```diff
@@ -159,12 +198,32 @@ ssl_pem=$SCITRAN_RUNTIME_SSL_PEM
...
+  if [[ $STOP == "1" ]]; then
+    if paster serve --stop-daemon > /dev/null; then
+      echo "Paster killed."
+    fi
+  fi
+
+  if [[ -s "$SCITRAN_PERSISTENT_DB_PATH/mongod.lock" ]]; then
+    MONGO_PID=`cat "$SCITRAN_PERSISTENT_DB_PATH/mongod.lock"`
+    kill $MONGO_PID
+    waitfor $MONGO_PID
```
gsfr: How is this busy-loop superior to the `wait` builtin?

bin/run.sh
```diff
@@ -89,66 +124,70 @@ else
...
 
 
-# Launch mongod
-mongod --dbpath $SCITRAN_PERSISTENT_DB_PATH --smallfiles --port $SCITRAN_PERSISTENT_DB_PORT &
-MONGO_PID=$!
 
-# Set python path so scripts can work
-export PYTHONPATH=.
+###### STARTUP TASKS
+if [[ $START == "1" || $DEV == "1" ]]; then
```
gsfr: I'm no bash expert, but I think these comparisons can be done as integers with `-eq` and no quotes.

gsfr:
Thanks, @josschne.

Would using uWSGI help with the coverage issue?

Tentatively LGTM, pending clarifications or minor updates on various line comments.

@andynemzek: Would appreciate a quick review of the diff on the two .sh files.

bin/run.sh
gsfr: Never mind, I get it. We need to know the filename on the second run for stopping the daemonized process.

bin/run.sh
josschne: Yep, we could do that. I thought this would be easier for newbies - the tests fail without this setting.

bin/run.sh
josschne: Using == is easier to read in my opinion. I changed all of them to be consistent in the file.

bin/run.sh
josschne: :+1:

bin/run.sh
josschne: Builtin `wait` can only wait on child processes. Mongo and paster are no longer children when daemonized. This busy-loop handles both cases.

bin/run.sh
josschne: They can - I just thought this was easier to read.  The boolean OR has to be a "-o" as well.

josschne:
Don't think using uWSGI would help with the coverage issue. In either case we need to enable coverage within the python environment, which can be done but is non-trivial. I'd like to focus on getting unit test coverage up and use the integration tests for their functional value, rather than any coverage.

.travis.yml
```diff
@@ -1,10 +1,21 @@
-language: python
+#See https://github.com/travis-ci/travis-ci/issues/2312 for info on why we need
+# to install python manually
+# Python isn't supported OOTB, so pick a language that makes Travis happy
+language: objective-c
+os: osx
```
kofalt: Why are we testing on OSX?
At most I'd expect us to matrix build against both linux and OSX.

andynemzek:
@gsfr didn't see anything pop out at me, but run.sh has grown quite a bit since I last worked on it.

.travis.yml
```diff
@@ -1,10 +1,21 @@
...
 sudo: false
-python:
-- '2.7'
+env:
+  - SCITRAN_RUNTIME_BOOTSTRAP=test/test_bootstrap.json
 install:
+  - brew update
+  - brew install python
+before_script:
   - bin/install.sh --ci
```
kofalt: `install.sh` does not appear to take a `--ci` flag.

bin/run.sh
```diff
@@ -81,7 +116,7 @@ else
     echo "Installed Virtualenv"
 fi
```
kofalt: I would vote for removing lines 87-117, which are a set of `which || brew` commands.

This is already handled by the Travis config, and for normal users we should document rather than script IMO.

kofalt:
I'm going to stop reviewing here, as it started to become clear that much of the script's non-portability pre-dates this PR. I played with it a bit, but... for now I'll just accept that I can't run the tests this way, and maybe we worry about that (and CI on linux) after this lands.

bin/run.sh
gsfr: I see. Why don't we have `runtests.sh` set this to true, but not `run.sh` itself?

bin/run.sh
gsfr: Ok, so double bracket allows you to use double equals. Right?

bin/run.sh
gsfr: Makes sense. But the quotes still seem useless and confusing.

.travis.yml
josschne: This script assumes OSX - it uses brew. Would be nice to support multiple environments, but my understanding was we were limiting to OSX to start. Is that correct @gsfr?

.travis.yml
josschne: Not yet... :)

bin/run.sh
josschne: Yes.

.travis.yml
gsfr: That is correct. We wanted to get started with Travis asap and we had an OS X launcher.

kofalt:
@gsfr @josschne What are the remaining blockers on getting this merged? No activity in awhile... I see one unanswered question (the first thread from the top).

gsfr:
Thanks for the ping, @kofalt. Joe and I are on it. No action needed at the moment.

josschne:
So... this got more complicated at the end than I was expecting.

We initially planned to use paster running locally, on an OSX box up on the Travis cloud. We evolved that idea to instead use our Docker container in order to improve maintainability. I subsequently discovered virtualbox (and thus docker-machine) cannot be run on the Travis OSX infra. So I went back to the Linux Trusty image where I'm able to run docker-compose without issue.

Would appreciate @rentzso and @gsfr's feedback on the following:
- The integration tests now use py.test fixtures and have been heavily refactored. 
- I had to make a change to bootstrap.py to allow non-https api paths since that is the default in our Dockerfile
- I found a bug in the download handler using the tests that I fixed. Hooray!

Would appreciate @ryansanford's check on the Dockerfile change I needed to make. uWSGI is configured to close each connection after sending the response by default. This causes an issue with the requests.Session connection pooling, which on localhost is fast enough to get the next request out after uWSGI closes the connection on its side, but before requests processes the closed connection. This results in a BadStatusLine exception popping out from the bowels of the Python standard library. Adding the various keepalive options to the Dockerfile uwsgi commandline fixes this issue.

And I'd like @kofalt's once-over just cause he's awesome (and he's going to look it over regardless).

ryansanford:
@gsfr @josschne 

Looks like we could extend docker-compose.yml to include nginx to support https out of the box for this runtests entrypoint. Thoughts?

.travis.yml
```diff
@@ -1,10 +1,18 @@
...
+sudo: required
+dist: trusty
 install:
-  - bin/install.sh --ci
+  - sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
+  - sudo sh -c "echo 'deb https://apt.dockerproject.org/repo ubuntu-trusty main' > /etc/apt/sources.list.d/docker.list"
+  - sudo apt-get update -qq
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y -q docker-engine
+  - sudo curl -o /usr/local/bin/docker-compose -L https://github.com/docker/compose/releases/download/1.6.2/docker-compose-`uname -s`-`uname -m`
+  - sudo chmod +x /usr/local/bin/docker-compose
```
kofalt: In general I prefer to keep the lines of vendor-specific CI files sparse, and instead throw these lines in a script where it's easier to document, maintain and understand. `bin/ci-install-docker-ecosystem.sh`?

Not a blocker for merge.

test/integration_tests/conftest.py
```diff
@@ -0,0 +1,152 @@
...
+    fixture_data.user_1 = user_1
+    fixture_data.user_2 = user_2
+    return fixture_data
+
+
+@pytest.fixture(scope="module")
+def data_builder(api_as_admin):
+    class DataBuilder:
+
+        # This madness allows us nicer reading code by auto-generating fucntions.
```
kofalt: A comment after my own heart :) 

Mind adding another sentence about which _specific_ madness is being solved, how?

kofalt:
> And I'd like @kofalt's once-over just cause he's awesome 

:sunglasses: 

This looks pretty great. At the architecture level, I question if we really need to boot virtualbox... I would have assumed we could use Travis' docker infra and just ran our containers on it. I'm assuming you've either already been down that road or decided it wasn't worth it. Either way, this approach is A-OK since it gets us that delicious green checkmark on PRs.

The test layout is a tad hairy but makes sense; I like the fixtures. Great work!

My read on the uwsgi config is that it probably won't affect us since we're using UWSGI in fast-cgi mode in prod, so I presume keepalive settings don't apply. Could be wrong!

josschne:
@ryansanford @kofalt: Thanks for the review!

We only need to boot virtualbox on OSX (via docker-machine). The Travis CI tests are now running on Linux and assume docker is ready (Travis CI's OSX infra actually can't run virtualbox - which is where I got stuck).

Not sure on the uwsgi comments. @ryansanford thoughts? What's the best way to get the keepalive options into the mix without breaking prod?

@ryansanford I don't think http nor nginx are necessary here. I'm not sure I understood the motivation behind your comment.

test/integration_tests/conftest.py
josschne: Done

gsfr:
Thanks, @josschne. Nice work!

I would like to still understand @ryansanford's comments before merging.

Also, we may want to bring this in line with the conflicting #185 before merging.

josschne:
Ping @ryansanford. Can you provide more detail on your nginx comment?

@gsfr - Looks like integration with #185 could be a large change to this PR. Would it make sense to separate that from this PR?

gsfr:
Thanks, @josschne. After team discussion, we decided to go ahead with merging #185, which just happened. @ryansanford has committed to update this PR to fall inline with #185 as well as his latest infra.

josschne:
@gsfr - Alright. I'll consider my part on this PR complete.

kofalt:
:+1:  :smile:

ryansanford:
@gsfr Conflicts should be resolved, and travis build completes without error. Any final words before merging?

gsfr:
Thanks, @ryansanford. Let's have @rentzso and @nagem take a final look.

rentzso:
LGTM

nagem:
LGTM

PR #149, opened 2016-02-03
validate UIDs in roles and permissions

rentzso:
I am not happy with this change.
I think it should be properly fixed in the more general context of all the validations that need to call the db, but for now this hot fix should work.

@gsfr @mrDarcyMurphy @kofalt Let me know if you want it in Bali


gsfr:
Thanks, @rentzso. I don't think this is critical for bali. I'd rather have it fixed properly over the next couple of weeks. These types of temporary fixes tend to become permanent.

We should also step back and think about all the "foreign key-type" validators we might want.

rentzso:
thanks @gsfr  I will close this PR and remove the issue from Bali.

PR #150, opened 2016-02-03
Add some more job stats

kofalt:
At the request of @gsfr, added some more job stats to the stats endpoint.
Now looks like this:

```json
{
    "by-state": {
        "complete": 0, 
        "failed": 0, 
        "pending": 74, 
        "running": 0
    }, 
    "by-tag": [
        {
            "count": 37, 
            "tags": [
                "dcm_convert", 
                "converter"
            ]
        }, 
        {
            "count": 37, 
            "tags": [
                "dicom_mr_classifier", 
                "classifier"
            ]
        }
    ], 
    "permafailed": 0
}
```

### DIFF
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -16,6 +16,8 @@ from . import util
 
 log = config.log
 
+# How many times a job should be retried
+MAX_ATTEMPTS = 3
 
 JOB_STATES = [
     'pending',  # Job is queued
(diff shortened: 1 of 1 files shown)


api/jobs.py
```diff
@@ -16,6 +16,8 @@
 
 log = config.log
 
+# How many times a job should be retried
+MAX_ATTEMPTS = 3
```
gsfr: Probably a post-bali candidate for our global config.

gsfr:
LGTM. Minimal risk since the code change is limited to the stats endpoint. Merge at will.

PR #161, opened 2016-02-17
add mimetype for each file in mongo

### LINKED ISSUE #153: Type conflating with MIME type
Ref #151 about `type: "gif"` being returned on a container's file:

1. `gif` isn't a mime type (contrast `image/gif`)
1. This field is also used for scientific data types (`dicom`)

This is troublesome when trying to make UI-relevant decisions (deciding if a viewer icon should be visible, which viewer to launch, etc).

My proposal:

1. The `type` field is only used for scientific data - so, `dicom` and similar. Blank (empty string) if there isn't a scientific data type flagged on this file.
1. A new `mime` field is sent on all files (data and attachments) that has a best-guess mimetype on it. Defaults to `application/octet-stream`.

Thoughts? @mrDarcyMurphy @dpuccetti-flywheel

rentzso:
closes #153

### DIFF
--- a/api/files.py
+++ b/api/files.py
@@ -85,6 +85,7 @@ class FileStore(object):
             self.payload = request.POST.mixed()
             self.filename = filename or self.payload.get('filename')
             self._save_body_file(dest_path, filename, hash_alg)
+        self.mimetype = util.guess_mimetype(self.filename)
         self.path = os.path.join(dest_path, self.filename)
         self.duration = datetime.datetime.utcnow() - start_time
         # the hash format is:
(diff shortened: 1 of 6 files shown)


kofalt:
LGTM :+1:

gsfr:
LGTM.

Do we prefer `type` and `mimetype` over `type` and `mime`? I'm fine with either.

rentzso:
I would keep ```mimetype``` if it's fine for everybody

kofalt:
`type` and `mimetype` are fine by me. Merging

PR #162, opened 2016-02-17
Add Dockerfile

ryansanford:
This will create a generic image that has a single external dependency of a mongodb instance. Example usage is included in the README.md.

Having the Dockerfile within the scitran/core repository allows reliable cache re-use and busting.

requirements-docker.txt Is located in the docker directory to reflect the scope of it's influence.

@gsfr  @kofalt


kofalt:
LGTM :+1:

ryansanford:
I'll merge based upon Gunnar's out-of-band comments quoted below. I did rename files based upon his comments.

> Sorry to jump the gun here if you are not ready for a review yet, but I don’t want to miss this while traveling tomorrow.
> 
> Generally, LGTM.
> 
> If you don’t mind, let’s rename additional.requirements.txt to requirements_prod.txt to be in line with our other requirements files. And maybe even move it to the top level.
> 
> Also, other files in this directory use a mixture of dot, dash, and underscore separators. Would be nice to standardize. Sorry for being a stickler.
> 
> Thanks,
>    Gunnar
> 

>

PR #167, opened 2016-02-23
return JSON responses for errors

### LINKED ISSUE #59: All responses, including all errors, should be in JSON format
Single exception: ```GET /api``` will return HTML for human consumption.

For errors, emit a predictable error struct.

rentzso:
closes #59

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -134,11 +134,13 @@ routes = [
 
 
 def dispatcher(router, request, response):
-    rv = router.default_dispatcher(request, response)
-    if rv is not None:
-        response.write(json.dumps(rv, default=util.custom_json_serializer))
-        response.headers['Content-Type'] = 'application/json; charset=utf-8'
-
+    try:
+        rv = router.default_dispatcher(request, response)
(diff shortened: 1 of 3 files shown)


rentzso:
for reference I followed this to handle errors:
https://webapp-improved.appspot.com/guide/exceptions.html

and this to handle routing errros:
https://webapp-improved.appspot.com/api/webapp2.html#webapp2.Router.default_dispatcher

kofalt:
Excellent and LGTM :+1: 

One quick question, about logic that pre-dates this PR:

```python
rv = router.default_dispatcher(request, response)
if rv is not None:
     response.write(json.dumps(rv, default=util.custom_json_serializer))

```

In what scenario is rv None? Why don't we do anything when happens?

gsfr:
> In what scenario is rv None? Why don't we do anything when happens?

When `rv` is `None`, we expect the request handler to have already written the response. `Root.get()` may be the only occurrence of that.

gsfr:
:+1: Merge as desired.

PR #170, opened 2016-02-24
create proxy route to elasticsearch

### LINKED ISSUE #168: Bridge to elasticsearch
create a basic route to authenticate search requests and forward them to elasticsearch.

rentzso:
closes #168

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -16,6 +16,7 @@ from handlers import userhandler
 from handlers import grouphandler
 from handlers import containerhandler
 from handlers import collectionshandler
+from handlers import searchhandler
 
 log = config.log
 
@@ -130,6 +131,8 @@ routes = [
 
     webapp2.Route(_format(r'/api/<par_cont_name:groups>/<par_id:{group_id_re}>/<cont_name:projects>'),          containerhandler.ContainerHandler, name='cont_sublist_groups', handler_method='get_all', methods=['GET']),
     webapp2.Route(_format(r'/api/<par_cont_name:{cont_name_re}>/<par_id:{cid_re}>/<cont_name:{cont_name_re}>'), containerhandler.ContainerHandler, name='cont_sublist', handler_method='get_all', methods=['GET']),
+    webapp2.Route(_format(r'/api/search'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
+    webapp2.Route(_format(r'/api/search/<cont_name:{cont_name_re}>'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
 ]
 
 
--- a/api/api.py
+++ b/api/api.py
@@ -131,7 +131,7 @@ routes = [
 
     webapp2.Route(_format(r'/api/<par_cont_name:groups>/<par_id:{group_id_re}>/<cont_name:projects>'),          containerhandler.ContainerHandler, name='cont_sublist_groups', handler_method='get_all', methods=['GET']),
     webapp2.Route(_format(r'/api/<par_cont_name:{cont_name_re}>/<par_id:{cid_re}>/<cont_name:{cont_name_re}>'), containerhandler.ContainerHandler, name='cont_sublist', handler_method='get_all', methods=['GET']),
-    webapp2.Route(_format(r'/api/search'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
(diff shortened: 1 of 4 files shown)


api/handlers/searchhandler.py
```diff
@@ -0,0 +1,31 @@
...
+        super(SearchHandler, self).__init__(request, response)
+
+    def get(self, cont_name=None, **kwargs):
+        if self.public_request:
+            self.abort(403, 'search is available only for authenticated users')
+        size = self.get_param('size')
+        body = self.request.json_body if self.request.body else {
+          'query': {
+            'match_all': {}
+          }
```
kofalt: Am I correct in thinking this... defaults to returning all documents if no query was passed?

api/handlers/searchhandler.py
```diff
@@ -0,0 +1,31 @@
...
+        }
+        try:
+            results = es_client.search(index='scitran', doc_type=cont_name, body=body, _source=['_id'], size=size or 10)
+        except elasticsearch.exceptions.ConnectionError as e:
+            self.abort(503, 'elasticsearch is not available')
+        return results['hits']['hits']
```
kofalt: No newline at end of file

api/config.py
```diff
@@ -47,6 +47,9 @@
         'db_server_selection_timeout': '3000',
         'data_path': os.path.join(os.path.dirname(__file__), '../persistent/data'),
     },
+    'elasticsearch': {
+        'es_endpoint': 'flywheel-elasticsearch:9200'
```
kofalt: @gsfr Has expressed desire before to keep the word `flywheel` out of this code base :)
I personally don't mind either way, since it's only a default. Up to ya'll.

kofalt:
Looks like it accomplishes what's desired from the title. From earlier discussion I expected this to be a straight reverse-proxy, but this approach (using an ES client) makes sense. Did not test against our infra. Punting on broader design commentary since I don't know enough about ES.

@ryansanford Looks like there's a new key in `config.py` to set.

api/handlers/searchhandler.py
rentzso: yes that's correct

api/config.py
rentzso: if this can be configured outside of the core module, I am happy to remove it from here.

rentzso:
@kofalt this change is the first step to have ES available from the API.
A broader design discussion should happen after @lmperry @wandell input

requirements.txt
```diff
@@ -10,3 +10,4 @@ requests==2.9.1
 rfc3987==1.3.4
 webapp2==2.5.2
 WebOb==1.5.1
+elasticsearch>=1.0.0,<2.0.0
```
gsfr: Please pin this to an exact version like all other requirements.

api/handlers/searchhandler.py
gsfr: That does not seem like a good idea.

api/api.py
```diff
@@ -130,6 +131,8 @@ def _format(route):
 
     webapp2.Route(_format(r'/api/<par_cont_name:groups>/<par_id:{group_id_re}>/<cont_name:projects>'),          containerhandler.ContainerHandler, name='cont_sublist_groups', handler_method='get_all', methods=['GET']),
     webapp2.Route(_format(r'/api/<par_cont_name:{cont_name_re}>/<par_id:{cid_re}>/<cont_name:{cont_name_re}>'), containerhandler.ContainerHandler, name='cont_sublist', handler_method='get_all', methods=['GET']),
+    webapp2.Route(_format(r'/api/search'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
+    webapp2.Route(_format(r'/api/search/<cont_name:{cont_name_re}>'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
```
gsfr: Alignment is a little wonky here.

api/config.py
gsfr: Let's add `'elasticsearch_host': 'localhost:9200'`to `persistent` instead.

requirements.txt
rentzso: I was following this guidelines: http://elasticsearch-py.readthedocs.org/en/master/

gsfr:
:+1:

Issue #66, opened 2015-11-18
Reduce HTTP awareness in non handler layers


rentzso:
As a general rule, classes should fall into two distinct categories:
- Classes that have functions named after HTTP verbs, and (basically) nothing else
- Classes that don't know anything about HTTP verbs or protocol details

Rather than self.abort, raise a variety of typed exceptions. This would be a better, drop-in replacement for what self.abort does anyway. At the very top, and/or whichever layers are appropriate, catch typed exceptions and handle a JSON error return there.

gsfr:
Sounds like a good plan to me.

gsfr:
Optional for bali-2, since no external impact.

kofalt:
Renzo and I discussed the stack-trace logging in `base.py` today - one conclusion is that we can log stack traces in fewer scenarios when this is done.

Right now, a perfectly-reasonable non-200 is a stack trace in the log, and we can't easily get rid of it because there are instances of `self.abort(5` in the code ATM. After this ticket is fixed we can be more taciturn :)

gsfr:
- [ ] create explicit request context map, stored on `self`
- [ ] never pass `self` (`RequestHandler`)
- [ ] never pass entire context map
- [ ] replace all `self.abort()` calls with `raise`
  - [ ] define `SciTranException` and inherit from it
  - [ ] create meaningful exception hierarchy using multiple inheritance using `SciTranException` and `webapp2.HTTPException`
  - [ ] catch all exceptions at the dispatcher level

nagem:
As reworking of our error handling is done, we should keep this request from #243 in mind: 

>Here's one: better error messages. While building a feature I ended up trying to add a user to a group that was already in the group. The response was:
>
>HTTPNotFound: Element not added in list roles of container groups foo
>
>404 seems like the wrong error code for that, and it would've been more helpful if it told me why the request was rejected not merely that it was (ie. "User is already in that group").

PR #171, opened 2016-02-25
Stop using handler.abort in validators

gsfr:
raise exceptions instead (caught by `base.RequestHandler.handle_exception`)

partially addresses #66

### DIFF
--- a/api/validators.py
+++ b/api/validators.py
@@ -1,11 +1,16 @@
 import os
 import copy
-import webapp2
 import jsonschema
 
 from . import config
 
 log = config.log
+
+class InputValidationException(Exception):
+    pass
+
+class DBValidationException(Exception):
+    pass
 
 # following https://github.com/Julian/jsonschema/issues/98
 # json schema files are expected to be in the schemas folder relative to this module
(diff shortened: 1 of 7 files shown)


gsfr:
LGTM, btw.

api/validators.py
```diff
@@ -105,10 +103,10 @@ def g(payload, method):
             try:
                 _validate_json(payload, _schema, resolver_input)
             except jsonschema.ValidationError as e:
-                handler.abort(400, str(e))
+                raise webapp2.exc.HTTPBadRequest(str(e))
```
gsfr: Not having this be `webapp2`-aware would actually be nicer, wouldn't it?

rentzso: yeah probably the best thing to do is to catch the ValidationError in the handler calling the method

gsfr: I'd prefer a little more encapsulation. Your code shouldn't expose its internals, such as its use of `jsonschema`. You should rather define a `ValidationException` so that I can then say `except validators.ValidationException`.

rentzso: Just pushed a new commit. Added two exceptions to distinguish when the exception is critical(mongo schema validations).

gsfr:
:+1:

kofalt:
LGTM, merge as desired

Issue #159, opened 2016-02-11
Upload unification & upgrades


gsfr:
Edited by @kofalt - itemized list:
1. [x] Lay initial groundwork for upload unification. (@kofalt)
2. [x] EnginePlacer is written but needs to be rigged up and connected to the endpoint. (@nagem)
3. [x] LabelPlacer and (new) UidPlacer should replace most logic currently in upload.reaper and upload.uploader. (@rentzso)
4. [x] After that is complete, consider unifying the LabelPlacer & UidPlacer. (@gsfr)
5. [x] Generate closures for on-the-fly processing; performance optimization. (@kofalt)
6. [x]  Once all the above is complete, consider input schema unification(@gsfr)
7. [x]  Once all the above is complete, consider Placer naming changes (@kofalt)

kofalt:
Attached: some workshop notes from a discussion last week.
Might not make a ton of sense if you weren't there, sorry :)

```
Use cases
- Engine
- Connector / Reaper
- User
- User with a set of dicoms

Features
                                                | Engine | Reaper | User | User-dicom
Upload (may) target a specific container        | Y                  Y        Y
Upload to different parts of the hierarchy      | Y         Y                  
Create hierarchy on demand                      |           Y                  
Send metadata with each file                    | Y         Y        Y        Y
Take a set of files (dicom) and produce zipfile |                             Y

Proposal
- Multi-file
- Unify the metadata
- Based on auth, we decide if you're allowed
	- If you're not a reaper, reject if hierarchy isn't there
	- 

/api/containers/3/files <-- endpoint

Ordered proposal:
3. Multi-file (multipart form uploads)
4. Optionally consumes metadata (required in the reaper case)

Later:
1. Targetted
2. General   /api/upload

Code sample:
# Updates a single container
# Moving the file
# Update / create the db
class PreparedFileUpdate(object):
​
    def __init__(self, cont_name):
        pass
​
    def add_query_condition(self, condition):
        pass
​
    def add_update_sections(self, update_sections):
        pass
​
    def new_files(self, files, force=False):
        pass
​
    def exec_update(self):
        pass
​
    def check_target(self):
        pass
```

kofalt:
[...]

PR #178, opened 2016-03-01
Retrofit of upload logic

kofalt:
This constitutes a significant overhaul of file upload processing into logic that can be deduplicated, made consistent, and eventually optimized.

Two of four upload scenarios are covered by this change: targeted (files to a single container), and a new mode packfile (files to a single zip archive). Connector and engine uploads have not been modified, but clear groundwork is laid for that transition.

Relates #159.

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -166,9 +166,10 @@ class PackfilePlacer(Placer):
         self.s_label = self.metadata['session']['label']
         self.a_label = self.metadata['acquisition']['label']
 
-        # Get project permissions
+        # Get project info that we need later
         project = config.db['projects'].find_one({ '_id': bson.ObjectId(self.p_id)})
         self.permissions = project.get('permissions', {})
+        self.g_id = project['group']
 
         # If a timestamp was provided, use that for zip files. Otherwise use a set date.
         # Normally we'd use epoch, but zips cannot support years older than 1980, so let's use that instead.
@@ -239,7 +240,8 @@ class PackfilePlacer(Placer):
         # Get or create a session based on the hierarchy and provided labels.
         s = {
             'project': bson.ObjectId(self.p_id),
-            'label': self.s_label
+            'label': self.s_label,
+            'group': self.g_id
         }
 
         # self.permissions
--- a/api/placer.py
+++ b/api/placer.py
@@ -166,6 +166,10 @@ class PackfilePlacer(Placer):
         self.s_label = self.metadata['session']['label']
         self.a_label = self.metadata['acquisition']['label']
 
+        # Get project permissions
+        project = config.db['projects'].find_one({ '_id': bson.ObjectId(self.p_id)})
+        self.permissions = project.get('permissions', {})
+
         # If a timestamp was provided, use that for zip files. Otherwise use a set date.
         # Normally we'd use epoch, but zips cannot support years older than 1980, so let's use that instead.
         # Then, given the ISO string, convert it to an epoch integer.
@@ -238,11 +242,17 @@ class PackfilePlacer(Placer):
             'label': self.s_label
         }
 
+        # self.permissions
+
         # Add the subject if one was provided
         new_s = copy.deepcopy(s)
         subject = self.metadata['session'].get('subject')
         if subject is not None:
             new_s['subject'] = subject
+        new_s = util.mongo_dict(new_s)
+
+        # Permissions should always be an exact copy
+        new_s['permissions'] = self.permissions
 
         session = config.db['session' + 's'].find_one_and_update(s, {
                 '$set': new_s
@@ -257,8 +267,11 @@ class PackfilePlacer(Placer):
             'label': self.a_label
         }
 
+        new_a = copy.deepcopy(fields)
+        new_a['permissions'] = self.permissions
+
         acquisition = config.db['acquisition' + 's'].find_one_and_update(fields, {
-                '$set': fields
+                '$set': new_a
             },
             upsert=True,
             return_document=pymongo.collection.ReturnDocument.AFTER
--- a/api/placer.py
+++ b/api/placer.py
@@ -1,0 +1,278 @@
+import bson
+import copy
+import datetime
+import dateutil
+import os
+import pymongo
+import zipfile
+
+from . import base
+from . import config
+from . import files
+from . import rules
+from . import tempdir as tempfile
+from . import util
+from . import validators
+from .dao import reaperutil, APIStorageException
+
+
+class Placer(object):
+    """
+    Interface for a placer, which knows how to process files and place them where they belong - on disk and database.
+    """
+
+    def __init__(self, container_type, container, id, metadata, timestamp):
+        self.container_type = container_type
+        self.container	  = container
+        self.id			 = id
+        self.metadata	   = metadata
+        self.timestamp	  = timestamp
+
+    def check(self):
+        """
+        Run any pre-processing checks. Expected to throw on error.
+        """
+        raise NotImplementedError()
+
+    def process_file_field(self, field, info):
+        """"
+        Process a single file field.
+        """
+        raise NotImplementedError()
+
+    def finalize(self):
+        """
+        Run any post-processing work. Expected to return output for the callee.
+        """
+        raise NotImplementedError()
+
+    def requireTarget(self):
+        """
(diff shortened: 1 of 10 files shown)


gsfr:
Ran out of time today. Will review tomorrow.

api/upload.py
```diff
@@ -1,18 +1,130 @@
...
+    Features:
+                                               | targeted | connector | engine | packfile
+        Must specify a target container        |     X    |           |    X   |
+        May create hierarchy on demand         |          |     X     |        |     X
+
+        May  send metadata about the files     |     X    |     X     |    X   |     X
+        MUST send metadata about the files     |          |     X     |        |     X
+
+        Creates a packfile from uploaded files |          |           |        |     X
+    """
```
gsfr: "connector" is a Flywheel marketing term. The correct scitran terminology is "reaper". Please change globally.

api/placer.py
```diff
@@ -0,0 +1,277 @@
...
+        self.p_id  = self.metadata['project']['_id']
+        self.s_label = self.metadata['session']['label']
+        self.a_label = self.metadata['acquisition']['label']
+
+        # If a timestamp was provided, use that for zip files. Otherwise use a set date.
+        # Normally we'd use epoch, but zips cannot support years older than 1980, so let's use that instead.
+        # Then, given the ISO string, convert it to an epoch integer.
+        minimum = 'Tue Jan  1 00:00:00 1980'
+        stamp   = self.metadata['acquisition'].get('timestamp', minimum)
+        self.ziptime = int(dateutil.parser.parse(stamp).strftime('%s'))
```
gsfr: `minimum = datetime.datetime(1980, 1, 1).isoformat()`

gsfr:
Looks good overall. Little confused by the two commits.

I know we've talked about various optimization opportunities, but how usable is the current implementation? Is it usable for a [realistic dataset](https://cni.stanford.edu/nims/file?id=195048&filename=6255_7_2_dicoms.tgz) on a realistic storage system (not an idle SSD)?

@rentzso: Please also take a look. If you see obvious de-dup opportunities, especially with your existing code, please note them in another ticket.

Maybe we could all get together on Friday to discuss next steps. Optimization, de-dup, schema unification.

kofalt:
> Little confused by the two commits.

Yeah, that was meant to be squashed. I'll leave it until merge to avoid disturbing the review threads.

> I know we've talked about various optimization opportunities, but how usable is the current implementation? Is it usable for a realistic dataset on a realistic storage system (not an idle SSD)?

Not a clue; I think the easiest way to test is going to be in the browser.  Alternatively, generating a very long curl command. This version doesn't allow repeated form fields - accessing a form field name that was repeated returns a list, and I don't handle that case - so it's not as easy as `file@a, file@b`, etc. I've passed that URL to Dan for testing.

gsfr:
@dpuccetti: Please let us know what you find?

api/placer.py
kofalt: Nice, fixed.

api/upload.py
kofalt: Fixed.

api/upload.py
```diff
@@ -1,18 +1,130 @@
...
+    # or "file1", "file2", etc (if multiple). Following this convention is probably a good idea.
+    # Here, we accept any
+    file_fields = filter(lambda x: form[x].filename is not None, form)
+
+    # TODO: Change schemas to enabled targeted uploads of more than one file.
+    # Ref docs from placer.TargetedPlacer for details.
+    if strategy == Strategy.targeted and len(file_fields) > 1:
+        raise Exception("Targeted uploads can only send one file")
+
+    for field in file_fields:
```
rentzso: Are we processing all the file information in the metadata?
It should be possible in the engine to send metadata about a file without sending the binary.

api/util.py
```diff
@@ -79,6 +79,24 @@ def guess_mimetype(filepath):
...
+    Best-effort attempt to remove blatantly poor characters from a string before turning into a filename.
+
+    Happily stolen from the internet.
+    http://stackoverflow.com/a/7406369
+    """
+
+    keepcharacters = (' ','.','_')
+    return "".join([c for c in value if c.isalnum() or c in keepcharacters]).rstrip()
+
+def obj_from_map(map):
```
rentzso: map is a reserved word in python. You can use map_

api/upload.py
kofalt: That's a great point.

The two use cases hooked up in this PR, Targeted and Packfile, will work as designed - they will always send files. When I hook up Engine - ref #159 - the metadata processing should probably be in `EnginePlacer.check()` or `EnginePlacer.check()` such that metadata processing happens even if no files were uploaded.

api/util.py
kofalt: Fixed

kofalt:
All comments addressed; rebased and squashed. Ready for merge.

api/upload.py
rentzso: Sounds good. Just thinking that a more appropriate name for `Placer.check` could be `Placer.initialize`.

rentzso:
LGTM
Please consider renaming `Placer.check` to `Placer.initialize`

gsfr:
Shouldn't the `check()` code simply go into `__init__()`?

kofalt:
I chose not to do that because I need control over when that function is called, which might become important to accomplish a few things in #159. This reduces the likelihood that I'll have to break the interface while Renzo is working on things, etc.

api/placer.py
```diff
@@ -0,0 +1,278 @@
...
+
+            # OPPORTUNITY: packfile endpoint could be extended someday to take additional metadata.
+            'instrument': None,
+            'measurements': [],
+            'tags': [],
+            'metadata': {}
+        }
+
+        # Get or create a session based on the hierarchy and provided labels.
+        s = {
```
rentzso: just noticing that the packfile placer is not setting the group on the session. This is a required field as from the [datamodel](https://github.com/scitran/core/wiki/Data-Model,-v2#session-extends-container).

api/placer.py
```diff
@@ -0,0 +1,291 @@
...
+from .dao import reaperutil, APIStorageException
+
+
+class Placer(object):
+    """
+    Interface for a placer, which knows how to process files and place them where they belong - on disk and database.
+    """
+
+    def __init__(self, container_type, container, id, metadata, timestamp):
+        self.container_type = container_type
```
rentzso: Something we need to consider in the Placer structure is also that there could be multiple target container for a single upload.
@kofalt I will try to address this in the LabelPlacer development, but let me know what you think about this.

api/placer.py
kofalt: Wait... okay help me reconcile that with the schemas.
The [input session schema](https://github.com/scitran/core/blob/master/api/schemas/input/session.json) has no group key, but the [mongo session schema](https://github.com/scitran/core/blob/master/api/schemas/mongo/session.json) does?

Are you asking that I grab the group ID from the project object and place it on the session object? Should I also do that to the acquisition?

api/placer.py
kofalt: So far, I've only been using `container_type, container, id` for a single target, and leaving it blank when there are multiple hierarchical interactions (such as the PackfilePlacer).

I think this needs to be addressed across the entire codebase: we don't have a consistent way to describe a container (which may or may not exist in memory at the moment), and when we establish that perhaps a placer takes an array of targets, or something.

rentzso: sounds good. Let's re-evaluate this later.

api/placer.py
rentzso: yeah you need to get the group id from the project object. The input session schema doesn't have the group so that you just need the parent project to create a session inside it. I don't remember the reason why we need to store the group id in the session (Cc @gsfr) but it is a required field.

The acquisition doesn't need any other key.

kofalt:
@rentzso Check out 4fdbb93, which adds group id to the upserted session.
If that looks good to you then I'll merge :)

rentzso:
LGTM!

api/placer.py
gsfr: > I don't remember the reason why we need to store the group id in the session (Cc @gsfr) but it is a required field.

The field is needed to easily filter the sessions list by group.

PR #181, opened 2016-03-03
Initialize Elasticsearch w/o connecting to Mongo

### LINKED ISSUE #180: deadlocks with pymongo possible
Deadlock can happen because the global "db" variable is used to connect to the DB via the api.wsgi entrypoint before uwsgi forks the process. 

I initially looked into this because of this message in the uwsgi log:
```
uwsgi_1          | 2016-03-03 15:38:05      scitran.api                config.py   121:DEBU Refreshing configuration from database
uwsgi_1          | /usr/local/lib/python2.7/dist-packages/pymongo/topology.py:74: UserWarning: MongoClient opened before fork. Create MongoClient with connect=False, or create client after forking. See PyMongo's documentation for details: http://api.mongodb.org/python/current/faq.html#using-pymongo-with-multiprocessing>
uwsgi_1          |   "MongoClient opened before fork. Create MongoClient "
```

@rentzso and @gsfr 
Besides avoiding the config.get_item() call in this case, can a unit test be created to expose similar regressions in the future? Maybe api.app_factory() could be moved to its own file to reduce the surface area of global includes?

Deadlock Background here:
http://api.mongodb.org/python/current/faq.html#using-pymongo-with-multiprocessing

Offending line:
https://github.com/scitran/core/blob/master/api/handlers/searchhandler.py#L10

Stack trace:
uwsgi_1          | Traceback (most recent call last):
uwsgi_1          |   File "bin/api.wsgi", line 3, in <module>
uwsgi_1          |     from api import api
uwsgi_1          |   File "./api/api.py", line 19, in <module>
uwsgi_1          |     from handlers import searchhandler
uwsgi_1          |   File "./api/handlers/searchhandler.py", line 10, in <module>
uwsgi_1          |     es_client = elasticsearch.Elasticsearch([config.get_item('persistent', 'elasticsearch_host')])
uwsgi_1          |   File "./api/config.py", line 139, in get_item

gsfr:
closes #180 

@ryansanford, @rentzso: Please take a look.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -3,6 +3,7 @@ import copy
 import logging
 import pymongo
 import datetime
+import elasticsearch
 
 
 logging.basicConfig(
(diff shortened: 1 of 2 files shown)


rentzso:
LGTM

api/config.py
```diff
@@ -84,22 +85,22 @@
 ).get_default_database()
 log.debug(str(db))
 
+es = elasticsearch.Elasticsearch([__config['persistent']['elasticsearch_host']])
```
kofalt: Does this basically hard-code the elasticsearch host, and never pull it from the database?

gsfr: Yes. A more flexible approach would be to determine the es host in `SearchHandler.get()`. Seems like the right thing to do. Any objections?

kofalt: Eh, I'm fine either way I guess.
Do we have a list somewhere of which config fields are pushed to the database but never read?

gsfr: Not really, but all variables -- except `db` -- that are used outside of the config module are periodically updated with the latest DB values. Arguably, `es` is similar to `db` in that a socket connection is established or at least configured. If we really wanted to refresh at runtime, we should at least detect if the config params have actually changed. Come to think of it, it seems like overkill right now. Will discuss with @rentzso on Monday, but inclined to leave as is.

kofalt: SGTM!

ryansanford:
@gsfr @rentzso Tested deadlock branch, and LGTM

PR #185, opened 2016-03-08
Remove data bootstrapping; add folder reaping

gsfr:
This retires data bootstrapping in favor of folder reaping.

@ryansanford: Please and update the fly/fly dev infra accordingly. Note that `bootstrap.py` no longer parses environment variables in favor of command line arguments. See `run.sh` for invocation reference.

There is still a branch reference in `requirements_dev.txt` that needs to be removed before merging. Requires scitran/reaper#17.

Also, due to changes to `bootstrap.py`, this conflicts with #146.


ryansanford:
@gsfr,
Working on updating fly/fly now. Change there will be backwards compatible.

PR #193, opened 2016-03-11
Add origin tracking to files

### LINKED ISSUE #163: Record file origin
[This proposed change](https://github.com/scitran/core/wiki/Data-Model,-v2/_compare/e66de283ad7fb1c297ec1391967282c3ef861b36...ff2d25bd8dec3f1022c74fb4e2bf778f65c86fc2) to the data model adds an `origin` field to each file, which identifies the type and ID of an uploading actor (a user or device).

This ticket tracks implementing that change.

kofalt:
All files now have an `origin` map, with a `type` string (device, user, job, or unknown) and an `id` field pointing to the corresponding document. If a file was stored before this change, the type returns unknown.

In addition, routes that return an object, or objects, with a `files` array, can now take an optional `join=origin` request parameter. This will add a `join-origin` key to each object, containing a context map of the various origin documents. Use this to prevent N calls to various other endpoints. 

Example:

```js
// on the file object
 "origin": {
      "type": "user",  // can be user | device | job | unknown
      "id": "nathanielkofalt@invenshure.com"
    }
```

Example of a `join-origin`:

```js
{
    "files": [ 
        { "origin": { "type":"user", "id":"foo" }}
        // ... 
    ],

    "join-origin": {
        "users":    {"foo": obj, "bar": obj },
        "devices":  {"foo": obj},
        "jobs":     {"foo": obj}
    }
}
```

Along the way, the job generator had to gain additional insight to throw job IDs onto upload URLs, the Placer gained an origin field for placing files, not-Placer uploads had origin hacked into them, and the base request init now knows a bit more about the requester.

Closes #163.

### DIFF
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -370,7 +370,7 @@ class FileListHandler(ListHandler):
         container, permchecker, storage, mongo_validator, payload_validator, keycheck = self._initialize_request(cont_name_plural, list_name, _id)
         permchecker(noop)('POST', _id=_id)
 
-        return upload.process_upload(self.request, upload.Strategy.targeted, cont_name, _id, self.origin)
+        return upload.process_upload(self.request, upload.Strategy.targeted, container_type=cont_name, id=_id, origin=self.origin)
 
     def packfile(self, cont_name, **kwargs):
         _id = kwargs.pop('cid')
@@ -396,4 +396,4 @@ class FileListHandler(ListHandler):
             else:
                 raise Exception('Not authorized')
 
-        return upload.process_upload(self.request, upload.Strategy.packfile, None, None, self.origin)
+        return upload.process_upload(self.request, upload.Strategy.packfile, origin=self.origin)
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -370,7 +370,7 @@ class FileListHandler(ListHandler):
         container, permchecker, storage, mongo_validator, payload_validator, keycheck = self._initialize_request(cont_name_plural, list_name, _id)
         permchecker(noop)('POST', _id=_id)
 
-        return upload.process_upload(self.request, upload.Strategy.targeted, cont_name, _id)
+        return upload.process_upload(self.request, upload.Strategy.targeted, cont_name, _id, self.origin)
 
     def packfile(self, cont_name, **kwargs):
         _id = kwargs.pop('cid')
@@ -396,4 +396,4 @@ class FileListHandler(ListHandler):
             else:
                 raise Exception('Not authorized')
 
-        return upload.process_upload(self.request, upload.Strategy.packfile)
+        return upload.process_upload(self.request, upload.Strategy.packfile, None, None, self.origin)
--- a/api/jobs.py
+++ b/api/jobs.py
@@ -172,7 +172,7 @@ def retry_job(db, j, force=False):
         log.info('permanently failed job %s (after %d attempts)' % (j['_id'], j['attempt']))
 
 
-def generate_formula(algorithm_id, i):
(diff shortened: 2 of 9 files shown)


api/types.py
```diff
@@ -0,0 +1,9 @@
+from . import util
+
+# Origin represents the different methods a request can be authenticated.
+Origin = util.Enum('Origin', {
+    'user':    'user',    # An authenticated user
+    'device':  'device',  # A connected device (reaper, script, etc)
+    'job':     'job',     # Made on behalf of a job (downloading data, uploading results, etc)
+    'unknown': 'unknown', # Other or public
+})
```
kofalt: `Origin` is referenced all over the codebase, and it's not going in `base` because import cycles.

I want to start placing the other Enums and stuff (from jobs, etc) here.

nagem:
LGTM - I like the enum for origin type.

api/handlers/listhandler.py
```diff
@@ -394,4 +394,4 @@ def packfile(self, cont_name, **kwargs):
             else:
                 raise Exception('Not authorized')
 
-        return upload.process_upload(self.request, upload.Strategy.packfile)
+        return upload.process_upload(self.request, upload.Strategy.packfile, None, None, self.origin)
```
rentzso: to avoid this `None, None,` you can do `origin=self.origin`

api/jobs.py
```diff
@@ -172,7 +172,7 @@ def retry_job(db, j, force=False):
         log.info('permanently failed job %s (after %d attempts)' % (j['_id'], j['attempt']))
 
 
-def generate_formula(algorithm_id, i):
+def generate_formula(algorithm_id, i, job_id=None):
```
rentzso: would it be reasonable to put the job_id as required?

rentzso:
LGTM - just a couple of comments.

api/jobs.py
kofalt: I thought about that, but I wasn't sure I wanted to make an existing job record and id a prerequisite for generating this kind of static content.

The origin stuff will pick up on the engine as a fallback origin if no job parameter is present, so I'd like to keep this optional unless someone has a strong opinion on it.

api/handlers/listhandler.py
kofalt: That's a great point. I don't like doing stuff like this because the params inevitably change and I mess up :) Changing now.

kofalt:
Comments addressed and rebased.
I was going to call it good for merge, but when testing a packfile upload:

```
  File "webapp2.py", line 570, in dispatch
  File "./api/handlers/listhandler.py", line 399, in packfile
    return upload.process_upload(self.request, upload.Strategy.packfile, origin=self.origin)
  File "./api/upload.py", line 80, in process_upload
    placer.check()
  File "./api/placer.py", line 163, in check
    validators.validate_data(self.metadata, 'packfile.json', 'POST')
  File "./api/validators.py", line 29, in validate_data
    validator = from_schema_path(schema_url)
  File "./api/validators.py", line 101, in from_schema_path
    schema, resolver = _resolve_schema(schema_url)
  File "./api/validators.py", line 67, in _resolve_schema
    base_uri, schema_name = re.match('(.*/)(.*)', schema_url).groups()
AttributeError: 'NoneType' object has no attribute 'groups'
```

Any thoughts on this @rentzso?
I only ran into it after rebasing, so it's a good thing I did :smile: 

I'll be out for a decent chunk of today, so if you have ideas feel free to throw them on this branch.
Otherwise we can work on it later.

rentzso:
@kofalt I will check. It is related to the new validation endpoint for serving schemas.

kofalt:
New commit 6b5a42a has fixed the problem as far as I can tell. Merging.

PR #194, opened 2016-03-11
Add archiving functionality for containers

### LINKED ISSUE #187: Add session and acquisition archiving
Add functionality to allow Sessions and Acquisitions to be marked as "archived". Only return unarchived sessions and acquisitions for get_all requests unless an archived=true flag is passed in the request.

nagem:
This request was fulfilled by creating a generic way to push changes down the hierarchy tree. If a property is added to the propagated_properties list on the config in containerhandler.py, any PUT with a payload containing that property will also be applied to children of that object. For example, if a project is marked as 'archived': true, all sessions and acquisitions under that project will also be marked as archived. 

All calls to get_all on a container will only return unarchived objects unless an archived=true flag is passed in the request.

Some assumptions/decisions made:
  - direct GET requests on a specific object will return the object regardless of archived status
  - currently only allows the entire payload to be propagated values or not, throws a 400 if the payload is of mixed property types (this could be changed in the future by creating a different payload on the fly for the child objects instead)
  - only does input and mongo validation on original container type being updated (checking the entire tree each time would only check that the API writers didn't add something to the propagated values list that shouldn't logically be in there)
  - eventually permissions changes and other changes that push down the hierarchy tree can be rolled into the list
  - the PUT will still return modified: 1 when many documents are modified. 

Closes #187

### DIFF
--- a/api/dao/containerstorage.py
+++ b/api/dao/containerstorage.py
@@ -73,6 +73,8 @@ class ContainerStorage(object):
         elif self.cont_name == "sessions":
             config.db.acquisitions.update_many(
                 {'session': _id}, update)
+        else:
+            raise ValueError('changes can only be propagated from project or session level')
 
     def _delete_el(self, _id):
         if self.use_object_id:
--- a/api/dao/containerstorage.py
+++ b/api/dao/containerstorage.py
@@ -23,7 +23,8 @@ class ContainerStorage(object):
     def get_container(self, _id):
         return self._get_el(_id)
 
-    def exec_op(self, action, _id=None, payload=None, query=None, user=None, public=False, projection=None):
+    def exec_op(self, action, _id=None, payload=None, query=None, user=None, 
+                public=False, projection=None, recursive=False):
         """
         Generic method to exec an operation.
         The request is dispatched to the corresponding private methods.
@@ -36,7 +37,7 @@ class ContainerStorage(object):
         if action == 'DELETE':
             return self._delete_el(_id)
         if action == 'PUT':
-            return self._update_el(_id, payload)
+            return self._update_el(_id, payload, recursive)
         if action == 'POST':
             return self._create_el(payload)
         raise ValueError('action should be one of GET, POST, PUT, DELETE')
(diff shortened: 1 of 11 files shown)


api/dao/containerstorage.py
```diff
@@ -23,7 +23,8 @@ def __init__(self, cont_name, use_object_id = False):
     def get_container(self, _id):
         return self._get_el(_id)
 
-    def exec_op(self, action, _id=None, payload=None, query=None, user=None, public=False, projection=None):
+    def exec_op(self, action, _id=None, payload=None, query=None, user=None, 
+                public=False, projection=None, recursive=False):
```
gsfr: This reminds me that we still haven't set a character line limit. I've been casually rolling with 150 by setting my browser to that width, but also going over when convenient.

Let's pick a linter asap and start with some very lax rules.

gsfr:
LGTM, but haven't tested.

@rentzso: Could you please do a review?

api/handlers/containerhandler.py
```diff
@@ -251,6 +255,13 @@ def put(self, cont_name, **kwargs):
 
         payload = self.request.json_body
         payload_validator(payload, 'PUT')
+        # Check if any payload keys are a propogated property, ensure they all are
+        rec = False
+        if set(payload.keys()).intersection(set(self.config['propagated_properties'])):
```
rentzso: This will fail if the `self.config` doesn't have a 'propagated_properties`(and this is true for acquisitions for example).
Doing something like `self.config.get('propageted_properties', [])` will fix it.

api/dao/containerstorage.py
```diff
@@ -54,8 +55,25 @@ def _update_el(self, _id, payload):
...
+
+        if self.cont_name == "projects":
+            session_ids = [s['_id'] for s in config.db.sessions.find({'project': _id}, [])]
+            config.db.sessions.update_many(
+                {'project': _id}, update)
+            config.db.acquisitions.update_many(
+                {'session': {'$in': session_ids}}, update)
+        elif self.cont_name == "sessions":
+            config.db.acquisitions.update_many(
+                {'session': _id}, update)
```
kofalt: Add an `else` that throws an exception - right now, `_propagate_changes` will happily do nothing if `cont_name` is not in the list.

kofalt:
LGTM pending some tiny changes

PR #203, opened 2016-03-17
catch Exceptions in method dispatch in api.py

### LINKED ISSUE #59: All responses, including all errors, should be in JSON format

rentzso:
closes #59


kofalt:
LGTM!

Issue #201, opened 2016-03-16
Text and PDF files render garbled.


mrDarcyMurphy:
Images and nifti files that are presented in the modal appear to be fine (though we might be getting lucky there). Text and PDF files have begun rendering poorly in the last week or two. Not sure why. I can provide the example documents if you need them.

Screenshots:

![screen shot 2016-03-16 at 1 18 18 pm](https://cloud.githubusercontent.com/assets/21331/13823918/a94e1cfe-eb79-11e5-91e2-c70414e183b5.png)

![screen shot 2016-03-16 at 1 18 38 pm](https://cloud.githubusercontent.com/assets/21331/13823917/a94b6478-eb79-11e5-9445-6815a9f4b319.png)

nagem:
Taking a look now, the API is trying to grab the file's mimetype when the file has to mimetype. The issue is fixed in this commit (https://github.com/scitran/core/commit/71e6464638a319b5f40ae450bb81b5371ee90509), but I will still investigate why the mimetypes are missing/why this is now a problem when it did not used to be one.

gsfr:
Setting the mime type on ingest is a relatively recent update. Legacy data wouldn't have it.

@mrDarcyMurphy Have you tried re-bootstrapping?

We might also be failing to set mime type on some intake paths. Hence my desire to finish "upload unification".

@nagem Not sure empty string is the best default. Maybe `application/octet-stream` is better?

nagem:
@gsfr You are definitely right, updating.

nagem:
@gsfr After a couple minutes of investigation, Chome downloads `application/octet-stream` mimetypes while it makes an okay judgement on how to open a file when no mimetype is given. I can dig in further to the "why" if you want, too.

nagem:
After reading [this](http://stackoverflow.com/questions/1176022/unknown-file-type-mime) SO question and testing it, `application/unknown` seems to be the best option.

gsfr:
@nagem Agreed.

mrDarcyMurphy:
@gsfr I've rebuilt multiple times. Most recently was monday morning after updating docker.

gsfr:
Thanks, @mrDarcyMurphy. In that case, we should really figure out how data is getting into our system without a mime type on it.

@nagem Could you please take that on.

PR #204, opened 2016-03-17
Fix Mimetype Error

nagem:
Closes https://github.com/scitran/core/issues/201

After discussion with @rentzso, we decided to go back to using `application/octet-stream` as the default mimetype. This fix is assuring files have proper mimetypes again so anytime there is a file without a mimetype it is a bug and we should prevent the browser from deciding what to do with the file.


kofalt:
LGTM, and confirmed the problem is solved on my box.

PR #205, opened 2016-03-18
Add 'public' as a propagated property for projects

### LINKED ISSUE #121: create route to make a project public with all its sessions and acquisitions

nagem:
Note this is only for projects. Closes #121.


rentzso:
LGTM

PR #207, opened 2016-03-18
Consistency checker

### LINKED ISSUE #58: Validate UIDs in roles and permissions
At least for the local site, we should only allow new roles and permissions for actual user IDs.

rentzso:
closes #58 

created new tests in [latest-integ-tests](/scitran/core/tree/latest-integ-tests)


nagem:
LGTM.

PR #211, opened 2016-03-21
allow access to containers for all users

rentzso:
Sensitive data like subject firstname and lastname will be left off.
Also file contents won't be available.

@gsfr @nagem could you give it a look?

@dpuccetti this enables getting the name of a project for all users.


nagem:
LGTM. 

In testing, an auth'd user without access to a container could GET a specific container, but could not GET a list of all containers.

rentzso:
@nagem this is the expected behavior (double checked with @gsfr).
Merging.

PR #213, opened 2016-03-23
Explicit schema versioning and upgrades

### LINKED ISSUE #206: Explicit schema versioning and upgrades
Relates #189. Additional details are currently in team member's heads and on the [upgrade branch](https://github.com/scitran/core/tree/upgrades).

nagem:
Closes #206.

A script to check if the current db schema version is behind the required db schema version. Also allows applying incremental upgrade changes if the db schema version is behind.

The database version is stored in the field "database" in a document with _id "version" in a new collection called "version". The existing "config" collection was avoided due to existing logic allowing multiple config documents in the config collection, accessing the correct one via a `{'latest': True}` query.

An API endpoint was added that returns the stored version information.

Requirements/of note:
  - The db schema version is an int, bumped when a schema upgrade is required
  - No version document assumes a current version of 0
  - The upgrade from 0 to 1 is the addition of a version document
  - The version document is purposefully generic, allowing other version types to be added in the future
  - There are many try/except blocks to prevent any errors in database.py escaping with an exit code of 0.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -32,6 +32,11 @@ class Config(base.RequestHandler):
             ';'
         )
 
+class Version(base.RequestHandler):
+
+    def get(self):
+        return config.get_version()
+
 #regexes used in routing table:
 routing_regexes = {
     # group id regex
@@ -77,7 +82,8 @@ routes = [
         webapp2.Route(r'/sites',            centralclient.CentralClient, handler_method='sites', methods=['GET']),
         webapp2.Route(r'/register',         centralclient.CentralClient, handler_method='register', methods=['POST']),
         webapp2.Route(r'/config',           Config, methods=['GET']),
-        webapp2.Route(r'/config.js',        Config, handler_method='get_js', methods=['GET'])
+        webapp2.Route(r'/config.js',        Config, handler_method='get_js', methods=['GET']),
+        webapp2.Route(r'/version',          Version, methods=['GET']),
     ]),
     webapp2.Route(r'/api/users',            userhandler.UserHandler, handler_method='get_all', methods=['GET']),
     webapp2.Route(r'/api/users',            userhandler.UserHandler, methods=['POST']),
--- a/bin/database.py
+++ b/bin/database.py
@@ -1,13 +1,13 @@
+#!/usr/bin/env python
 import json
 
-from .. import config
+from api import config
 
 
 print 'This is the stub database maintenance script.'
 
-print 'DB uri: ' + str(config.get_item('persistent', 'db_uri')))
+print 'DB uri: ' + str(config.get_item('persistent', 'db_uri'))
 
 user = config.db.users.find_one({})
-user_j = json.dumps(user)
 
-print 'Example user from db: ' + user_j
+print 'Example user from db: {0}'.format(user)
--- a/bin/database.py
+++ b/bin/database.py
@@ -1,0 +1,13 @@
+import json
+
+from .. import config
+
+
+print 'This is the stub database maintenance script.'
+
+print 'DB uri: ' + str(config.get_item('persistent', 'db_uri')))
+
+user = config.db.users.find_one({})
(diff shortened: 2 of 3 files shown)


api/api.py
```diff
@@ -77,7 +82,8 @@ def _format(route):
         webapp2.Route(r'/sites',            centralclient.CentralClient, handler_method='sites', methods=['GET']),
         webapp2.Route(r'/register',         centralclient.CentralClient, handler_method='register', methods=['POST']),
         webapp2.Route(r'/config',           Config, methods=['GET']),
-        webapp2.Route(r'/config.js',        Config, handler_method='get_js', methods=['GET'])
+        webapp2.Route(r'/config.js',        Config, handler_method='get_js', methods=['GET']),
```
kofalt: Could you add database version in the config endpoint returns?

bin/database.py
```diff
@@ -0,0 +1,72 @@
...
+                 requirements and cannot be upgraded,
+                 perhaps because code is at lower version
+                 than the DB schema version.
+    """
+
+    db_version = get_db_version()
+    if not isinstance(db_version, int) or db_version > CURRENT_DATABASE_VERSION:
+        sys.exit(43) 
+    elif db_version < CURRENT_DATABASE_VERSION:
+        sys.exit(42)
```
kofalt: In both of these first two cases, please print some details to stderr.
In particular, we should know if the script failed because db_version wasn't integer.

```python
import sys
# ...
sys.stderr.write('spam\n')
```

bin/database.py
```diff
@@ -0,0 +1,72 @@
...
+    Upgrades db to the current schema version
+    
+    Returns (0) if upgrade is successful
+    """
+
+    db_version = get_db_version()
+    try:
+        if db_version < 1:
+            # scitran/core issue #206
+            config.db.version.insert_one({'_id': 'version', 'database': CURRENT_DATABASE_VERSION})
```
kofalt: These migrations will / should quickly become the biggest part of the script.
How about moving the `if < N; do things` logic each to their own functions, at the end of the file?

```python
    db_version = get_db_version()
    try:
        if db_version < 1:
            upgrade_to_1()
        if db_version < 2:
            upgrade_to_2()

# ...

def upgrade_to_1():
    config.db.version.insert_one({'_id': 'version', 'database': CURRENT_DATABASE_VERSION})
```

bin/database.py
```diff
@@ -0,0 +1,72 @@
...
+        elif sys.argv[1] == 'upgrade_schema':
+            upgrade_schema()
+        else:
+            print 'Unknown method name given as argv to database.py'
+            sys.exit(1)
+    else:
+        print 'No method name given as argv to database.py'
+        sys.exit(1)
+except Exception as e:
+    print e
```
kofalt: All prints in the script should be stderr, IMO, since they're all error messages.

kofalt:
LGTM pending review commentary, but I'd like to test it after everything is ready.

gsfr:
How about not using `print` statements at all, but only `logging`, which can be configured to print to one or more destinations?

kofalt:
I thought about that, but I worry that it would conflate the logs - when we run our infrastructure in detached mode, it goes to a different place that might not be immediately obvious, and I think there might be something similar in prod. I'm worried that it wouldn't be immediately obvious what the script said.

gsfr:
Hmm, I don't see the difference. You could just have logging default to stderr.

kofalt:
Oh, you meant the package? I thought you meant the actual config object, which presumably is rigged to go wherever. Or maybe I'm misunderstanding how that's all set up.

Yeah, as long as it goes to stderr that's fine by me!

gsfr:
Yup, meant the package, sorry.

nagem:
Suggest changes committed.

kofalt:
LGTM!

PR #217, opened 2016-03-26
Update to latest testdata commit

gsfr:
The latest testdata has more realistic labels, which is important for prober UI development.

There is an intentional label collision that will be handled by a future version of the folder reaper, but that's still under development. The current reaper works, but two sessions get conflated. I'd still like to merge asap.


ryansanford:
This change tests out ok for me. Travis failure was transient, re-run was successful.

LGTM.

Issue #224, opened 2016-03-30
Change of authtoken index expire time causes conflict when index is already created


nagem:
Caused by https://github.com/scitran/core/pull/218.

`OperationFailure: command SON([('createIndexes', u'authtokens'), ('indexes', [{'expireAfterSeconds': 604800, 'name': u'timestamp_1', 'key': SON([('timestamp', 1)])}])]) on namespace scitran.$cmd failed: exception: Index with name: timestamp_1 already exists with different options`

kofalt:
Haha, amazing.

![](https://cloud.githubusercontent.com/assets/265071/14155901/5f2d52dc-f689-11e5-91a9-7b8de78632ae.gif)

I've never been thrilled with document expiry being handled by the database; I'll file a new ticket to track future improvements beyond this bugfix.

nagem:
I'm currently implementing create_or_recreate_ttl_index, please let me know if you have any better ideas on how to handle this without completely changing how and where we do db initialization.

nagem:
Mongo does not allow you to update a ttl index, and they suggest dropping the index and recreating it if you need to change the ttl

kofalt:
That sounds like a good approach for now - in the ticket I just created & referenced I have a quick link to how we can get a list of indexes for a collection / database.

Most storage products have some mechanism by which you can upsert an index - lacking an obvious native feature, we'll probably end up with a few functions of the sort you're proposing.

gsfr:
Ref my [recent comment](https://github.com/scitran/core/issues/225#issuecomment-203632576) on #225. Can we just solve #212 and then drop and re-create all indexes at start up - once?

nagem:
A fix to #212 would solve this, but for now we'll merge #228 and remove that during the work for #212.

PR #228, opened 2016-03-30
Create or recreate ttl index on db intialization

nagem:
Fixes https://github.com/scitran/core/issues/224.

Mongo does not allow indexes to be modified, suggests they be dropped and recreated if something (like `expireAfterSeconds`) needs to be changed.

For reference, this is an example of what `index_information()` returns for the authtokens collection:
`{"_id_": 
	{  "ns": "scitran.authtokens", 
	   "key": [["_id", 1]], "v": 1
	}, 
 "timestamp_1": 
 	{  "key": [["timestamp", 1]], 
 	   "ns": "scitran.authtokens", 
 	   "expireAfterSeconds": 600, 
 	   "v": 1
 	}
}`

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -5,6 +5,7 @@ import logging
 import pymongo
 import datetime
 import elasticsearch
+import json
 
 
 logging.basicConfig(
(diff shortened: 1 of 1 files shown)


api/config.py
```diff
@@ -147,6 +147,21 @@
 
 assert input_schemas == expected_input_schemas, '{} is different from {}'.format(input_schemas, expected_input_schemas)
 
+def create_or_recreate_ttl_index(coll_name, index_name, ttl):
+    index_list = db[coll_name].index_information()
+    if index_list:
+        for index in index_list:
+            if index_list[index]['key'][0][0] == index_name:
```
nagem: I could be convinced to just assume indexes will be generated with the name `<index_name>_1` to get rid of this for loop and [0][0] disaster.

nagem:
There are four scenarios encasulated here, 
 - no indexes exist on this collection -> create it
 - this specific index doesn't exist on this collection -> create it
 - the index exists but has the wrong ttl -> drop it, create it
 - the index exists with the proper ttl -> do nothing

api/config.py
kofalt: Eh. I'd say document what the `[0][0]` is doing exactly, and leave it.

kofalt:
LGTM pending a comment request

PR #234, opened 2016-03-31
add 'null' to schema

### LINKED ISSUE #233: Subject's sex can't be null.
API says it's expecting `'male'` or `'female'`. [Input schema](https://github.com/scitran/core/blob/master/api/schemas/input/subject.json#L9) says `null` is ok, [mongo schema](https://github.com/scitran/core/blob/master/api/schemas/mongo/subject.json#L9) doesn't.

mrDarcyMurphy:
Fixes #233

At least, it should.


nagem:
Ah, sorry @mrDarcyMurphy! I pushed a fix to master a few minutes ago.

PR #236, opened 2016-04-03
Improve drone name extraction for UI display

gsfr:
@ryansanford Please merge prior to bug bash and reset dev instance to reflect improved bootstrapping results.

Also, the referenced reaper commit is what should be the latest bali tag in scitran/reaper.

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -185,7 +185,7 @@ class RequestHandler(webapp2.RequestHandler):
             name = self.request.headers.get('X-SciTran-Name')
 
             self.origin = {
-                'id': method + '_' + name,
+                'id': (method + '_' + name).replace(' ', '_'),
                 'type': str(Origin.device),
                 'method': method,
                 'name': name
--- a/api/base.py
+++ b/api/base.py
@@ -92,7 +92,7 @@ class RequestHandler(webapp2.RequestHandler):
             else:
                 self.superuser_request = False
 
-        self.set_origin(drone_request, drone_name.partition(' ')[2])
+        self.set_origin(drone_request, drone_name)
 
     def authenticate_user(self, access_token):
         """
@@ -175,6 +175,9 @@ class RequestHandler(webapp2.RequestHandler):
         this map serves a different purpose, and specifically matches the desired file-origin map.
         Might be a good future project to remove one or the other.
         """
+
+        if drone_name.startswith('Reaper') or drone_name.startswith('Importer'):
+            drone_type, _, drone_name = drone_name.partition(' ')
 
         if self.uid is not None:
             self.origin = {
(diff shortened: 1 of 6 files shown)


nagem:
LGTM

gsfr:
*Importing* seems to work, although the UI isn't quite right yet. Looks like the UI is currently rendering the `id` field, rather than the `name` field.

```
origin: {
    method: "importer",
    type: "device",
    id: "importer_Admin Import",
    name: "Admin Import"
}
```

Reaping also looks good, but same display problem.

```
origin: {
    method: "reaper",
    type: "device",
    id: "reaper_cnimr",
    name: "cnimr"
}
```

nagem:
I'll open a ticket on the frontend to use the new fields.

gsfr:
Looks like we're getting some errors from our cron jobs:
```
uwsgi_1          | 2016-04-07 23:58:01      scitran.api                  base.py   305:WARN None 400 X-SciTran-Method or X-SciTran-Name header missing
uwsgi_1          | [pid: 17|app: 0|req: 7/9] 172.17.0.9 () {34 vars in 397 bytes} [Thu Apr  7 23:58:01 2016] POST /api/jobs/reap => generated 84 bytes in 0 msecs (HTTP/1.1 400) 3 headers in 120 bytes (2 switches on core 1)
```

nagem:
Talked with Ryan, opening up a ticket in fly/fly.

gsfr:
Rebased

nagem:
Pending review from @gsfr or @kofalt of the DB upgrade script, this is good to go.

gsfr:
The DB updates are not atomic. That may or may not be a problem depending on how we apply them. Could our DB be modified by other processes while the updates are being applied? Ping @ryansanford.

ryansanford:
@nagem @gsfr Flywheel dev environment and production scripts stop all containers before checking and upgrading the schema. This protects the main-line use cases only.

nagem:
And unfortunately we'd need [this ticket](https://jira.mongodb.org/browse/SERVER-1243) to find and update all the file origins in the file array in one go.

api/base.py
```diff
@@ -182,9 +180,15 @@ def set_origin(self, drone_request, drone_name):
                 'id': self.uid
             }
         elif drone_request:
+
+            method = self.request.headers.get('X-SciTran-Method')
+            name = self.request.headers.get('X-SciTran-Name')
```
kofalt: Propose that we add `.strip()` to the end of each of these to remove whitespace.

kofalt:
LGTM.

Non-atomic schema upgrades is the way to go - frequently, those run without locks or durability as well :) It should be abundantly clear that serving requests in halfway-house mode has undefined behaviour.

Maybe a comment to that effect at the top of the script would be wise?

PR #232, opened 2016-03-31
UID and Label based uploads

rentzso:
works on #215

Still pending engine upload refactoring (partial work on [this branch](https://github.com/scitran/core/tree/engine-uploads)

Needs to sync with [scitran/reaper](https://github.com/scitran/reaper) before merge.

Also this will need some changes to the endpoint used for bootstrapping in fly/fly.
[This route](https://github.com/scitran/core/pull/232/files#diff-e0978ea022e5e946d30dfa3911881b5eR85) has been kept temporarily for retrocompatibility.

PR #237, opened 2016-04-04
Add stateful uploads

kofalt:
This changes the packfile endpoints to accept batches of files, staging them until they're explicitly completed. Dodges various limitations in the server leaking file descriptors, and the browser saying no to exceedingly large uploads.

Accomplished with minimal drama by implementing an intermediary placers which is very simple, and modifying the existing Packfile placer to consume the result. Does not implement orphan ticket cleanup.

### DIFF
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -4,6 +4,7 @@ import copy
 import datetime
 import dateutil
 import json
+import uuid
 
 from .. import base
 from .. import config
@@ -442,7 +443,7 @@ class FileListHandler(ListHandler):
         query = {
             'type': 'packfile',
             'project': project_id,
-            '_id': bson.ObjectId(token_id),
+            '_id': token_id,
         }
 
         # Server-Sent Events are fired in the browser in such a way that one cannot dictate their headers.
@@ -463,7 +464,7 @@ class FileListHandler(ListHandler):
 
         # Update token timestamp
         config.db['tokens'].update_one({
-            '_id': bson.ObjectId(token_id)
+            '_id': token_id,
         }, {
             '$set': {
                 'modified': datetime.datetime.utcnow()
@@ -500,6 +501,7 @@ class FileListHandler(ListHandler):
 
         # Save token for stateful uploads
         result = config.db['tokens'].insert_one({
+            '_id': str(uuid.uuid4()),
             'type': 'packfile',
             'user': self.uid,
             'project': _id,
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -3,6 +3,7 @@ import bson
 import copy
 import datetime
 import dateutil
+import json
 
 from .. import base
 from .. import config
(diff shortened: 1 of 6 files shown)


api/placer.py
```diff
@@ -208,12 +210,61 @@ def finalize(self):
...
+    def finalize(self):
+        print os.listdir(self.folder)
+
+        for path in self.paths:
+            dest = os.path.join(self.folder, os.path.basename(path))
+
+            print 'FROM ' + path
+            print 'TO   ' + dest
+            shutil.move(path, dest)
+            print '----'
```
kofalt: Debugging code can be removed.

api/placer.py
```diff
@@ -255,15 +306,27 @@ def check(self):
...
+            os.utime(p, (self.ziptime, self.ziptime))
+
+            # Place file into the zip folder we created before
+            self.zip.write(p, os.path.join(self.dir, os.path.basename(path)))
+
         self.zip.close()
 
+        print self.folder
+        print 'SAFETY'
+        # shutil.rmtree(self.folder)
```
kofalt: Disabled for obvious reasons while developing. Can be turned on and debug statements removed.

api/upload.py
```diff
@@ -184,16 +179,19 @@ def upload(self, strategy):
...
+        if not self.superuser_request:
+            self.abort(402, 'uploads must be from an authorized drone')
+
         level = self.get_param('level')
         if level is None:
             self.abort(404, 'container level is required')
+
         if level != 'acquisition':
             self.abort(404, 'engine uploads are supported only at the acquisition level')
+
```
kofalt: Some leftover diff from ~space~. Doesn't seem to be invalid so I'm leaving it ¯\_(ツ)_/¯

api/upload.py
```diff
@@ -73,16 +74,10 @@ def process_upload(request, strategy, container_type=None, id=None, origin=None)
...
     metadata = None
     if 'metadata' in form:
-        # Slight misnomer: the metadata field, if present, is sent as a normal form field, NOT a file form field.
-        metadata_file = form['metadata'].file
-        try:
-            metadata = json.loads(metadata_file.getvalue())
-        except AttributeError:
-            raise files.FileStoreException('wrong format for field "metadata"')
-
+        metadata = json.loads(form['metadata'].value)
```
kofalt: This turned out to be a more succinct way to read a text form field... it clobbered changes from #232 during a merge conflict. We could add a try/catch again if desired.

kofalt:
Fails CI:

```
        r = api_as_admin.post('/upload/label', files=data.files)
>       assert r.status_code == 400

E       assert 500 == 400
```

I'm going to claim that fixing this is blocked by #66 and that we should modify the test with a comment.

nagem:
LGTM with debug prints removed and the CI test passing, but I'd like to have @rentzso take a look at that integ test failure as these are new tests he wrote as part of yesterday's PR.

kofalt:
Agreed, I'm still not sure how to run unit tests locally and Travis doesn't show the API logs.

rentzso:
to run the tests locally:
0) stop fly/fly
1) create a python virtual env
2) activate the env and run core/bin/install.sh (edited)
3) connect to the default docker env
4) run the script core/bin/runtests.sh integration in daemonized mode
the script creates two docker containers
to get the full logs you can run:
`docker logs <uwsgi_container_id>`

@ryansanford should be able to give you some help if something is not working

api/upload.py
rentzso: Please add the try except around that line.
I guess this is what makes the test to fail.

kofalt: Oh! That makes total sense. Adding that catch & raise back.

test/integration_tests/test_uploads.py
```diff
@@ -122,6 +122,6 @@ def test_label_upload(with_group_and_file_data, api_as_admin):
 
     data.files['metadata'] = metadata
     r = api_as_admin.post('/upload/label', files=data.files)
-    assert r.status_code == 400
+    assert r.status_code == 500
```
rentzso: to fix this properly catch the right error on line 79 of `upload.py`. I am guessing it is not an AttributeError, most likely a TypeError.

gsfr:
I'd throw the cleanup under `/packfile` as well

ryansanford:
There should be docs as part of this PR that it is up to the client to call the `clean-packfiles` endpoint when using the packfile upload feature.

gsfr:
@kofalt Note the use of `uuid.uuid4()` when creating [download tickets](https://github.com/scitran/core/blob/master/api/util.py#L65-L75).

api/handlers/listhandler.py
```diff
@@ -430,7 +432,50 @@ def post(self, cont_name, list_name, **kwargs):
...
+            'project': project_id,
+            '_id': token_id,
+        }
+
+        # Server-Sent Events are fired in the browser in such a way that one cannot dictate their headers.
+        # For these endpoints, authentication must be disabled because the normal Authorization header will not be present.
+        # There are several valid ways to fix this, including switching to cookies, or allowing the inclusion of the Authorization token as a request parameter.
+        #
+        # In lieu of solutions, this workaround allows users to merely provide their token id for these endpoints.
+        # This is not ideal because mongo ObjectIds are very predictable and not exactly nonces...
```
nagem: Now that you're using uuids, want to update this comment?

api/placer.py
```diff
@@ -298,50 +399,53 @@ def finalize(self):
...
+        # Updates if existing
+        updates = {}
+        updates['permissions'] = self.permissions
+        updates['modified']    = self.timestamp
+        updates = util.mongo_dict(updates)
+
+        # Extra properties on insert
+        insert_map = copy.deepcopy(query)
+        insert_map['created'] = self.timestamp
+        insert_map.update(self.metadata['session'])
```
nagem: This will only add the session properties passed from metadata if the session does not already exist. For example, the subject object will only be set on insert. In the diff it looks like we were previously setting the subject on inserts and updates.

nagem:
LGTM, with comments.

api/placer.py
kofalt: Mongo prevents us from doing that:

```
pymongo.errors.OperationFailure: command ... on namespace scitran.$cmd failed: exception: Cannot update 'subject.code' and 'subject' at the same time
```

api/handlers/listhandler.py
kofalt: Comment updated

PR #239, opened 2016-04-06
Create restrictions on file fields

### LINKED ISSUE #210: File Input Schema - 'null' type for files
Currently when uploading files at the acquisition level, files that do not have a type set are rejected and the `folder_reaper` is halted. 

The current "fix", in place on the deployed instances, is for:
 `   "type":           { "type": "string" },`

to become:

`    "type":           { "type": ["string", "null"] },`

https://github.com/scitran/core/blob/master/api/schemas/input/file.json#L6

Other thoughts on how to handle this?

### LINKED ISSUE #231: Create restrictions on file fields
Use an enum in our input json schema to restrict values saved to type and modality fields

nagem:
Closes #231.
Closes #210.

Restrictions to type, instrument (becoming modality when `data-model-v2` is merged), and measurements working as expected for targeted upload. `EnginePlacer` also sets these fields and uses the `file.json` schema so although untested it should also work as expected. 

@gsfr please review this prototype when you have a chance.

Whitelisted values are temporary placeholders, do not merge until they are replaced with real values.

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -104,6 +104,8 @@ class TargetedPlacer(Placer):
         self.saved = []
 
     def process_file_field(self, field, info):
+        if self.metadata is not None:
+            info.update(self.metadata)
         self.save_file(field, info)
         self.saved.append(info)
 
--- a/api/schemas/input/file.json
+++ b/api/schemas/input/file.json
@@ -1,16 +1,17 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
+  "oneOf": [
+      { "$ref": "file.json#/definitions/MR" },
+      { "$ref": "file.json#/definitions/physio" },
+      { "$ref": "file.json#/definitions/CT" }
+  ],
   "properties": {
     "name":           { "type": "string" },
-    "type":           { "type": "string" },
+    "type":           { "enum": ["dicom", "nifti", null] },
     "mimetype":       { "type": "string" },
-    "instrument":     { "type": "string" },
-    "measurements": {
-      "items": { "type": "string"},
-      "type": "array",
-      "uniqueItems": true
-    },
+    "instrument":     {},
+    "measurements":   {},
     "tags": {
       "items": { "type": "string"},
       "type": "array",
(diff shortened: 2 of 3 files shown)


kofalt:
Looks like there's a merge conflict

api/schemas/input/file.json
```diff
@@ -1,18 +1,19 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
+  "oneOf": [
+      { "$ref": "file.json#/definitions/MR" },
+      { "$ref": "file.json#/definitions/physio" },
+      { "$ref": "file.json#/definitions/CT" }
+  ],
```
kofalt: What does this do at the top level of the schema?

kofalt:
LGTM other than comments

api/schemas/input/file.json
nagem: It allows the instrument and measurements fields to be merged with the rest of the fields in properties. Otherwise you could add the oneOf restraint to a field in properties, in which case instrument and measurements would have to be a named json object

gsfr:
LGTM.

@rentzso Would appreciate your thoughts, too.

Are we looking to roll this out asap? Pre model v2 and pre jinja templating?

api/placer.py
```diff
@@ -93,6 +93,8 @@ def check(self):
         self.saved = []
 
     def process_file_field(self, field, info):
+        for x in ('type', 'instrument', 'measurements', 'tags', 'metadata'):
```
rentzso: shouldn't this check be handled by the schema?

rentzso:
LGTM pending comment and rebase.

api/placer.py
nagem: Ah yes I did this when I thought things other than this whitelist might be in metadata, this can be removed and `info.update(self.metadata)` can replace it.

nagem:
Rebased

kofalt:
A mild goof - I didn't realize that this wasn't to land. Master tuned up & PR re-created :)

Issue #242, opened 2016-04-12
Collections acq endpoint is missing the join-origin


dpuccetti:
When calling /collections/cid/acquisitions?session=sid&join=origin the join-origin field is not included in the response.  This prevents the UI from being able to display the origin icons.

The join-origin field is added to the response from /sessions/sid/acquisitions?join=origin as expected.

kofalt:
Join logic is in `ContainerHandler.handle_origin`, probably just needs to get called for collections as well.

PR #244, opened 2016-04-12
Add join=origin logic to acquisitions on collections

nagem:
Closes https://github.com/scitran/core/issues/242

@dpuccetti make sure this works for you when you get a chance and I'll merge it onto master.

Would we also want the `join=origin` param available for `api/collections/<cid>/sessions`?


nagem:
I added the optional `join=origin` param to `api/collections/<cid>/sessions` as well since we allow it on all containers in `containerhandler.py`.

dpuccetti:
@nagem For the time being the UI really only needs the origin of acquisitions, but I could see it being relevant to sessions as well in the future.  As for whether or not it makes sense to add it now, I'll defer to you and @kofalt.

dpuccetti:
@nagem Looks good to me.  I'm getting what I need from the API.

PR #251, opened 2016-04-18
Return entire group obj to users with Read Only permissions

### LINKED ISSUE #249: Can't fetch group details when user has only read or read/write access.
This is to support dashboard filtering on https://github.com/flywheel-io/frontend/issues/127.

At this point the `/api/groups/foo` request fails with or without the `?root=true` query parameter. So a user could be a member of a group and be unable to use that group's tags for filtering.

Would it be possible to return the tags on the groups when GET'ing `/api/groups`? I'm open to suggestions though.

nagem:
Fixes #249


rentzso:
LGTM

nagem:
@mrDarcyMurphy Let me know if you need anything else from this, otherwise will merge soon.

mrDarcyMurphy:
👍

PR #254, opened 2016-04-19
Lower-case group id in exact group matching

gsfr:
closes flywheel-io/flywheel#169

Sorry for the biggish diff. Took the opportunity to rename `group_name` to `group_id` throughout.


nagem:
LGTM. I also confirmed for myself that the API and the UI properly prevent group `_id`s with non-lowercase characters.

PR #257, opened 2016-04-20
Lookup gravatar and save info on login

### LINKED ISSUE #241: Google overrides Gravatar by default on login
Gunnar and I found this bug last week, but it's been sitting in my queue for awhile and I want to work on other priorities, so documenting it here.

**How to reproduce:**

1. Have a gravatar
1. Configure scitran to use google for authentication
1. Log into the system before anyone has called your `/users/you@example.com/avatar` endpoint
1. Observe that you have a google avatar, rather than your gravatar

**Root cause:**

To load your avatar, the `base.py` loading code will not check if you have a gravatar before saving google as your default avatar. Because google returns a default image, rather than a 404 or similar, this means your gravatar will never be checked or displayed, even by later calls to `/users/you@example.com/avatar`.

This is a natural consequence of switching gravatar checks from on user creation to on-demand.

**How to fix:**

In the base.py [authN layer](https://github.com/scitran/core/blob/6826826c5274fc1d5fbf7cc1846002a1ff5c7f9c/api/base.py#L151-L161), we should also check for a gravatar if none is specified before saving the default. All the relevant code is currently [here](https://github.com/scitran/core/blob/6826826c5274fc1d5fbf7cc1846002a1ff5c7f9c/api/handlers/userhandler.py#L103-L163) in the user handler.

1. Move gravatar code somewhere that can be imported by both files (new module?)
1. Refactor user handler to use new code
1. Add check to base authN to match described, desired behaviour
1. Validate change against:
    1. User with a google avatar
    1. User with a gravatar
    1. User with neither a google avatar nor gravatar

nagem:
Fixes #241.

Tried to find the cleanest "fix" to this. Now on login, we check for a user's gravatar and save it to the avatars list if it is not present. For users logging in for the first time, we will save their gravatar as default if it exists instead of their provider's avatar. Users experiencing issues from #241 should still be able to change their default to their gravatar in user profile settings.

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -156,17 +156,15 @@ class RequestHandler(webapp2.RequestHandler):
             query.pop('sz', None)
             u = u._replace(query=urllib.urlencode(query, True))
             provider_avatar = urlparse.urlunparse(u)
-            default = provider_avatar
-
-            gravatar = util.resolve_gravatar(uid)
-            if gravatar is not None:
-                default = gravatar
-
-            # If the user has no avatar set, mark this default as their chosen avatar.
-            config.db.users.update_one({'_id': uid, 'avatar': {'$exists': False}}, {'$set':{'avatar': default, 'modified': timestamp}})
-
             # Update the user's provider avatar if it has changed.
             config.db.users.update_one({'_id': uid, 'avatars.provider': {'$ne': provider_avatar}}, {'$set':{'avatars.provider': provider_avatar, 'modified': timestamp}})
+
+            # If the user has no avatar set, mark their provider_avatar as their chosen avatar.
+            config.db.users.update_one({'_id': uid, 'avatar': {'$exists': False}}, {'$set':{'avatar': provider_avatar, 'modified': timestamp}})
+
+        # Look to see if user has a Gravatar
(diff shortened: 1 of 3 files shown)


nagem:
The api endpoint `/api/users/self`* might be a better place to put the Gravatar logic as the client would only ever use it for login and profile settings, but checking and updating the user during a GET on a handler class ended up being messier than expected. The PR as is will fix the issue, any future reorganization can happen when a bigger change to the api avatar endpoints is designed. 

*`/api/users/<uid>` is called too frequently for external Gravatar lookups.

gsfr:
@nagem Works for me. I get the Google silhouette initially, but then I can change to my Gravatar.


> *`/api/users/<uid>` is called too frequently for external Gravatar lookups.

@dpuccetti Could you please help me understand this? As far as I can tell, this call isn't even allowed without `root=true`.

dpuccetti:
@gsfr I'm not fully in the loop on this, so I'll defer to what you and @nagem think is best.

nagem:
@gsfr that might have been a misunderstanding on my part during my conversation with @dpuccetti. Going to merge this for now.

gsfr:
👍

PR #262, opened 2016-04-22
Delete user and cleanup permissions

### LINKED ISSUE #252: Add endpoint to permanently delete a user
Add `DELETE /users/<uid>` endpoint for internal maintenance. Require `root=true`. Eventually, this endpoint may be exposed to site-admins in the UI.

Before actually deleting the user, this endpoint needs to do the following:
- remove user from all group roles
- remove user form all project permissions
- remove user from all collection permissions
- delete all collections where user is curator

Relates to #247. Ping @ryansanford, @lmperry, and @nagem. Medium priority and straightforward. Should be prioritized when an equivalent maintenance script would need to be written instead.

nagem:
Fixes #252.

Holding on how to handle [this predicament](https://github.com/scitran/core/issues/252#issuecomment-213565819)

### DIFF
--- a/api/handlers/userhandler.py
+++ b/api/handlers/userhandler.py
@@ -9,6 +9,7 @@ from .. import config
 from .. import validators
 from ..auth import userauth, always_ok, ROLES
 from ..dao import containerstorage
+from ..dao import noop
 
 log = config.log
 
@@ -54,7 +55,11 @@ class UserHandler(base.RequestHandler):
         self._init_storage()
         user = self._get_user(_id)
         permchecker = userauth.default(self, user)
-        result = permchecker(self.storage.exec_op)('DELETE', _id)
+        # Check for authZ before cleaning up user permissions
+        permchecker(noop)('DELETE', _id)
+        self._cleanup_user_permissions(user.get('_id'))
+        log.debug('2')
+        result = self.storage.exec_op('DELETE', _id)
         if result.deleted_count == 1:
             return {'deleted': result.deleted_count}
         else:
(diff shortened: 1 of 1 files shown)


api/handlers/userhandler.py
```diff
@@ -100,6 +105,19 @@ def post(self):
     def _init_storage(self):
         self.storage = containerstorage.ContainerStorage('users', use_object_id=False)
 
+    def _cleanup_user_permissions(self, uid):
```
nagem: This is making db calls, so I'd like it if it weren't in a handler class. It also would only ever be called by userhandler, so it doesn't really work when added to one of the storage classes. 

A proposal would be to maybe have a `user.py` class under `dao` with static methods that pertain to user objects that require db access.

nagem:
#263 has been fixed in #264 and has been merged. This PR is no longer blocked.

api/handlers/userhandler.py
```diff
@@ -100,6 +105,19 @@ def post(self):
...
+        try:
+            config.db.collections.delete_many({'curator': uid})
```
kofalt: This makes me nervous. I don't like the idea of shares evaporating, especially if others are using them. Had a conversation with @elisapoquette on how we want to proceed; I'll let her comment from here

kofalt:
LGTM other than comment

api/handlers/userhandler.py
elisapoquette: If a user curates a collection, and adds at least 1 other admin, I would expect that collection to still exist. But given that we're not exposing this in the UI (and only using this internally) maybe that's ok.

kofalt:
Based on Elisa's comment I'll assign this to @gsfr for a go/no-go

gsfr:
Let's only delete the collection if no other users have access to it. If we do delete, we also need to update the collection pointers on the member acquisitions.

kofalt:
In the case where others users have access, should one of them be transferred to the curator?
Do we leave the curator field alone? Remove it entirely?

gsfr:
Just leave the curator alone.

nagem:
Changed to only delete collections where user is the only user in the permissions list. 

Retested all functionality again, this is good to go - will merge tomorrow unless someone has any additional comments.

PR #264, opened 2016-04-25
Add curator to collection on POST, cleanup existing collections without a curator field

### LINKED ISSUE #263: Store collection creator as 'curator' during POSTs
We are currently adding the user as admin in a collection's permissions list instead of saving their `uid` as `curator`. 

See here: https://github.com/scitran/core/blob/59e35687bd433872889e42dae8c23938450ff2a8/api/handlers/collectionshandler.py#L34-L38

If there is only one admin in the permissions list we can probably assume they are the curator and fix this in a db update script.

nagem:
Fixes #263.

Mongo pipeline for finding collections with one user that has admin permissions is a bit complex, on Mongo 3.2 we could use `$filter` like so:

```
pipeline = [
        {'$match': {'curator': {'$exists': False}, 'permissions.access': 'admin'}},
        {'$project': {
             'cid': '$_id',
             'perms': {
                '$filter': {
                   'input': '$permissions',
                   'as': 'perms',
                   'cond': {'permissions.access': 'admin'}
                }
             }
        }},
        {'$match': {'perms': {'$size': 1}}}
    ]
```

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -83,25 +83,15 @@ def upgrade_to_3():
     """
     scitran/core PR #263
 
-    Set User as curator of collection if:
-      - collection has no curator
-      - User is the only user with admin perms for the collection
+    Set first user with admin permissions found as curator if one does not exist
     """
 
-    pipeline = [
-        {'$match': {'curator': {'$exists': False}, 'permissions.access': 'admin'}},
-        {'$unwind': '$permissions'},
-        {'$project': {'cid': '$_id', 'access': '$permissions.access', 'site': '$permissions.site',  'user': '$permissions._id'}},
-        {'$group' : { '_id' : {'cid': '$cid', 'access': '$access'}, 'users': {'$push': '$user' }}},
-        {'$match': {'_id.access': 'admin', 'users': {'$size': 1}}}
-    ]
-
-    collections = config.db.command('aggregate', 'collections', pipeline=pipeline)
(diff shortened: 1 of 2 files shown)


gsfr:
Handler change LGTM. Update script, too. We need to make sure all collections have a curator. Even at the risk of getting it wrong when there is more than one admin, I suppose.

nagem:
If we want to take any admin we find, the update becomes much simpler:

`config.db.collections.update_many({'curator': {'$exists': False}, 'permissions.access': 'admin'}, {'$set': {'curator': 'permissions.$._id'}})`

or something like that.

nagem:
Apparently mongo doesn't let you refer to existing document values in the `$set` operator, doing it in python instead.

bin/database.py
```diff
@@ -79,6 +79,19 @@ def update_file_origins(cont_list, cont_name):
...
+def upgrade_to_3():
+    """
+    scitran/core PR #263
+
+    Set first user with admin permissions found as curator if one does not exist
+    """
+
+    collections = config.db.collections.find({'curator': {'$exists': False}, 'permissions.access': 'admin'})
+    for coll in collections:
+        admin = coll['permissions'][0]['_id']
```
nagem: This would return all users, have to search for admin, fixing

gsfr:
Well, I guess, the list is ordered and the first one should be the right one.

PR #266, opened 2016-04-26
Add ids to session subjects

### LINKED ISSUE #253: Add unique subject id
When creating a new session in the DB, add an `_id` field of type `ObjectId` to the `subject` subdocument.

If another session with the same `subject.code` already exists in the new session's project, copy the `subject._id` from the existing session, else generate a new one.

Also, allow `subject._id` to be set explicitly on new and existing sessions via `/sessions`.

nagem:
Closes #253, check out the issue for information about the requirements.

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -1,0 +1,28 @@
+import bson.objectid
+from .. import config
+
+log = config.log
+
+
+def add_id_to_subject(subject, pid):
+    """
+    Add a mongo id field to given subject object (dict)
+
+    Use the same _id as other subjects in the session's project with the same code
+    If no _id is found, generate a new _id
+    """
+    result = None
+    if subject is None:
+        subject = {}
+    if subject.get('_id') is not None:
+        return subject
+    if subject.get('code') is not None:
+        query = {'subject.code': subject['code'],
+                 'project': pid,
+                 'subject._id': {'$exists': True}}
+        result = config.db.sessions.find_one(query)
+    if result is not None:
+        subject['_id'] = str(result['subject']['_id'])
+    else:
+        subject['_id'] = str(bson.ObjectId())
+    return subject
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -15,14 +15,20 @@ def add_id_to_subject(subject, pid):
     if subject is None:
         subject = {}
     if subject.get('_id') is not None:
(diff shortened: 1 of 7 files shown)


api/dao/containerutil.py
```diff
@@ -0,0 +1,28 @@
+import bson.objectid
```
nagem: This one broke the camel's back of needing a better place to put functions that act on specific containers and interact with the db. Suggesting this module as a midstep between ad hoc as needed methods outside of the `doa` package and having container-specific modules.

api/dao/containerutil.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Use the same _id as other subjects in the session's project with the same code
+    If no _id is found, generate a new _id
+    """
+    result = None
+    if subject is None:
+        subject = {}
+    if subject.get('_id') is not None:
+        return subject
+    if subject.get('code') is not None:
```
kofalt: Suggest adding whitespace before this line to separate out the conditionals for readability

api/dao/containerutil.py
```diff
@@ -0,0 +1,28 @@
...
+        query = {'subject.code': subject['code'],
+                 'project': pid,
```
kofalt: `subject` is checked for `None`, while `pid` is not. Intended?

kofalt:
LGTMs after comment

api/dao/containerutil.py
```diff
@@ -0,0 +1,28 @@
...
+                 'subject._id': {'$exists': True}}
+        result = config.db.sessions.find_one(query)
+    if result is not None:
+        subject['_id'] = str(result['subject']['_id'])
+    else:
+        subject['_id'] = str(bson.ObjectId())
+    return subject
```
gsfr: Missing newline at EOF.

Also surprised to see that you are storing the `ObjectId` as a string in the DB. Why?

nagem: Good catch, I was missing `"ensure_newline_at_eof_on_save": true` in my Sublime prefs.

I was playing with the idea of keeping the `_id` field as a string to cut back on having to ensure it is a bson ObjectId on every update/create of a session object throughout the codebase. After thinking about it, it's now back to being a bson ObjectId.

nagem:
Rebased and retested, will merge tomorrow unless there are any new comments.

PR #271, opened 2016-05-02
Add ability to replace metadata fields in API

### LINKED ISSUE #267: Allow a way to replace rather than patch objects via API
Specifically needed to address the issue of deleting fields inside unstructured json fields like `info`, the API needs to allow replacing the data at a resource endpoint. Currently, PUT only allows patch functionality.

nagem:
Closes #267 as it solves the immediate problem. 

This acts as a halfway point to having dedicated endpoints for modifying `metadata` fields. As there is no way currently to delete fields inside the arbitrary `metadata` map, sending the flag `replace_metadata` with a put request will change all metadata field updates in the container to be a replace rather than a patch.

More thought should be put into how to properly handle unstructured map updates and deletes.

### DIFF
--- a/api/auth/containerauth.py
+++ b/api/auth/containerauth.py
@@ -48,8 +48,10 @@ def default_container(handler, container=None, target_parent_container=None):
 
             if has_access and projection:
                 return exec_op(method, _id=_id, payload=payload, projection=projection)
+            if has_access and recursive:
+                return exec_op(method, _id=_id, payload=payload, recursive=recursive, r_payload=r_payload, replace_metadata=replace_metadata)
             elif has_access:
-                return exec_op(method, _id=_id, payload=payload)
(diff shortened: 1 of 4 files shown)


gsfr:
LGTM as interim solution.

@rentzso Could you please take a quick look?

gsfr:
@nagem Coveralls sometimes gets stuck. Just ignore it.

rentzso:
Just a question. Don't we need to allow this also for files metadata?
LGTM though

nagem:
@rentzso there isn't an immediate need for file metadata to be modified, hopefully we can get the more robust solution finalized before that need arises. Good catch, though.

gsfr:
`metadata` exist throughout the hierarchy. Any long-term solution should address it everywhere. But I think we are good to go here right now.

PR #276, opened 2016-05-04
Reporting Engine

### LINKED ISSUE #258: Generate json report for Site and Project stats
Project Report (Project Admin, Group Admin)
- [x] Configure by Project (Select one or multiple, only the ones you have permission to). 
- [x] Configure by Date Range 
- [x] Project Name
- [x] Group Name
- [x] Project Admin
- [x] Number of Sessions 
- [x] Unique Subjects
- [x] Male Subjects
- [x] Female Subjects
- [x] Subjects under 18
- [x] Subjects over 18

Site Report (Site Manager only)
- [x] number of groups  
- [x] number of projects per group  
- [x] number of sessions per group

nagem:
Closes #258. 

A first pass at creating site and project reports.


api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+    if output.get('ok', 0) is not 1.0:
+        result = output.get('result')
+        if result is not None:
+            return result[0]
+
+    raise APIReportException
+
+class Report(object):
+
+    def user_can_generate(self, uid):
```
nagem: This looks more like an interface than an abstract class right now but future mentioned functionality like generating a csv from the json data would go here as a common method. Or it could be handled at the handler level. Curious see hear any thoughts on designing this module.

api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    if output.get('ok', 0) is not 1.0:
+        result = output.get('result')
+        if result is not None:
+            return result[0]
+
+    raise APIReportException
+
+class Report(object):
```
nagem: Looks like I just deleted the comment that was here. It went something like this:

Currently this looks more like an interface contract than an abstract base class. In the future this class can hold shared functionality like the future feature of creating a csv from build's json result. Curious to hear any opinions on how the reporting module was designed.

api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    Report of statistics about the site, generated by Site Managers
+
+    Report includes:
+      - number of groups
+      - number of projects per group
+      - number of sessions per group
+    """
+
+    def user_can_generate(self, uid):
```
nagem: I allow site admins to request a site report without making an explicit superuser request, can be talked out of it.

api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+
+            report = ProjectReport(map(bson.ObjectId, project_list),
+                                   start_date=start_date,
+                                   end_date=end_date)
+
+        else:
+            # They should never even get this far because of filtering in api.py
+            self.abort(400, 'The report type {} is not supported'.format(report_type))
+
+        if self.superuser_request or report.user_can_generate(self.uid):
```
nagem: superusers requesting as such do not need to have individual permissions on each project.

api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+            end_date = self.get_param('end_date')
+
+            if len(project_list) < 1:
+                self.abort(400, 'List of projects requried for Project Report')
+            if start_date is not None:
+                start_date = dateutil.parser.parse(self.get_param('start_date'))
+            if end_date is not None:
+                end_date = dateutil.parser.parse(self.get_param('end_date'))
+
+            report = ProjectReport(map(bson.ObjectId, project_list),
```
kofalt: Nice map!

api/handlers/reporthandler.py
kofalt: I like this system. My only comment so far is that validation & defaults could be the Report subclass' job: `len(project_list) < 1` works in the http handler, but it feels like it belongs in `ProjectReport.__init__`.

api/handlers/reporthandler.py
```diff
@@ -0,0 +1,257 @@
...
+import dateutil
+import copy
+
+from .. import base
+from .. import config
+from .. import util
+
+log = config.log
+
+EIGHTEEN_YEARS_IN_SEC = 568036800
```
kofalt: Generally, I prefer these kind of "we made a const of a duration because we have to, sorry" declarations to be a call to the language's time functions, or at least the actual arithmetic. 

`18 * 365.25 * 24 * 60 * 60` is comprehensible in a way that `568036800` isn't :smile:

kofalt:
LGTM other than comments!

api/handlers/reporthandler.py
nagem: Good point, will do.

nagem:
Merging in the next hour.

PR #285, opened 2016-05-06
Fix zipfile bugs, simplify logic

### LINKED ISSUE #283: Fix zipfile import error, cleanup FileListHandler responses
Using any of the optional params on `FileListHandler.get` result in a 500 as the `zipfile` lib is not being imported.

While looking into this error I noticed our responses for `GET` on a file with the params `info=true` or `comment=true` are not JSON. That should be cleaned up.

nagem:
Fixes #283.

 - Fixed import issue
 - Combined `input` and `comment` requests into one JSON response
 - Simplified `FileListHandler.get` to make it a bit more readable

### DIFF
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -5,6 +5,7 @@ import datetime
 import dateutil
 import json
 import uuid
+import zipfile
 
 from .. import base
 from .. import config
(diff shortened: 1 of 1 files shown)


gsfr:
LGTM and seems to work as expected.

Issue #202, opened 2016-03-16
Stack traces logged during benign non-200 returns


kofalt:
Stack traces catch the eye when developing or reviewing logs, and right now we sometimes dump a stack in cases that are expected and benign:

```
2016-03-16 15:53:27      scitran.api                  base.py   189:ERRO Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "./api/jobs.py", line 292, in next
    self.abort(400, 'No jobs to process')
  File "./api/base.py", line 254, in abort
    webapp2.abort(code, detail=detail, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1830, in abort
    raise cls(*args, **kwargs)
HTTPBadRequest: No jobs to process
```

This is partially blocked by #66 for the reasons mentioned in that thread, but we might be able to solve this sooner by changing our usage of `self.abort`. Filed as part of an ops conversation with @ryansanford. _Update_: @lmperry just ran into this in prod as well.

kofalt:
Just a side note that the above stack trace is printed on (to my knowledge) every production deployment into the api log, every 5 seconds. I'd really love to get rid of this...

lmperry:
+1 on that. It drives me just a little crazy.

PR #287, opened 2016-05-06
Add some sanity to jobs logic

### LINKED ISSUE #273: Remove jobs.add_raw
Awhile after #270 lands & confirmed working, we should remove `jobs.add_raw` for obvious reasons :)

kofalt:
Jobs.py was a 450-line file with four distinct concepts tightly interwoven, which became a total nightmare that could not be tolerated one more instant. This change attempts to tease them apart, and organize each logical portion into a file or class.

No API changes. Closes #273. Progress towards #66.


api/api.py
```diff
@@ -258,11 +261,20 @@ def _format(route):
...
 
 
+def custom_json_serializer(obj):
+    if isinstance(obj, bson.objectid.ObjectId):
+        return str(obj)
+    elif isinstance(obj, datetime.datetime):
+        return pytz.timezone('UTC').localize(obj).isoformat()
+    elif isinstance(obj, Job):
+        return obj.map()
+    raise TypeError(repr(obj) + " is not JSON serializable")
```
kofalt: This had to be bumped over from `util` due to the import graph.

api/jobs/gears.py
```diff
@@ -0,0 +1,44 @@
...
+        {'gear_list': { '$elemMatch': {
+            'name': name
+        }}
+    })
+
+    if gear_doc is None:
+        raise Exception('Unknown gear ' + name)
+
+    # Mongo returns the full document: { '_id' : 'gears', 'gear_list' : [ { .. } ] }, so strip that out
+    return gear_doc['gear_list'][0]
```
kofalt: Could totally be under a Gear class; these functions are sufficiently nascent that we'd inevitably change it all later.

api/api.py
gsfr: I don't understand. You are still importing util in this file. What's the problem?

kofalt:
Ref lemurheavy/coveralls-public#776 for CI failure

api/api.py
kofalt: At the time, `Job` imported `util`, which could not in turn import `Job`.

In the most recent commit, that's no longer required, but it stands that any file which uses utility methods could not export a class for json serialization.

api/jobs/handlers.py
```diff
@@ -0,0 +1,105 @@
...
+        return Queue.get_statistics()
+
+    def next(self):
+        if not self.superuser_request:
+            self.abort(403, 'Request requires superuser')
+
+        job = Queue.start_job()
+
+        if job is None:
+            self.abort(400, 'No jobs to process')
```
nagem: @kofalt Looks like you're still throwing a 400 when there are no jobs to process, so the fix for #202 will still work as intended. 

Going to open a ticket specifically addressing the handling of exceptions in non-handler classes from #66 so we can finalize the way we would like to pass server and client error exceptions from non-handler classes.

api/jobs/gears.py
```diff
@@ -0,0 +1,32 @@
...
+log = config.log
+
+
+def get_gears():
+    """
+    Fetch the install-global gears from the database
+    """
+
+    gear_doc  = config.db.static.find_one({'_id': 'gears'})
+    return gear_doc['gear_list']
```
nagem: Could have error-checking around a missing/malformed gears doc but that would be overkill and the 500 that will get logged for the KeyError will make the actual error clear enough.

api/jobs/handlers.py
```diff
@@ -0,0 +1,105 @@
...
+    """Provide /jobs API routes."""
+
+    def get(self):
+        """
+        List all jobs.
+        """
+        if not self.superuser_request:
+            self.abort(403, 'Request requires superuser')
+
+        return list(config.db.jobs.find())
```
nagem: What typically consumes this endpoint?

I tested on my dev environment with a large dataset (8900 jobs) and the request was ~2000 ms, for reference. Depending on what consumes this and how large we expect that container to grow, this might be a good candidate for pagination in the future.

nagem:
LGTM, had a few questions/comments but no real changes need to be made.

api/jobs/handlers.py
kofalt: Agreed - *nobody* should consume this endpoint :)

I would be happy to remove it; this is a pre-existing endpoint originally added to match the convention of other endpoints that dump entire tables. Since this PR aims to have no exported changes, I'd like to leave it for now.

PR #288, opened 2016-05-06
Allow certain file properties to be null

gsfr:
Minor change needed for uploads that cannot be typed by the folder uploader. I hopes this does not violate some implicit assumption that all files are typed.

Ping @rentzso @nagem @kofalt @lmperry

### DIFF
--- a/api/schemas/input/file.json
+++ b/api/schemas/input/file.json
@@ -3,9 +3,9 @@
   "type": "object",
   "properties": {
     "name":           { "type": "string" },
-    "type":           { "type": "string" },
-    "mimetype":       { "type": "string" },
-    "instrument":     { "type": "string" },
+    "type":           { "type": ["string", "null"] },
+    "mimetype":       { "type": ["string", "null"] },
+    "instrument":     { "type": ["string", "null"] },
     "measurements": {
       "items": { "type": "string"},
       "type": "array",


kofalt:
The rules module at minimum does: `file_['type'] == match_param`

I would be happier with emptystring than null. Frequently I've found the amount of maybe-key-is-there, maybe-key-is-even-the-right-type! busywork to be a preponderance of effort to manage relative to the actual complexity of our structures. 

Emptystring is far more convenient, and a reasonable way to say "we don't know the type of this file", relative to "we deny that the concept of a type exists".

gsfr:
@kofalt I don't share that view, as it is contradictory to our current data model which makes very few guarantees. Also, writing `file_.get('type', '')` doesn't seem to be too much to ask for.

I'll leave it up to the api developers to evaluate the global consequences of tightening our data model to make more such guarantees.

We should fix this quickly because we currently have a discrepancy between our lax schemas (which don't require the presence of the `type` field) and non-conservative coding as in the example above.

kofalt:
Keeping consistency is a good point. I would be fine with going ahead on this change and revisiting the issue later - I don't agree that `.get(x, y)` is a solution, but that can be discussed later.

To land this branch, it'll need fixes to wherever we access these properties, presumably six searches: `'type'`, `.type`, etc. We also need to know from @dpuccetti what UI assumptions are invalidated by these keys changing type.

nagem:
@gsfr we default those values to `None` (or a best guess from the filename for mimetype) if they are not provided in [upload.py](https://github.com/scitran/core/blob/master/api/upload.py#L110-L123). I think the best solution would be to not send those keys from the folder reaper if they are unknown. One usecase for allowing `null` is to unset those attributes, not sure to that need or if it should be allowed.

nagem:
We also do some assumptions with the `type` field in packfile uploads but it is a required field in `packfile.json`.

gsfr:
Ok. I have accommodated for this in the reaper instead (https://github.com/scitran/reaper/commit/29db44cec54f26a4613b62ce2a23c59a387ed0a7). I now suggest closing this ticket and picking up the `.get(x, y)` discussion another day.

PR #289, opened 2016-05-09
Only log stacktrace for 500 errors

### LINKED ISSUE #202: Stack traces logged during benign non-200 returns

nagem:
Fixes #202.

@rentzso @gsfr @lmperry: @kofalt, @ryansanford and I discussed only logging stacktraces for 500 errors, let me know if there are any unconsidered edge cases where a client error (4xx) should have a stacktrace logged.

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -239,12 +239,12 @@ class RequestHandler(webapp2.RequestHandler):
         return self.request.GET.get(param, default)
 
     def handle_exception(self, exception, debug):
-        # Log the error.
-        tb = traceback.format_exc()
-        log.error(tb)
+        """
+        Send JSON response for exception
 
-        # If the exception is a HTTPException, use its error code.
-        # Otherwise use a generic 500 error code.
+        For HTTP and other known exceptions, use its error code
+        For all others use a generic 500 error code and log the stack trace
+        """
         if isinstance(exception, webapp2.HTTPException):
             code = exception.code
         elif isinstance(exception, validators.InputValidationException):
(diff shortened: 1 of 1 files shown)


api/base.py
```diff
@@ -255,6 +255,11 @@ def handle_exception(self, exception, debug):
             code = 400
         else:
             code = 500
+
+        if code == 500:
```
nagem: This could also be a check for 5xx errors - we return 503s for DB or Elastic unavailability. Not sure we would need a stacktrace for that.

kofalt:
LGTM

gsfr:
LGTM, but others have move visibility.

rentzso:
LGTM

PR #290, opened 2016-05-10
Create restrictions on file fields (2)

### LINKED ISSUE #210: File Input Schema - 'null' type for files

### LINKED ISSUE #231: Create restrictions on file fields

kofalt:
Replacement PR for #239. Not yet ready to land.

Original description from @nagem:

---

Closes #231.
Closes #210.

Restrictions to type, instrument (becoming modality when `data-model-v2` is merged), and measurements working as expected for targeted upload. `EnginePlacer` also sets these fields and uses the `file.json` schema so although untested it should also work as expected. 

@gsfr please review this prototype when you have a chance.

Whitelisted values are temporary placeholders, do not merge until they are replaced with real values.

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -104,6 +104,8 @@ class TargetedPlacer(Placer):
         self.saved = []
 
     def process_file_field(self, field, info):
+        if self.metadata is not None:
+            info.update(self.metadata)
         self.save_file(field, info)
         self.saved.append(info)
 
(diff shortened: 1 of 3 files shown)


nagem:
I don't see any movement on this PR in the near future, will close but preserve the branch.

PR #293, opened 2016-05-11
Add /container/id/jobs endpoint

kofalt:
This endpoint will allow a user to see jobs that consumed a particular container.
Unlike all other job endpoints currently, it is accessible to non-superusers (read access is required).

Also supported are filtering by both tags and states.

For example: `/acquisitions/3/jobs?states=pending&states=failed&tags=a&tags=b` -->

```javascript
[
    {
        "_id": "57336257ea360500272c2f11", 
        "algorithm_id": "dicom_mr_classifier", 
        "state": "pending", 
        "tags": [
            "a"
        ]
        //...
    }, 
    {
        "_id": "57336257ea360500272c2f12", 
        "algorithm_id": "dcm_convert", 
        "state": "failed", 
        "tags": [
            "b"
        ]
       // ...
    }
]
```

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -156,16 +156,64 @@ class ContainerHandler(base.RequestHandler):
             result['permissions'] = [user_perm] if user_perm else []
 
     def get_jobs(self, cont_name, cid):
-        try:
-            self.config = self.container_handler_configurations[cont_name]
-            self.storage = self.config['storage']
-            c = self._get_container(cid)
-            permchecker = self._get_permchecker(c)
-            result = permchecker(noop)('GET', cid)
-        except APIStorageException as e:
-            self.abort(400, e.message)
+        """
+        .. http:get:: /api/(cont_name)/(cid)/jobs
 
-        cr     = containerutil.ContainerReference(cont_name[:-1], cid)
+            Return any jobs that mention this container as an input.
+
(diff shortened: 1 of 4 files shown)


ryansanford:
Please add REST Interface docs for get_jobs(). Examples of REST interface docs are included in the below merge.

3d18253e81a87e377e71d19ca3e75eb38a206786

api/handlers/containerhandler.py
```diff
@@ -154,6 +155,21 @@ def _filter_permissions(self, result, uid, site):
...
 
+    def get_jobs(self, cont_name, cid):
+        try:
+            self.config = self.container_handler_configurations[cont_name]
+            self.storage = self.config['storage']
+            c = self._get_container(cid)
+            permchecker = self._get_permchecker(c)
+            result = permchecker(noop)('GET', cid)
+        except APIStorageException as e:
+            self.abort(400, e.message)
```
nagem: I would remove the try block here. Permissions checking shouldn't be wrapped in a try block as it will self.abort with it's own error code and message, as will `_get_container` (but will error with an APIStorageException as this will here). A lookup in our configuration map is safe.

nagem:
LGTM other than unnecessary try block I mentioned.

kofalt:
Docs added, block removed; merging when CI green.

PR #295, opened 2016-05-12
Move all singletons to new mongo collection

### LINKED ISSUE #291: Move all singletons to a new Mongo collection
Move the following singleton documents to a new Mongo collection called `singletons`.
- config
- version
- gears

Ultimately, gears should be migrated to a separate collection.

nagem:
Closes #291

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -52,26 +52,27 @@ DEFAULT_CONFIG = {
     },
 }
 
-__config = copy.deepcopy(DEFAULT_CONFIG)
+def apply_env_variables(config):
+    # Overwrite default config values with SCITRAN env variables if available
+    for outer_key, scoped_config in config.iteritems():
+        if outer_key in ['core', 'site', 'auth', 'persistent']
+            for inner_key in scoped_config:
+                key = 'SCITRAN_' + outer_key.upper() + '_' + inner_key.upper()
+                if key in os.environ:
+                    value = os.environ[key]
+                    if value.lower() == 'true':
+                        value = True
+                    elif value.lower() == 'false':
+                        value = False
+                    elif value.lower() == 'none':
+                        value = None
+                    config[outer_key][inner_key] = value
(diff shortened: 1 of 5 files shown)


api/config.py
```diff
@@ -52,27 +52,28 @@
     },
 }
 
-__config = copy.deepcopy(DEFAULT_CONFIG)
+def apply_env_variables(config):
+    # Overwrite default config values with SCITRAN env variables if available
+    for outer_key, scoped_config in config.iteritems():
+        if outer_key in ['core', 'site', 'auth', 'persistent']:
```
gsfr: What's the purpose of this check? I wouldn't anticipate a code change when adding a new outer config key.

gsfr:
LGTM, pending one question. Also, untested. And I'm still sick and probably shouldn't be trusted with code reviews today.

api/config.py
nagem: It's because this can now be called on a db config object, which has keys like `created` and `modified`.

gsfr: Hmm, I don't like repeating the keys. Does this really have to be a callable? You're only calling it once, and that's right below the definition.

nagem: I also call it [here](https://github.com/scitran/core/pull/295/files/a935657d5a852fc569c6394c2ae0c7de3e7e71a7#diff-283256b6a6d2885d51431bd002292cdaR199) to preserve any environment variables on startup after applying configs from the db. I could also check if `scoped_config` has an `__iter__` property so we don't try to iterate over the dates.

nagem: Before we ignored anything in db config object on startup and overwrote it with env vars and default values. This change is setting the order of precedence to be env > db > default on startup.

gsfr: I didn't realize that. Good catch. You could `try` the iteration or use `DEFAULT_CONFIG.keys()`.

Issue #299, opened 2016-05-16
Add tags field to /jobs/next


kofalt:
Filter job queue retrievals by some set of tags: `/jobs/next?tags=a,b`.

The endpoint will behave identically when the `tags` field is present, except that results (if any) must have at least one of the tags specified.

Requested by @lmperry.

kofalt:
Resolved in #301 :+1:

Issue #296, opened 2016-05-12
Allow users to call /jobs/add


kofalt:
From `api/jobs/handlers.py`:

```python
	def add(self):
		"""
		Add a job to the queue.
		"""

		# TODO: Check each input container for R, check dest container for RW
		if not self.superuser_request:
			self.abort(403, 'Request requires superuser')
```

@rentzso, would you mind lending me a hand?

Based on our conversation yesterday I was able to add simple permission checking for the new endpoint `/container/id/jobs`, but I am not sure how to implement the `TODO` mentioned in the code above.

rentzso:
Sure. Assigning this to myself.

kofalt:
Cool.

Further on after that snippet, you'll see that the `inputs` variable is a `string -> FileReference` map, and the `destination` variable is a `ContainerReference`. My assumption is the auth check will have to move there, right before the `job = Job(` line.

Oh, and both of those types can be found in `api/dao/containerutil.py`.

elisapoquette:
Dependency for Adhoc gear prototype - short term, comment it out so @dpuccetti can start work

kofalt:
Workaround placed in branch for scitran/core#301.

kofalt:
This was resolved in scitran/core#301 :+1:

PR #301, opened 2016-05-16
Gear routes and ad-hoc UI enablers

kofalt:
Add `/gears` and `/gears/x` routes to the API.

Must not be merged until #296 is solved here.
Note to @rentzso - feel free to put the change for that on this branch.

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -6,27 +6,44 @@ from .. import config
 
 log = config.log
 
+# For now, gears are in a singleton, prefixed by a key
+SINGLETON_KEY = 'gear_list'
 
-def get_gears():
+def get_gears(fields=None):
     """
     Fetch the install-global gears from the database
     """
 
-    gear_doc  = config.db.singletons.find_one({'_id': 'gears'})
-    return gear_doc['gear_list']
+    projection = { }
+
+    if fields is None:
+        fields = [ ]
+        projection = { SINGLETON_KEY: 1 }
+    else:
+        fields.append('name')
+
+    query = {'_id': 'gears'}
+
+    for f in fields:
+        projection[SINGLETON_KEY + '.' + f] = 1
+
+    gear_doc = config.db.singletons.find_one(query, projection)
+
+    # print gear_doc
+    return gear_doc[SINGLETON_KEY]
 
 def get_gear_by_name(name):
 
     # Find a gear from the list by name
     gear_doc = config.db.singletons.find_one(
         {'_id': 'gears'},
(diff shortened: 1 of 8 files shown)


api/jobs/gears.py
```diff
@@ -6,27 +6,44 @@
...
 
-    gear_doc  = config.db.singletons.find_one({'_id': 'gears'})
-    return gear_doc['gear_list']
+    projection = { }
+
+    if fields is None:
+        fields = [ ]
+        projection = { SINGLETON_KEY: 1 }
+    else:
+        fields.append('name')
```
nagem: Does mongo error if you specify a projected field twice?

api/jobs/handlers.py
```diff
@@ -7,12 +7,121 @@
...
+                        },
+                    },
+                }
+
+        """
+
+        if self.public_request:
+            self.abort(403, 'Request requires login')
+
+        return get_gear_by_name(_id)
```
nagem: Probably want to throw a 404 if the return is None.

nagem: I see you throw an exception in `get_gear_by_name` if it is not found. Maybe add a try around here so you can throw the proper status code.

nagem:
LGTM pending comments.

api/jobs/gears.py
kofalt: I checked - it does not! Otherwise I could do the `list(set(x))` trick.

kofalt:
Renamed a few redundant variables, and added tag filtering to `/jobs/next`. Closes #299.

kofalt:
Rebased to resolve conflicts with #304

rentzso:
@kofalt just pushed a new commit that should solve #296 
I haven't tested it though. Could you give it a try?

If it doesn't work, please provide some test scripts that I can use to reproduce.

kofalt:
Thanks! I'll have a go at it.

bin/database.py
```diff
@@ -257,6 +257,50 @@ def upgrade_to_9():
...
+    FR.container_id   --> id
+    FR.filename       --> name
+
+    CR.container_type --> type
+    CR.container_id   --> id
+    """
+
+    def switch_keys(doc, x, y):
+        doc[y] = doc[x]
+        doc.pop(x, None)
```
nagem: you could also one line it like:

`doc[y] = doc.pop(x, None)`

kofalt: Heh, you're not wrong...

nagem:
New changes LGTM.

PR #302, opened 2016-05-16
Bootstrap

gsfr:
Various bootstrap.py and run.sh improvements.

@ryansanford Please ensure I'm not violating some implicit assumption about bootstrap.py.

### DIFF
--- a/bin/bootstrap.py
+++ b/bin/bootstrap.py
@@ -29,17 +29,26 @@ def users(filepath, api_url, http_headers, insecure):
         rs.verify = not insecure
         rs.headers = http_headers
         for u in input_data.get('users', []):
-            log.info('    ' + u['_id'])
-            rs.post(api_url + '/users', json=u)
+            r = rs.post(api_url + '/users', json=u)
+            if r.ok:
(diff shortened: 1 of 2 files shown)


ryansanford:
@gsfr I like the response checking in bootstrap.py. Adding a check on getting the site_id would be welcome. Suggest exiting with non-zero return in that case.

run.sh:
I think you want to init `$BOOTSTRAP_USERS` and `$BOOTSTRAP_TESTDATA` to 0, then set to 1 when flagged?

I'm going to open a separate issue for response codes used when inserting duplicate users and roles. Right now duplicate user insertion results in a 500 response (mongo duplicate key), and duplicate role insertion is resulting in 404.

gsfr:
@ryansanford Are you asking to hard abort when a `site_id` field exists in the JSON file?

Capitalized options are intended to disable behavior. run.sh works as expected.

ryansanford:
@gsfr I wasn't aware of that convention with using capitalized parameters. Suggest expanding the `USAGE` text to explain those flags. Otherwise LGTM

gsfr:
`USAGE` string updated.

PR #304, opened 2016-05-17
Ensure timestamps are not blank strings

### LINKED ISSUE #292: Ensure timestamps on sessions and acquisitions are an ISO formatted string (or null)
After a discussion with @rentzso about improperly formatted timestamps causing ElasticSearch headaches, a few places need to be changed:

`"timestamp":    {"type": ["string", "null"]}` needs to become `"timestamp":    {"type": "string", "format": "date-time"}` in:
 - input/session.json
 - input/acquisition.json
 - input/uidupload.json
 - input/labelupload.json
 - input/enginemetadata.json
 - mongo/session.json
 - mongo/acquisition.json

Checks for timestamps in session/acquisition creation in hierarchy.py need to look for `None`.

@rentzso please fill in any missing details for this ticket.

nagem:
Fixes #292.

@rentzso here is the fix you requested - the date-time format does not allow blank strings so in-code checks are unnecessary. This is also not a breaking change to any of our frontend logic.

### DIFF
--- a/api/validators.py
+++ b/api/validators.py
@@ -32,7 +32,7 @@ def validate_data(data, schema_json, schema_type, verb, optional=False):
     validator(data, verb)
 
 def _validate_json(json_data, schema, resolver):
-    jsonschema.validate(json_data, schema, resolver=resolver)
+    jsonschema.validate(json_data, schema, resolver=resolver, format_checker=jsonschema.FormatChecker())
 
 class RefResolver(jsonschema.RefResolver):
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ pytz==2015.7
 requests==2.9.1
 requests-toolbelt==0.6.0
 rfc3987==1.3.4
+strict-rfc3339==0.7
 webapp2==2.5.2
 WebOb==1.5.1
 elasticsearch==1.9.0
--- a/api/schemas/input/acquisition.json
+++ b/api/schemas/input/acquisition.json
@@ -13,7 +13,7 @@
         "uid":          {"type": "string"},
         "instrument":   {"type": "string"},
         "measurement":  {"type": "string"},
-        "timestamp":    {"type": "string"},
+        "timestamp":    {"type": "string", "format": "date-time"},
         "timezone":     {"type": "string"}
     },
     "required": ["label"],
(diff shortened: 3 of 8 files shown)


nagem:
This and #301 have conflicting database upgrades. Whichever merges second will need a rebase.

api/validators.py
```diff
@@ -32,7 +32,7 @@ def validate_data(data, schema_json, schema_type, verb, optional=False):
     validator(data, verb)
 
 def _validate_json(json_data, schema, resolver):
-    jsonschema.validate(json_data, schema, resolver=resolver)
+    jsonschema.validate(json_data, schema, resolver=resolver, format_checker=jsonschema.FormatChecker())
```
kofalt: Wow. For those who, like me, were wondering why this is necessary, it turns out `format` is optional:

> JSON Schema impls are not required to implement this part of the specification, and many of them do not.

No big deal, but something to keep in mind if we a) change libraries, or b) want to validate in another language.

requirements.txt
```diff
@@ -9,6 +9,7 @@ pytz==2015.7
 requests==2.9.1
 requests-toolbelt==0.6.0
 rfc3987==1.3.4
+strict-rfc3339==0.7
```
kofalt: I don't see this explicitly imported anywhere.
Is this one of those sneaky-conditional python imports for the format checking?

kofalt:
LGTM pending a quick question.

Presumably this will be easy to land soon; go ahead & claim v9, I will take v10 :laughing:

rentzso:
@nagem 
as from @ryansanford comment https://github.com/flywheel-io/flywheel/pull/195#issuecomment-219788965, we should include a cleanup script for the DB in `bin/database.py`:
```
db.acquisitions.update_many({'timestamp':''}, {'$unset': {'timestamp': ''}})
db.sessions.update_many({'timestamp':''}, {'$unset': {'timestamp': ''}})
```

requirements.txt
nagem: Yep, I found out through an arduous search that the date-time validation inside `jsonschema` requires it but silently fails if it's not present.

nagem:
Retested, merging in 15 mins.

PR #305, opened 2016-05-17
bootstrap exit 1 if http requests fail

ryansanford:
@gsfr 
Intent is to have bootstrap.py exit non-zero if DB state doesn't match intent of provided json file. With this addition, callers can proceed with confidence on exit=0


ryansanford:
@gsfr 

This PR now alters pattern to insert/update rather than check/act. Please review.

gsfr:
LGTM

PR #306, opened 2016-05-18
Remove SCITRAN_RUNTIME_PROTOCOL

ryansanford:
This variable was a temporary measure from https://github.com/scitran/core/pull/190 and never fully removed until now.


ryansanford:
@nagem and @gsfr -- can you test in your respective environments please including bootstrap process? Thanks.

nagem:
LGTM, my only question is the differences between `sample.config` and the `DEFAULT_CONFIG` in [config.py](https://github.com/scitran/core/blob/master/api/config.py#L23-L53). I'm curious why `RUNTIME` configs only exist in sample.config and if they are required to remain there or if setting `SCITRAN_SITE_API_URL` is enough. This might be more of a legacy question for @gsfr.

ryansanford:
@nagem https://github.com/scitran/core/pull/306/commits/6056ddfa9825c528022caf9045060165d7928050 adds a comment regarding the RUNTIME vars. Essentially, they are only consumed by bin/run.sh for it to setup a standalone local instance hosted by paster, and not meant to be consumed by the API.

Let me know if that doesn't answer your question.

gsfr:
👍 Tested.

PR #307, opened 2016-05-18
Delete unused schema method

### LINKED ISSUE #269: Audit all optional parameter default values for mutability
Today I learned about this: https://stackoverflow.com/questions/1132941/least-astonishment-in-python-the-mutable-default-argument 

Aggravating, but not without humour :smile: 

screenshot-area-2016-04-29-141313

This has already caused at least one bug - `tags=[]` in `jobs.queue_job` - and I know that I've personally written a decent amount of code with optional params. I don't know what else might need fixing.

For now, I'll leave this ticket assigned to me until I've cleared out `jobs.py` and `rules.py`, then probably assign to @nagem for handling at your discretion.

gsfr:
closes #269

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -316,9 +316,3 @@ class RequestHandler(webapp2.RequestHandler):
             }
         log.warning(str(self.uid) + ' ' + str(code) + ' ' + str(detail))
         webapp2.abort(code, detail=detail, **kwargs)
-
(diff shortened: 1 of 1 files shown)


kofalt:
LGTM

nagem:
LGTM2

PR #309, opened 2016-05-18
Advanced search + Analyses

rentzso:
Changes proposed in this pull request:
- add ElasticSearch advanced search
- add endpoint to add analyses to sessions, acquisitions, collections, projects

@nagem @gsfr could you review this? Search on notes depends on this change https://github.com/flywheel-io/flywheel/pull/207
It won't return an error though, just no result.

PR #246, opened 2016-04-13
Bring API up-to-date with DM v2 documentation

nagem:
Closes #189, #379

List of changes available in issue ticket linked above. Still need to address "Ensure all italicized fields are returned in all GET requests (list and singular)", but since this is a large change, getting eyes on it and testing done sooner rather than later will help.

PR #311, opened 2016-05-19
add analyses endpoint

rentzso:
This is the second PR replacing #309

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -266,6 +266,23 @@ routes = [
     webapp2.Route(_format(r'/api/search/<cont_name:{cont_name_re}>'),                 searchhandler.SearchHandler, name='es_proxy', methods=['GET']),
     webapp2.Route(_format(r'/api/schemas/<schema:{schema_re}>'),                      schemahandler.SchemaHandler, name='schemas', methods=['GET']),
     webapp2.Route(r'/api/report/<report_type:site|project>',                          reporthandler.ReportHandler, methods=['GET']),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>'),
+                                                                                      listhandler.AnalysesHandler, name='analysis_post', methods=['POST']),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>'),
+                                                                                      listhandler.AnalysesHandler, name='analysis',
+                                                                                      methods=['GET', 'DELETE']),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/files'),
+                                                                                      listhandler.AnalysesHandler, handler_method='download',
+                                                                                      methods=['GET'], name='analysis_files'),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/files/<name:{filename_re}>'),
+                                                                                      listhandler.AnalysesHandler,
+                                                                                      handler_method='download', name='analysis_single_file'),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/notes'),
+                                                                                      listhandler.AnalysesHandler, handler_method='add_note',
+                                                                                      methods=['POST'], name='analysis_add_note'),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/notes/<note_id:{cid_re}>'),
+                                                                                      listhandler.AnalysesHandler, handler_method='delete_note',
+                                                                                      methods=['DELETE'], name='analysis_delete_note'),
 ]
 
 
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -182,3 +182,49 @@ class StringListStorage(ListStorage):
         result = self.dbc.find_one(query, projection)
         if result and result.get(self.list_name):
             return result.get(self.list_name)[0]
+
+
+class AnalysesStorage(ListStorage):
+
+    def get_fileinfo(self, _id, analysis_id, filename = None):
+        _id = bson.ObjectId(_id)
+        query = [
+            {'$match': {'_id' : _id}},
+            {'$unwind': '$' + self.list_name},
+            {'$match': {self.list_name+ '._id' : analysis_id}},
+            {'$unwind': '$' + self.list_name + '.files'}
+        ]
+        if filename:
+            query.append(
+                {'$match': {self.list_name + '.files.name' : filename}}
+            )
+        log.error(query)
+        return [cont['analyses'] for cont in self.dbc.aggregate(query)]
+
+    def add_note(self, _id, analysis_id, payload):
+        _id = bson.ObjectId(_id)
+        query = {
+            '_id': _id,
+            'analyses._id': analysis_id
+        }
+        update = {
+            '$push': {
+                'analyses.$.notes': payload
+            }
+        }
+        return self.dbc.update_one(query, update)
+
+    def delete_note(self, _id, analysis_id, note_id):
+        _id = bson.ObjectId(_id)
+        query = {
+            '_id': _id,
+            'analyses._id': analysis_id
+        }
+        update = {
+            '$pull': {
+                'analyses.$.notes': {
+                    '_id': note_id
+                }
+            }
+        }
+        return self.dbc.update_one(query, update)
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -198,7 +198,6 @@ class AnalysesStorage(ListStorage):
             query.append(
                 {'$match': {self.list_name + '.files.name' : filename}}
             )
-        log.error(query)
         return [cont['analyses'] for cont in self.dbc.aggregate(query)]
 
     def add_note(self, _id, analysis_id, payload):
--- a/api/download.py
+++ b/api/download.py
@@ -47,6 +47,32 @@ def _append_targets(targets, container, prefix, total_size, total_cnt, optional,
                 total_cnt += 1
     return total_size, total_cnt
 
+def symlinkarchivestream(ticket, data_path):
+    for filepath, arcpath, _ in ticket['target']:
+        t = tarfile.TarInfo(name=arcpath)
+        t.type = tarfile.SYMTYPE
+        t.linkname = os.path.relpath(filepath, data_path)
+        yield t.tobuf()
+    stream = cStringIO.StringIO()
+    with tarfile.open(mode='w|', fileobj=stream) as archive:
+        pass
+    yield stream.getvalue() # get tar stream trailer
+    stream.close()
+
+def archivestream(ticket):
+    BLOCKSIZE = 512
+    CHUNKSIZE = 2**20  # stream files in 1MB chunks
+    stream = cStringIO.StringIO()
+    with tarfile.open(mode='w|', fileobj=stream) as archive:
+        for filepath, arcpath, _ in ticket['target']:
+            yield archive.gettarinfo(filepath, arcpath).tobuf()
+            with open(filepath, 'rb') as fd:
+                for chunk in iter(lambda: fd.read(CHUNKSIZE), ''):
+                    yield chunk
+                if len(chunk) % BLOCKSIZE != 0:
+                    yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
+    yield stream.getvalue() # get tar stream trailer
+    stream.close()
 
 class Download(base.RequestHandler):
 
@@ -125,33 +151,6 @@ class Download(base.RequestHandler):
         used_subpaths[parent_id] = used_subpaths.get(parent_id, []) + [path]
         return path
 
-    def _archivestream(self, ticket):
-        BLOCKSIZE = 512
-        CHUNKSIZE = 2**20  # stream files in 1MB chunks
-        stream = cStringIO.StringIO()
-        with tarfile.open(mode='w|', fileobj=stream) as archive:
-            for filepath, arcpath, _ in ticket['target']:
-                yield archive.gettarinfo(filepath, arcpath).tobuf()
-                with open(filepath, 'rb') as fd:
-                    for chunk in iter(lambda: fd.read(CHUNKSIZE), ''):
-                        yield chunk
-                    if len(chunk) % BLOCKSIZE != 0:
-                        yield (BLOCKSIZE - (len(chunk) % BLOCKSIZE)) * b'\0'
-        yield stream.getvalue() # get tar stream trailer
-        stream.close()
-
-    def _symlinkarchivestream(self, ticket, data_path):
-        for filepath, arcpath, _ in ticket['target']:
-            t = tarfile.TarInfo(name=arcpath)
-            t.type = tarfile.SYMTYPE
-            t.linkname = os.path.relpath(filepath, data_path)
-            yield t.tobuf()
-        stream = cStringIO.StringIO()
-        with tarfile.open(mode='w|', fileobj=stream) as archive:
-            pass
-        yield stream.getvalue() # get tar stream trailer
-        stream.close()
-
     def download(self):
         """
         .. http:get::  /api/download
@@ -207,9 +206,9 @@ class Download(base.RequestHandler):
             if ticket['ip'] != self.request.client_addr:
                 self.abort(400, 'ticket not for this source IP')
             if self.get_param('symlinks'):
-                self.response.app_iter = self._symlinkarchivestream(ticket, config.get_item('persistent', 'data_path'))
+                self.response.app_iter = symlinkarchivestream(ticket, config.get_item('persistent', 'data_path'))
             else:
-                self.response.app_iter = self._archivestream(ticket)
+                self.response.app_iter = archivestream(ticket)
             self.response.headers['Content-Type'] = 'application/octet-stream'
             self.response.headers['Content-Disposition'] = 'attachment; filename=' + str(ticket['filename'])
             for project_id in ticket['projects']:
--- a/api/handlers/listhandler.py
+++ b/api/handlers/listhandler.py
@@ -13,6 +13,7 @@ from .. import files
 from ..jobs import rules
 from .. import tempdir as tempfile
 from .. import upload
+from .. import download
 from .. import util
 from .. import validators
 from ..auth import listauth, always_ok
@@ -61,6 +62,13 @@ def initialize_list_configurations():
             'storage_schema_file': 'note.json',
             'input_schema_file': 'note.json'
         },
+        'analyses': {
+            'storage': liststorage.AnalysesStorage,
+            'permchecker': listauth.default_sublist,
+            'use_object_id': True,
+            'storage_schema_file': 'analysis.json',
+            'input_schema_file': 'analysis.json'
+        }
     }
     list_handler_configurations = {
         'groups': {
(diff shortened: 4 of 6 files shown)


nagem:
Thanks for splitting these up, @rentzso!

rentzso:
No problem @nagem 
Thank you for reviewing this two monster PRs !

api/download.py
```diff
@@ -125,33 +151,6 @@ def _find_new_path(path, list_used_subpaths):
...
-            t = tarfile.TarInfo(name=arcpath)
-            t.type = tarfile.SYMTYPE
-            t.linkname = os.path.relpath(filepath, data_path)
-            yield t.tobuf()
-        stream = cStringIO.StringIO()
-        with tarfile.open(mode='w|', fileobj=stream) as archive:
-            pass
-        yield stream.getvalue() # get tar stream trailer
-        stream.close()
-
```
kofalt: Was this change required as part of this PR?

Am I reading it correctly that the only change is a slight rename, and moving out of the `Download` class?

api/handlers/listhandler.py
```diff
@@ -576,3 +584,132 @@ def packfile_end(self, cont_name, **kwargs):
...
         return upload.process_upload(self.request, upload.Strategy.packfile, origin=self.origin, context={'token': token_id}, response=self.response, metadata=metadata)
+
+class AnalysesHandler(ListHandler):
+
+    def _check_ticket(self, ticket_id, _id, filename):
+        ticket = config.db.downloads.find_one({'_id': ticket_id})
+        if not ticket:
+            self.abort(404, 'no such ticket')
+        if ticket['ip'] != self.request.client_addr:
+            self.abort(400, 'ticket not for this source IP')
```
kofalt: Question: Is IP-matching of a ticket a new policy decision? If not, I would prefer we add it.
If it does exist, then this is good & let's keep it.

api/handlers/listhandler.py
```diff
@@ -576,3 +584,132 @@ def packfile_end(self, cont_name, **kwargs):
...
+        if not filename:
+            return self._check_ticket_for_batch(ticket)
+        if ticket.get('filename') != filename or ticket['target'] != _id:
+            self.abort(400, 'ticket not for this resource')
+        return ticket
```
kofalt: Is there a ticket-checking code somewhere else we could re-use?

api/handlers/listhandler.py
```diff
@@ -576,3 +584,132 @@ def packfile_end(self, cont_name, **kwargs):
...
+                util.path_from_hash(fileinfo['hash'])
+            )
+            filename = fileinfo['name']
+            self.response.app_iter = open(filepath, 'rb')
+            self.response.headers['Content-Length'] = str(fileinfo['size']) # must be set after setting app_iter
+            if self.is_true('view'):
+                self.response.headers['Content-Type'] = str(fileinfo.get('mimetype', 'application/octet-stream'))
+            else:
+                self.response.headers['Content-Type'] = 'application/octet-stream'
+                self.response.headers['Content-Disposition'] = 'attachment; filename=' + str(filename)
```
kofalt: Is the download behaviour here different than other containers? If so, can you explain & document?

api/api.py
```diff
@@ -264,6 +264,23 @@ def _format(route):
...
+                                                                                      listhandler.AnalysesHandler,
+                                                                                      handler_method='download', name='analysis_single_file'),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/notes'),
+                                                                                      listhandler.AnalysesHandler, handler_method='add_note',
+                                                                                      methods=['POST'], name='analysis_add_note'),
+    webapp2.Route(_format(r'/api/<cont_name:{cont_name_re}>/<cid:{cid_re}>/<list_name:analyses>/<_id:{cid_re}>/notes/<note_id:{cid_re}>'),
+                                                                                      listhandler.AnalysesHandler, handler_method='delete_note',
+                                                                                      methods=['DELETE'], name='analysis_delete_note'),
```
kofalt: One of the things Ryan has requested is that all new routes have Sphinx documentation before merging, and that's worked well for us so far. I'd like to see working docs for each of these routes - there's some good example in the [job handlers](https://github.com/scitran/core/blob/master/api/jobs/handlers.py).

api/handlers/listhandler.py
```diff
@@ -576,3 +584,132 @@ def packfile_end(self, cont_name, **kwargs):
...
+            self.abort(400, 'ticket not for this resource')
+        return ticket
+
+    def _check_ticket_for_batch(self, ticket):
+        if ticket.get('type') != 'batch':
+            self.abort(400, 'ticket not for this resource')
+        return ticket
+
+    def put(self, *args, **kwargs):
+        raise NotImplementedError("an analysis can't be modified")
```
kofalt: Interesting type choice here - does `NotImplemented` imply that Analyses will be mutable in the future? That seems to conflict with the error message text.

api/handlers/listhandler.py
```diff
@@ -576,3 +584,132 @@ def packfile_end(self, cont_name, **kwargs):
...
+        input_validator(payload, 'POST')
+        payload['_id'] = str(bson.objectid.ObjectId())
+        payload['user'] = payload.get('user', self.uid)
+        payload['created'] = datetime.datetime.utcnow()
+        permchecker(noop)('POST', _id=_id)
+        result = storage.add_note(_id=_id, analysis_id=analysis_id, payload=payload)
+        if result.modified_count == 1:
+            return {'modified':result.modified_count}
+        else:
+            self.abort(404, 'Element not added in list {} of container {} {}'.format(storage.list_name, storage.cont_name, _id))
```
kofalt: Sorry to comment on similar stuff a couple times, but this is a lot of new code. Can you help me understand why this can't be code reuse from other container / list code?

api/placer.py
```diff
@@ -82,13 +82,15 @@ def save_file(self, field, info):
         """
 
         # Save file
-        files.move_form_file_field_into_cas(field)
+        if field is not None:
+            files.move_form_file_field_into_cas(field)
```
kofalt: Why call `save_file` function at all if it's not being passed a file to save?

api/placer.py
```diff
@@ -82,13 +82,15 @@ def save_file(self, field, info):
...
 
         # Update the DB
-        hierarchy.upsert_fileinfo(self.container_type, self.id, info)
+        if info is not None:
+            hierarchy.upsert_fileinfo(self.container_type, self.id, info)
 
-        # Queue any jobs as a result of this upload
-        rules.create_jobs(config.db, self.container, self.container_type, info)
+            # Queue any jobs as a result of this upload
+            rules.create_jobs(config.db, self.container, self.container_type, info)
```
kofalt: This change would mean that gears will not be scheduled to run if the file does not have info. That might work for our current default set of rules, but they can arbitrarily modified on each instance's project docs. I think this breaks how gears are scheduled.

api/handlers/listhandler.py
```diff
@@ -61,6 +62,13 @@ def initialize_list_configurations():
             'storage_schema_file': 'note.json',
             'input_schema_file': 'note.json'
         },
+        'analyses': {
+            'storage': liststorage.AnalysesStorage,
+            'permchecker': listauth.default_sublist,
+            'use_object_id': True,
+            'storage_schema_file': 'analysis.json',
+            'input_schema_file': 'analysis.json'
```
nagem: I don't see this schema file in the PR, maybe it was left behind in the split? Also, if you have time, it would be nice to update the Data Model v2 wiki and add the Analysis object there.

api/placer.py
```diff
@@ -475,3 +477,32 @@ def finalize(self):
...
+
+class AnalysisPlacer(Placer):
+
+    def check(self):
+        self.requireMetadata()
+        #validators.validate_data(self.metadata, 'analysys.json', 'input', 'POST', optional=True)
+        self.saved = []
+
+    def process_file_field(self, field, info):
+        self.save_file(field)
```
kofalt: If `info` is now an optional parameter to `save_file`, may as well pass it along.

api/placer.py
```diff
@@ -475,3 +477,32 @@ def finalize(self):
             'event': 'result',
             'data': result,
         })
+
+class AnalysisPlacer(Placer):
```
kofalt: Nice :+1:

api/upload.py
```diff
@@ -23,6 +23,7 @@
     'packfile'   : pl.PackfilePlacer,   # Upload N files as a new packfile to a container.
     'labelupload': pl.LabelPlacer,
     'uidupload'  : pl.UIDPlacer,
+    'analysis' : pl.AnalysisPlacer    # Upload N files for an analysys (no db updates)
```
kofalt: There's a documentation table a few lines below this in `process_upload`, that has a feature matrix for each placer.
Could you add a column to that table for the analysis placer?

api/download.py
rentzso: yes, in this way they will be available also for analyses downloads

api/handlers/listhandler.py
rentzso: we have it for regular downloads

api/handlers/listhandler.py
rentzso: for downloads we check the ticket but the common code is little. 
Ticket-checking code seemed overkill at this point.

api/handlers/listhandler.py
rentzso: will do.

api/handlers/listhandler.py
rentzso: Generally Analyses should be considered immutable, but we already allow to add notes to them. For now, there is no other field mutable, but there could be, depending on how the [data model]( https://github.com/scitran/core/wiki/Data-Model,-v3#analysis-draft) will evolve.

If this is confusing I can change the Error type and add a more open error message.

api/handlers/listhandler.py
rentzso: No worries. I understand it's difficult to review all this code.
Thanks for doing it!

However, analyses are nested object different from collections/sessions/projects/acquisitions
So for example here you need two ids to add a note to an analysis the id of the container and the id of the analysis while previously you needed only the container id.
Does it make sense?

api/handlers/listhandler.py
rentzso: Good catch. The data model for analyses is specified [here]( https://github.com/scitran/core/wiki/Data-Model,-v3#analysis-draft), but it's still a draft. 

To avoid blocking clients, I didn't add the schema.
I will add one though.

api/handlers/listhandler.py
kofalt: Hmm, that makes more sense. I am not yet sold on the sub-document part, but it might be that I don't fully understand the feature. This makes more sense given how it's currently designed.

kofalt:
Incidentally, because this change is large, if you could avoid force-pushing that will make it a lot easier to review. Up to you, I know I force-push a lot haha :)

api/placer.py
rentzso: Yes renaming it and splitting could be the right thing, but I didn't want to modify/break the behavior of other placers.

For analyses, for example, I needed a different update logic for the file info, so I could reuse only the `move file` part of the save file.

api/placer.py
rentzso: Don't you need at least to pass the name of the file in the info?
Users can still send no info at all, if I am not missing something.

api/placer.py
rentzso: Passing info will update the db and trigger job creation (in the new implementation of `save_file` 😄).
DB operations will be done together in the finalize step and, for this first implementation of analysis, automatic job creation has been disabled to limit the complexity (and impact) of the change. 
But it is something we should definetely discuss in some dedicated issue.

rentzso:
Agreed, no force-pushing!

api/handlers/listhandler.py
nagem: Ah I didn't realize we had moved on to v3, thanks!

api/dao/liststorage.py
```diff
@@ -182,3 +182,49 @@ def _get_el(self, _id, query_params):
...
+class AnalysesStorage(ListStorage):
+
+    def get_fileinfo(self, _id, analysis_id, filename = None):
+        _id = bson.ObjectId(_id)
+        query = [
+            {'$match': {'_id' : _id}},
+            {'$unwind': '$' + self.list_name},
+            {'$match': {self.list_name+ '._id' : analysis_id}},
+            {'$unwind': '$' + self.list_name + '.files'}
+        ]
```
nagem: It was probably already discussed at some point, but we might want to consider moving analysis objects to their own container level to prevent mongo difficulties when dealing with a list inside a list. For example, it would be impossible to edit a specific field of one of these files in an update query, as using `$` more than once in mongo is not allowed. Something to think about for future improvements. I would add an extra layer to our already complex hierarchy, though.

nagem:
I'll take another pass at reviewing this on Monday but it LGTM as a first implementation of Analyses. My only comment from an architectural standpoint would be that we consider moving Analyses out of a subdocument space and into their own container so any additional enhancements aren't limited by what mongo can do in a list of a list (or having to pull the object into python and send it back). Unfortunately if we do that we'll break a lot of `cont_name + s` code 😉.

Nice work, @rentzso.

gsfr:
I think @nagem makes a good point regarding nested subdocument arrays.

@rentzso IIRC, Elasticsearch considerations went into the current design. What do you think about promoting analyses to their own collection?

Also, I think we should keep the `+ s` convention alive by choosing a different name, such as *computation*.

rentzso:
Yes. Elasticsearch was the main factor as having analyses external to containers increases the complexity of some searches on analyses, at least in its current design. The feature is not in any PR yet, but it probably will.

However, Changing this now could be a big time investment on something that (from first tests) seems to work fine.
I understand the concern for the double nested list, but mongo doesn't block updates on nested lists. I was concerned too at first. :)
But with mongo aggregation pipelines, it doesn't seem a big deal...
Let me know though if you have performance concerns.

Another thing that needs to be evaluated is if for each analysis there will be always a related container. If not the nested option could be problematic.

About the singular vs. plural issue I don't understand exactly why singular names arrived in the API. 
Mongo Collections are plural so why can't we make everything plural? :)
It could be grammatically incorrect sometimes, but it should save us some headaches. Or am I missing something?

A big thank to @nagem and @kofalt for reviewing this!

gsfr:
As stated before, I am firmly opposed to compromising on the modeling and implementation of our primary database in order to simplify the integration of external features, such as Elasticsearch.

Btw, two main takeaways from my visit to MongoDB office hours:
- design your data model such that common operations are simple and elegant
- stay away from nested arrays

👍 on grammatically incorrect, but simplified level naming.

kofalt:
For what it's worth Renzo, I'm certainly the cause of some of the plural versus singular mix - a `job.input.x.type` would be `acquisition` not `acquisitions`, for example, and that's reflected in various parts of the code. If the team decides otherwise we could change things, but that's low ROI :/

My personal assumption has been that, eventually, plural nouns will only be used for the URL routes (with plurals then stripped before app logic) and for mongo table names (added at the very end, ideally by helper functions), and be singular everywhere else. 

But I've largely left it alone because we have larger fish to fry :)

rentzso:
@kofalt thanks for the explanation. If we all agree, we should track this on a tech-debt issue to "make everything plural".

@gsfr the main reason was Elasticsearch, but it wasn't the only reason. 😄 
Having analyses as a subdocument was the simpler solution. Permission handling was trivial and, as analyses at the beginning were immutable, also their sublists were supposed to be immutable. Unfortunately, now, we have notes (and tags) that are not.

I am not opposed (in principle) to putting analyses in a separate collection, but please note that the current implementation seems to work well and refactoring will take time (3-4 man-days).

I am definitely interested in understanding more the reasons why we should avoid nested arrays.
Can we have another mongodb office hours when I am back or after OHBM? 
It could also be a good time to ask if what I did with the mongodb aggregation is a good/acceptable or bad practice. I have already followed this advise (cc: @nagem @kofalt): https://docs.mongodb.com/manual/core/aggregation-pipeline/#early-filtering 

Let me know what you think should be the next step on this issue.

nagem:
@rentzso I agree that currently we don't hit any issues with analyses being a subdocument (other than a slightly awkward file lookup with aggregations). This is the particular bit of [mongo doc](https://docs.mongodb.com/manual/reference/operator/update/positional/) that would make editing anything in these subarrays difficult:

>The positional $ operator cannot be used for queries which traverse more than one array, such as queries that traverse arrays nested within other arrays, because the replacement for the $ placeholder is a single value

Some potential tricky situations that may or may not apply depending on how immutable analyses are: 
  - With the addition tags, if they follow the group tagging pool logic (a tag renamed in the group pool is renamed everywhere), we would have to pull the tag and reinsert it. 
  - Any modification to a file (`name`, `type`, `measurement`, ...)
  - Modifying the contents of a note
  - Schema upgrades already have issues with updating an array of objects (must be done in code), but we can currently treat all containers the same (see [this](https://github.com/scitran/core/pull/246/files#diff-837f544139dd99be7b888e157752643bR249) update script)

nagem:
After a discussion with @gsfr and @rentzso, the plan for edits to the first iteration (this PR) of Analyses is as follows, and I would appreciate @kofalt, @lmperry, and @dpuccetti to weigh in on any problems they see.

One major point is there are two types of analysis, one is tied to a job internally and one whose outputs are created externally and uploaded by a user. This PR should support both. One major outstanding question is whether we denormalize data from the job object to the analysis object for internal analyses. 

## Major changes:
  - An analysis object is going to be a standalone container with its own Mongo collection and based on the [base container](https://github.com/scitran/core/wiki/Data-Model,-v3#container) in our data model
  - Move `AnalysesHandler`to its own handler file with plans to merge it into `ContainerHandler`in the future, but leaving the endpoints the same for now
  - Use existing logic in `ListHandler` where possible
  - Require `POST` to have parent container type and parent container id
  - `outputs` field will become `files` to match base class

## Analysis object:

*Italicized* fields are inherited from the base container. 

Field           | Type                  | Description
--------------- | --------------------- | -----------
*_id*           | string                | database ID
parent_id   | string                | parent's database id ID
parent_cont  | string                | projects/collections/sessions
*label*           | string                | label
description   | string                | description
user        | string                | user id of the submitter (automatically set)
version      | string                | version id (unknown if this is necessary or helpful)
job         | string                | identifier of the job that generated the analysis
inputs †      | [File, …]             | array of File objects with the analysis file inputs*
*created*       | timestamp             | creation time (automatically set)
*modified*      | timestamp             | modification time (automatically updated)
*notes*         | [Note, …]             | array of Note objects
*files* †         | [File, …]             | array of File objects that represent analysis file outputs*
*views*         | integer               | **not supported in v1 of analyses**
*downloads*     | integer               | **not supported in v1 of analyses**
*permissions*  | [Permission, …]       | array of Permission objects inherited from parent container
*public*          | boolean               | **not supported in v1 of analyses**
*archived*        | boolean               | **not supported in v1 of analyses**
*tags*          | [string, …]           | array of tags
*info*         | object                | map of arbitrary information

† These will exist on the job object if internal analysis. 


## API endpoints:

API endpoint | Method | Behavior/Use
----------------- | ------------- | ------------
`/api/analysis/<id>` | `GET` | retrieve analysis by `id`
`/api/analysis` | `POST` | create an analysis, must include `parent_id` and `parent_container`
`/api/analysis/<id>` | `DELETE` | delete analysis
`/api/analysis/files` | `GET` | retrieve file info for analysis
`/api/analysis/files/<filename>` | `GET` | download file
`/api/analysis/notes/` | `POST` | add note
`/api/analysis/notes/<note-id>` | `DELETE` | delete note by id
`/api/<collections|projects|sessions>/<id>/analysis/` | `GET` | retrieve analysis for container
`/api/upload/analysis` | `POST` | upload files and metadata related to analysis output. Used by engine (results of job) or UI (external analysis)

nagem:
Another option for dealing with duplicated info from the `job` object is to merge the info from the job into the `inputs` and `files` arrays during a `GET`call to an analysis object, and/or just inflate the job object from the id and add it to the returned analysis object.

kofalt:
Based on the PR so far, it seems as though rules will not run on analysis containers. If that's the case, I request that the changes to `Placer.save_file` be reverted, and the new Placer implementation can just call `files.move_form_file_field_into_cas(field)` directly. That will keep things clearer, IMO.

kofalt:
I dislike duplicating object data, but I've also never heard of this "internal" use case before now. I don't have many comments on that because I don't feel I understand the feature. 

If we do copy-paste, `a.inputs` should match the job doc, IMO. I would press for `outputs` to do the same, but I guess that doesn't make sense as this container will actually hold the files itself. That means, implicitly, and file on the doc is a gear/whatever output. Should we disable "Add Data"? 

Overheard some rumbling about immutability. Where did we land on this? Which keys can be modified? Will permissions work differently in that regard? Whichever keys are mutable should not be duplicates.

Suggest removing `user` and adding `origin`. More flexible & consistent.

nagem:
After a talk with @gsfr, 

 - No, rules will not run on analyses
 - We will not denormalize data (copy/paste) from the job, but will fill out those fields from the job for `GETs`. Internal (job) analyses will be a lot more sparse than external analyses but they should both look nearly the same when returned from the api
 - Internal (job) analysis will have immutable inputs and outputs

nagem:
Rebased.

nagem:
I think we have a lot to think about for tweaks we can make to inputs and outputs and how they align with the inputs and outputs of jobs. What we already have is a great start and should probably be merged in as is, with incremental tweaks (and any necessary schema upgrades) happening as we make more decisions on how analyses will be used and stored. 

Overall, great work @rentzso!

nagem:
Rebased and removed some debug logging, ready to merge when CI is green.

PR #323, opened 2016-05-24
Allow file upload to set guess-type parameter to trigger type hueristics

kofalt:
On request from @dpuccetti and @lmperry - allow an uploading client to set `guess-type=true` on any upload endpoint (that uses a Placer - ref #159) and set a data type based on extension heuristic. Handily avoids duplicating [this map](https://github.com/scitran/reaper/blob/travis/reaper/types.json) in three different places, and trivial to implement so I just went ahead :+1: 

Both asked if this could just be the default, but I figured a safer option for now is to hide it behind a flag. We can always make it the default later.

### DIFF
--- a/api/files.py
+++ b/api/files.py
@@ -240,31 +240,35 @@ class MultiFileStore(object):
                     }, os.path.join(dest_path, filename))
 
 
-FILE_EXTENSIONS = {
-    "archive":      [ ".zip", ".tbz2", ".tar.gz", ".tbz", ".tar.bz2", ".tgz", ".tar", ".txz", ".tar.xz" ],
-    "document":     [ ".docx", ".doc" ],
-    "image":        [ ".jpg", ".tif", ".jpeg", ".gif", ".bmp", ".png", ".tiff" ],
-    "pdf":          [ ".pdf" ],
-    "presentation": [ ".ppt", ".pptx" ],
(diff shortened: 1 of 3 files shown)


gsfr:
Great to centralize this into the API. I'll remove from reaper after this lands.

api/files.py
```diff
@@ -238,3 +238,33 @@ def _save_multipart_files(self, dest_path, hash_alg):
...
+    "nifti":        [ ".nii.gz", ".nii" ],
+    "pfile":        [ ".7.gz", ".7" ]
+}
+
+def guess_type_from_filename(filename):
+    for key in FILE_EXTENSIONS:
+        for extension in FILE_EXTENSIONS[key]:
+            if filename.endswith(extension):
+                return key
+    return None
```
gsfr: This may not work as intended with "double" extensions. `foo.gephysio.zip` may come back as `archive` or `gephysio`. Compare with [alternative implementation in reaper](https://github.com/scitran/reaper/blob/master/reaper/folder_uploader.py#L24-L38).

kofalt: Hah! Good point.

nagem:
LGTM, other than Gunnar's comment.

kofalt:
Made extension precedence explicit, and added unit tests :+1:

kofalt:
Landing based on conversation with @nagem

gsfr:
👍

PR #324, opened 2016-05-24
Google style docstring support

ryansanford:
Documentation generator support of google style docstrings.

Reference:
http://www.sphinx-doc.org/en/stable/ext/napoleon.html#module-sphinx.ext.napoleon
http://google.github.io/styleguide/pyguide.html#Comments


kofalt:
Beautiful. That is amazing 10/10 would document again :tada:

Issue #268, opened 2016-04-28
Move jobs.py const MAX_ATTEMPTS to config struct


kofalt:
Not sure if this would involve a schema change?

kofalt:
Closed in #326

PR #326, opened 2016-05-25
Add queue configuration options

### LINKED ISSUE #314: Disable job retry when explicitly failed
Retrying a job is an important & necessary feature when a job is orphaned.
This contrasts with all other failure scenarios that I'm currently aware of, which are not transient:

1. Gear is invalid
1. API-level incompatibilities
1. Bad environment (gears not available, corrupt, code issue, etc)

In all these scenarios, it doesn't make sense to spam a bunch of job failures.

I propose that we disable the [Queue.mutate](https://github.com/scitran/core/blob/582dffac908d0de250758babfda74f0e012e50d9/api/jobs/queue.py#L67-L69) retry logic.
The [restart-orphan](https://github.com/scitran/core/blob/582dffac908d0de250758babfda74f0e012e50d9/api/jobs/queue.py#L214-L216) behaviour would remain the same.

Relevant to @lmperry, @ryansanford. Assigning to @gsfr for team discussion.

kofalt:
Closes #314. #268.

Adds two configuration options: One to limit how many times a job will be retried (under any circumstances), and one to toggle retrying of jobs that were explicitly failed (to prevent retry spam).

New runtime configuration for @ryansanford & @lmperry.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -36,6 +36,10 @@ DEFAULT_CONFIG = {
         'registered': False,
         'ssl_cert': None,
     },
+    'queue': {
+        'max_retries': 3,
+        'retry_on_fail': False,
+    },
     'auth': {
         'client_id': '1052740023071-n20pk8h5uepdua3r8971pc6jrf25lvee.apps.googleusercontent.com',
         'id_endpoint': 'https://www.googleapis.com/plus/v1/people/me/openIdConnect',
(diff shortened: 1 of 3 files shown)


api/config.py
```diff
@@ -36,6 +36,10 @@
         'registered': False,
         'ssl_cert': None,
     },
+    'queue': {
+        'max_retries': 3,
+        'retry_on_fail': False,
+    },
```
nagem: Make sure to add the changes to `sample.config`, otherwise LGTM!

kofalt: Added

kofalt:
Landing based on conversation with @nagem

PR #327, opened 2016-05-25
Add extra error checking around upgrade_to_8

### LINKED ISSUE #325: DB schema update is failing from version 2 DB
End result is that there is no version document in the `singletons` collection, and the `version` collection still exists with a value of 2.

bin/database.py prints nothing on stderror or stdout, and return code is 0.

nagem:
Fix #325.

@ryansanford I tested moving from bali 3.1 to this commit without any issue, please test yourself to confirm.

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -234,7 +234,9 @@ def upgrade_to_8():
     """
 
     if 'singletons' not in config.db.collection_names():
-        config.db.singletons.insert_many(config.db.static.find({}))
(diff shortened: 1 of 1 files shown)


kofalt:
LGTM

ryansanford:
Confirmed internally that latest update fixes the bad state, and upgrades from a good state work as well. Thanks @nagem ! I'm merging.

PR #329, opened 2016-05-26
Allow timestamps to be set as null

nagem:
@dpuccetti requested this for when users clear a timestamp from a session. @rentzso and @ryansanford will this negate the effort we did to remove all blank string timestamps for Elastic or will it handle nulls correctly?


rentzso:
Nulls should be ok. Empty strings were problematic.

PR #333, opened 2016-06-03
Job-based Analyses

### LINKED ISSUE #332: Job-based Analyses
Allow for the creation of analyses that are tied to a specific job, with the analysis being the job's output destination. This will be a first pass to get things working for now. A reworking of both types of analyses, using what we learned from these two prototypes, is expected to be done soon after. 

 - [x] Create a new endpoint that creates a new analysis and job, then ties them together via the job field on an analysis and the destination of the job
 - [x] Create a new logic flow for `/api/engine` where `level=analysis`
 - [x] Create a new placer that inherits from AnalysisPlacer, used for uploads coming from the engine
 - [x] Modify all `GETs` to return both types of analyses the same way
 - [x] Modify download endpoint to work for job-based analyses

nagem:
Closes #332.

Allows the creation of analyses tied to jobs. Reuses as much of existing analyis code as possible. Consider this code a working prototype with intention to rework. 

screen shot 2016-06-06 at 3 25 34 pm
screen shot 2016-06-06 at 3 24 55 pm

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -59,7 +59,7 @@ def inflate_job_info(analysis):
         job = next_job
     if job['_id'] != analysis['job']:
         # Update analysis if job has changed
-        q = {'analyses._id': str(self.id)}
+        q = {'analyses._id': analysis['_id']}
         u = {'$set': {'analyses.$.job': job['_id']}}
         config.db.sessions.update_one(q, u)
     analysis['job'] = job
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -5,6 +5,9 @@ from ..auth import INTEGER_ROLES
 
 CONT_TYPES = ['acquisition', 'analysis', 'collection', 'group', 'project', 'session']
 
+def getPerm(name):
+    return INTEGER_ROLES[name]
+
 def add_id_to_subject(subject, pid):
     """
     Add a mongo id field to given subject object (dict)
@@ -33,9 +36,50 @@ def add_id_to_subject(subject, pid):
         subject['_id'] = bson.ObjectId()
     return subject
 
+def inflate_job_info(analysis):
+    """
+    Inflate job from id ref in analysis
 
-def getPerm(name):
-    return INTEGER_ROLES[name]
+    Lookup job via id stored on analysis
+    Lookup input filerefs and inflate into files array with 'input': True
+    If job is in failed state, look for most recent job referencing this analysis
+    Update analysis if new job is found
+    """
+    if analysis.get('job') is None:
+        return analysis
+    job = config.db.jobs.find_one({'_id': analysis['job']})
+    if job is None:
+        raise Exception('No job with id {} found.'.format(analysis['job']))
+
+    # If the job currently tied to the analysis failed, try to find one that didn't
+    while job.get('state') == 'failed' and job.get('_id') is not None:
+        next_job = config.db.jobs.find_one({'previous_job_id': job['_id']})
+        if next_job is None:
+            break
+        job = next_job
+    if job['_id'] != analysis['job']:
+        # Update analysis if job has changed
+        q = {'analyses._id': str(self.id)}
+        u = {'$set': {'analyses.$.job': job['_id']}}
+        config.db.sessions.update_one(q, u)
+    analysis['job'] = job
+
+    # Inflate files from job inputs, add to analysis file array
+    files = analysis.get('files', [])
+    inputs = job.get('inputs', {})
+    for i in inputs.keys():
+        fileref = create_filereference_from_dictionary(inputs[i])
+        contref = create_containerreference_from_filereference(fileref)
+        file_ = contref.find_file(fileref.name)
+        if file_:
+            file_['input'] = True
+            files.append(file_)
+
+    analysis['files'] = files
+    return analysis
+
+
+
 
 class ContainerReference(object):
     def __init__(self, type, id):
@@ -64,6 +108,13 @@ class ContainerReference(object):
         if result is None:
             raise Exception("No such " + self.type + " " + self.id + " in database")
         return result
+
+    def find_file(self, filename):
+        cont = self.get()
+        for f in cont.get('files', []):
+            if f['name'] == filename:
+                return f
+        return None
 
     def check_access(self, userID, perm_name):
         perm = getPerm(perm_name)
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -3,6 +3,8 @@ import bson.objectid
 from .. import config
 from ..auth import INTEGER_ROLES
 
+CONT_TYPES = ['acquisition', 'analysis', 'collection', 'group', 'project', 'session']
+
 def add_id_to_subject(subject, pid):
     """
     Add a mongo id field to given subject object (dict)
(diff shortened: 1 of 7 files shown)


nagem:
@dpuccetti FYI.

nagem:
An example request:

### URL
`https://<base_url>/api/sessions/574f4297258d9f82a18c8a99/analyses?job=true`

### JSON body
```
{
    "analysis": {
        "label": "Test Analysis 1"
    },
    "job" : {
        "gear": "dcm_convert",
        "inputs": {
            "dicom": {
                "type": "acquisition",
                "id": "<acquisition_id>",
                "name" : "test_acquisition_dicom.zip"
            }
        },
        "tags": ["analysis"]
    }
}
```

The `analysis` and `job` submaps are required. The `job` map follows the same format as the `/api/jobs/add` endpoint, but destination is not required and would be ignored.

nagem:
@rentzso I'd appreciate a review from you when you get a chance.

kofalt:
This code looks good to me, great work!
I defer to @rentzso or @gsfr for analyses review.

gsfr:
Thanks, @kofalt. Hoping that @rentzso and @lmperry will take a look today.

api/dao/containerutil.py
```diff
@@ -31,14 +36,55 @@ def add_id_to_subject(subject, pid):
...
+    Lookup job via id stored on analysis
+    Lookup input filerefs and inflate into files array with 'input': True
+    If job is in failed state, look for most recent job referencing this analysis
+    Update analysis if new job is found
+    """
+    if analysis.get('job') is None:
+        return analysis
+    job = config.db.jobs.find_one({'_id': analysis['job']})
+    if job is None:
+        raise Exception('No job with id {} found.'.format(analysis['job']))
```
nagem: We could also set `job` to `None` here and let the consumer deal with an improper Analysis object. This is called on all analysis for an object so we might not want to bring the whole thing down for one failure. 

If we run into this error we probably have much larger problems, though.

api/handlers/listhandler.py
```diff
@@ -607,19 +610,64 @@ def _check_ticket_for_batch(self, ticket):
...
+            if cont_name == 'sessions':
+                return self._create_job_and_analysis(cont_name, _id, storage)
+            else:
+                self.abort(400, 'Analysis created via a job must be at the session level')
+
         payload = upload.process_upload(self.request, upload.Strategy.analysis, origin=self.origin)
-        payload['created'] = datetime.datetime.utcnow()
-        payload['user'] = payload.get('user', self.uid)
+        default = self._default_analysis()
+        payload.update(default)
```
nagem: This should be the other way around, fixing.

api/handlers/listhandler.py
```diff
@@ -607,19 +610,64 @@ def _check_ticket_for_batch(self, ticket):
...
+
+    def _create_job_and_analysis(self, cont_name, cid, storage):
+        payload = self.request.json_body
+        analysis = payload.get('analysis')
+        job = payload.get('job')
+        if job is None or analysis is None:
+            self.abort(400, 'POST json body must contain map for "analysis" and "job"')
+
+        default = self._default_analysis()
+        analysis.update(default)
```
nagem: Here too.

api/dao/containerutil.py
kofalt: I agree with your decision.

ryansanford:
@nagem 
Please incorporate the example as sphinx-doc to `listhandler.AnalysesHandler.post()`. There's an example you can template from with `listhandler.FileListHandler.get()`.

What's the effort to add an integration test for this route?

nagem:
@ryansanford working on an integ test for engine placer now, will incorporate this change.

rentzso:
@nagem I'll review it in the morning

nagem:
Docs added:
screen shot 2016-06-06 at 12 39 26 pm

api/placer.py
```diff
@@ -504,5 +504,21 @@ def finalize(self):
             metadata_info = metadata_infos.get(info['name'], {})
             metadata_info.update(info)
             self.metadata['files'].append(metadata_info)
-        self.metadata['_id'] = str(bson.objectid.ObjectId())
         return self.metadata
+
+class AnalysisJobPlacer(AnalysisPlacer):
+    def check(self):
+        super(AnalysisJobPlacer, self).check()
+        self.metadata['outputs'] = self.metadata['acquisition'].pop('files', [])
```
rentzso: I guess this is a temporary hack to adapt the engine upload. Correct? :)

nagem: I've talked with @kofalt recently about making the metadata output more generic, but yeah definitely a hack "for now" 😎

api/placer.py
```diff
@@ -504,5 +504,21 @@ def finalize(self):
...
+
+    def finalize(self):
+        super(AnalysisJobPlacer, self).finalize()
+        # Search the sessions table for analysis, replace file field
+        if self.metadata.get('files'):
```
rentzso: Shouldn't this be `'outputs'`?

nagem: I call `super` first which modifies the metadata to have a `files` array.

api/placer.py
```diff
@@ -504,5 +504,21 @@ def finalize(self):
...
-        self.metadata['_id'] = str(bson.objectid.ObjectId())
         return self.metadata
+
+class AnalysisJobPlacer(AnalysisPlacer):
+    def check(self):
+        super(AnalysisJobPlacer, self).check()
+        self.metadata['outputs'] = self.metadata['acquisition'].pop('files', [])
+
+    def finalize(self):
+        super(AnalysisJobPlacer, self).finalize()
```
rentzso: if the AnalysisJobPlacer uses another schema, does it make sense to inherit from the AnalysisPlacer? Am I missing something?

api/handlers/listhandler.py
```diff
@@ -607,19 +610,64 @@ def _check_ticket_for_batch(self, ticket):
...
+        return analysis_obj
+
     def post(self, cont_name, list_name, **kwargs):
         _id = kwargs.pop('cid')
         container, permchecker, storage, mongo_validator, _, keycheck = self._initialize_request(cont_name, list_name, _id)
-
         permchecker(noop)('POST', _id=_id)
+
+        if self.is_true('job'):
+            if cont_name == 'sessions':
```
rentzso: What are the reasons for this limitation? Is it temporary?

api/placer.py
nagem: I inherit from `AnalysisPlacer` mostly to use the `finalize` logic, but also cuts down on some mild code reuse in `process_file_field` and `check`. Currently, neither of them use json schemas for metadata nor REST endpoints. Hoping to change that soon.

api/handlers/listhandler.py
nagem: In the requirements when talking with @elisapoquette. I would say it's a temporary requirement that job-based analyses are only at the session level, but I defer to her for if this is a lasting requirement or something for the first pass.

api/placer.py
rentzso: Now I got how you adapted the code for this use case. I still think it may be better to have a separate placer, but not a big deal.

rentzso:
LGTM 
looking forward for the final version!

nagem:
@ryansanford updated description with both doc entries. I'll bundle integ tests with `engine-placer` branch I'm finishing up if that's alright. 

@rentzso and @kofalt, thanks for the reviews!

Hoping to merge EOD today or early tomorrow.

ryansanford:
@nagem Thanks much, I like the docs, and integration tests in `engine-placer` branch is ok with me. 👍

PR #339, opened 2016-06-08
Allow user to limit search results

### LINKED ISSUE #337: Add search flag for searching all data
A user should be able to choose wether they want to search all data via advanced search or just data they have permission to view. Adding `?=all_data=true` to calls to `/api/search` will toggle this feature. The default will be `false`.

nagem:
Closes #337 

Using the 'all_data' flag, user can toggle wether to return all data or just data they have access to.

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -8,6 +8,13 @@ CONT_TYPES = ['acquisition', 'analysis', 'collection', 'group', 'project', 'sess
 def getPerm(name):
     return INTEGER_ROLES[name]
 
+def get_perm_name(cont_type):
+    if cont_type == 'group':
+        return 'roles'
+    if cont_type in ['projects', 'collections', 'sessions', 'acquisitions']:
+        return 'permissions'
+
+
 def add_id_to_subject(subject, pid):
     """
     Add a mongo id field to given subject object (dict)
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -8,13 +8,6 @@ CONT_TYPES = ['acquisition', 'analysis', 'collection', 'group', 'project', 'sess
 def getPerm(name):
     return INTEGER_ROLES[name]
 
-def get_perm_name(cont_type):
-    if cont_type == 'group':
-        return 'roles'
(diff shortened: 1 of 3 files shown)


nagem:
Still need to do more testing before ready to merge.

api/dao/containerutil.py
```diff
@@ -8,6 +8,13 @@
 def getPerm(name):
     return INTEGER_ROLES[name]
 
+def get_perm_name(cont_type):
```
rentzso: some cont_type are plural, some singular...

api/search/queryprocessor.py
```diff
@@ -26,21 +27,32 @@
...
             else:
                 self.child_targets.add(t)
-        self.results = None
+        if not all_data and user:
+            self.filter_for_user
+
+    def filter_for_user(self):
+        perm_name = get_perm_name(self.cont_name[:-1])
+        if perm_name is not None:
+            self.query = add_filter(self.query, perm_name+'_id', user)
```
rentzso: missing dot before `'_id'`?

rentzso:
a couple of typos but LGTM

rentzso:
as we were discussing groups permissions can create some edge cases though.
We will need to think about it...

rentzso:
A possible solution to any inconsistency is to put permissions on files and notes, in the mongo connector.

api/dao/containerutil.py
nagem: I take off the last char when coming from `queryrprocessor.py`, where they are plural. Seems like `containerutil.py` has been using singular, which is what I went with here. We really should decide on a standard. Also when analyses get their own perms, we'll have to rethink it all too.

nagem:
@rentzso I fixed the type and have it only filtering projects by permissions. Seems to work from the testing I've done!

rentzso:
cool! 
Feel free to merge it. You should just remove `get_perm_name` as you are not using it anymore.

nagem:
I left it because I've wanted to use that functionality before, but yeah I can just rewrite it if I end up needing it again. Not a good idea to leave some well intentioned but unused code around.
Will merge after CI goes green after dead code removal.

nagem:
@dpuccetti merging this one too.

PR #340, opened 2016-06-09
Change search verb from GET to POST

nagem:
Created a PR to document, many libraries and clients ignore entity bodies set on `GET` requests, [Angular being one of them](http://stackoverflow.com/questions/35184426/angular2-http-get-with-body) (Postman as well, curl allows it). HTTP docs have been more vague, [this SO](http://stackoverflow.com/a/15656853) having some discussion mentioning the ambiguity.

Therefore we're changing `api/search` requests to a `POST` rather than a `GET`.


nagem:
@dpuccetti FYI, merging now.

PR #342, opened 2016-06-10
Enable Warning checks with pylint

ryansanford:
Warning checks have been enabled which initially will cause the travis job to fail. Merge after all warnings have been addressed, and the travis job completes without error.

closes flywheel-io/flywheel#183


gsfr:
Any updates on this?

nagem:
It hasn't been prioritized in the current sprint, but I plan to chip away at it in the next two weeks

nagem:
I added `logging-format-interpolation` to the global ignore list as we do that *everywhere* and I don't see any major harm in it.

gsfr:
That's probably ok, although I fixed it throughout the reaper code, but it sure wasn't fun.

Here is a reference that I found with some further info: https://www.simonmweber.com/2014/11/24/python-logging-traps.html

nagem:
There are a few other issues that I line-by-line ignored because they would cause a messy refactor. Talking with @ryansanford today, I'm going to open up a "second pass" ticket that mentions the fixes we cut from this first pass. Getting pylint warnings enabled without too much refactoring or risky changes should be the goal of this ticket.

nagem:
The integ tests and pylint are happy and a short run-through bug bash doesn't turn up anymore errors, going to commit!

This was a pretty big diff even with a few `pylint-disable` cop-outs so we'll see if we can shake any other bugs out in the next week via normal use. Expect a followup ticket addressing any warnings solved with a `pylint-disable`.

gsfr:
Awesome, @nagem. Thanks for seeing this through.

PR #343, opened 2016-06-10
fix search processing bug

rentzso:
**Summary**
`self.query` should be updated even if it's not executed.
Exec one single query for each SearchContainer.


nagem:
LGTM!

PR #344, opened 2016-06-10
Augment search results

rentzso:
Search results will contain the full parent hierarchy

For example:
a file in an acquisition will have in its body the acquisition, the
parent session, the project and the groups.
a session will include the project and the group

This branch has been created from #343

### DIFF
--- a/api/handlers/searchhandler.py
+++ b/api/handlers/searchhandler.py
@@ -8,9 +8,10 @@ from ..search import pathparser, queryprocessor, es_query
 
 log = config.log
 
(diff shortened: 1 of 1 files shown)


nagem:
LGTM, other than removing some `log.error`s

rentzso:
Thanks Megan!
Just removed the `log.error`s. Merging...

rentzso:
@dpuccetti please note that this changes the search response schemas. It should be trivial to adapt the UI, but let me know if you need any help.

PR #345, opened 2016-06-15
Augment file types

lmperry:
Include a few more common source code and video formats, as well as other bvec and bval file extensions.


nagem:
LGTM.

PR #349, opened 2016-06-22
Have pip update itself in virtualenv first

tomck:
On macOS El Capitan, the virtualenv created uses an older version of pip which must compile the cryptography wheel. This fails per pyca/cryptography#2350. Having the virtualenv pip update itself before running resolves this issue which prevents ./bin/run.sh from completing "Installing Python requirements" on macOS 10.11.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -5,5 +5,6 @@ set -e
 unset CDPATH
 cd "$( dirname "${BASH_SOURCE[0]}" )/.."
 
+pip install -U pip
 pip install -r requirements.txt
 pip install -r requirements_dev.txt


kofalt:
LGTM

gsfr:
Thanks, @tomck.

Would it make sense to extend this a bit to cover other, similar use cases?
```
pip install -U pip setuptools wheel
```

tomck:
@gsfr I wouldn't do it in the same command at least, because the benefit of updating pip to an 8.x series is it downloads 'cryptography' as a binary instead of trying to compile it, and I don't know what will happen with 'setuptools' and 'wheel' when passed in the same command.

gsfr:
Actually, I don't understand why you are seeing this behavior, and I also don't think this is the right way to address it.

If you are using `run.sh`, homebrew should have installed a recent version of python and pip. Currently, that would be [pip 8.1.2](https://github.com/Homebrew/homebrew-core/blob/master/Formula/python.rb#L72-L75). I had 8.0.2 until a moment ago.

Maybe the right way to address this is to enforce a certain recentness of [python (and hence pip)](https://github.com/scitran/core/blob/master/bin/run.sh#L93-L99) in `run.sh`. Another option may be to add `pip>=8.0.x` to `requirements.txt`.

We also need to know from @ryansanford if `install.sh` plays a role in Flywheel's prod setup.

tomck:
I do use homebrew but added the pip update line after attempting to force the homebrew version of OpenSSL without luck. And the information about virtualenv using an old version of pip seems to corroborate it. http://stackoverflow.com/a/17495071

ryansanford:
@gsfr: `install.sh` Is not a dependency of the Flywheel managed dev/test/prod environments.

gsfr:
I'll merge this now, but it will probably be superseded soon by some larger infrastructure improvements. Thanks for catching this, @tomck.

PR #352, opened 2016-06-22
Add prefetch feature, defaults to off

### LINKED ISSUE #320: POST /gears/x adds a fetch job
After #319, add a job to the queue for prefetching of assets.
Requested by @ryansanford to minimize job latency.

kofalt:
Fixes #320. @ryansanford 

Adds a new option, `queue.prefetch`. If enabled, a sanity-check job is added to the queue when a gear is created or updated. This will cause a connected queue consumer to download the gear in question. The job is tagged `prefetch` for easy monitoring & early-warning.

Not a substitute for intelligent asset distribution & management; this is just a trivial fix for several users who only have one queue consumer to slightly decrease initial job latency. **Disabled by default.**

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -39,6 +39,7 @@ DEFAULT_CONFIG = {
     'queue': {
         'max_retries': 3,
         'retry_on_fail': False,
+        'prefetch': False
     },
     'auth': {
         'client_id': '1052740023071-n20pk8h5uepdua3r8971pc6jrf25lvee.apps.googleusercontent.com',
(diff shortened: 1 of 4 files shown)


rentzso:
Changes seem pretty harmless, but I don't fully understand the code.
I see that inputs and destination are optional now. Wondering if this could hide any issue.

kofalt:
Yeah, it's not ideal. A better approach might be a Job subclass that has different methods. The trick is knowing which class to instantiate when loading from the database. Probably a good idea for future improvements

nagem:
LGTM.

PR #354, opened 2016-06-23
Don't modify the dict when loading a job

kofalt:
Fixes a bug that crashes `/jobs/next` when running a job with a pre-existing request.

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -3,6 +3,7 @@ Jobs
 """
 
 import bson
+import copy
 import datetime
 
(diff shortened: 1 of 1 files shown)


api/jobs/jobs.py
```diff
@@ -76,9 +77,12 @@ def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previou
         self._id             = _id
 
     @classmethod
-    def load(cls, d):
+    def load(cls, e):
```
nagem: >e

._.

nagem:
LGTM

Issue #160, opened 2016-02-11
/engine endpoint should accept uploads to any container level


gsfr:
We currently only accept uploads to acquisitions.

We desire the following functionality:
- Uploads to an acquisition, its parent session and parent project
- Uploads to a session and its parent project
- Uploads to a project (not sure this is actually a thing)

kofalt:
Tentatively blocked by #159 (not a hard requirement, but might make life easier).

kofalt:
@nagem you were asking about this limitation earlier? Here's the ticket tracking that.

nagem:
As of #357, you are able to upload to an acquisition, a session, or a project. Uploading to multiple levels of the hierarchy at the same time, if still necessary, can be done in a separate PR.

kofalt:
SGTM. We don't have anything on the roadmap regarding gears simultaneously uploading to multiple levels of the hierarchy.

gsfr:
We should have our schemas enforce this limitation.

PR #357, opened 2016-06-28
Engine Placer

nagem:
Addresses scitran/core#159, scitran/core#160 and a few other requests made by team members.

Changes:
 - Engine uploads now use the Placer class style
 - Uploads can be made to projects, sessions, acquisitions or analyses*
 - Uploads can be files only
 - Uploads can be metadata only

*Work towards supporting uploads to multiple levels of the hierarchy at once (scitran/core#160) is being done in a separate branch

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -358,6 +358,7 @@ def _update_container(query, update, set_update, cont_name):
 
 def _update_hierarchy(container, container_type, metadata, update_timestamp=False):
     project_id = container['project'] # for sessions
+    now = datetime.datetime.utcnow()
 
     if container_type == 'acquisition':
         update = {}
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -344,8 +344,7 @@ def update_container_hierarchy(metadata, cid, container_type):
     if c_obj is None:
         raise APIStorageException('container does not exist')
     if container_type in ['session', 'acquisition']:
-        update_timestamp = True if c_metadata.get('timestamp') else False
-        _update_hierarchy(c_obj, container_type, metadata, update_timestamp)
+        _update_hierarchy(c_obj, container_type, metadata)
     return c_obj
 
 
@@ -357,34 +356,26 @@ def _update_container(query, update, set_update, container_type):
     )
 
 
-def _update_hierarchy(container, container_type, metadata, update_timestamp=False):
+def _update_hierarchy(container, container_type, metadata):
     project_id = container.get('project') # for sessions
     now = datetime.datetime.utcnow()
 
     if container_type == 'acquisition':
-        update = {}
         session = metadata.get('session', {})
         session_obj = None
-        if update_timestamp:
-            update['$min'] = dict(timestamp=container['timestamp'])
-            session['timezone'] = dict(timezone=container.get('timezone'))
-        if update.keys() or session.keys():
+        if session.keys():
             session['modified'] = now
-            session_obj = _update_container({'_id': container['session']}, update, session, 'sessions')
+            session_obj = _update_container({'_id': container['session']}, {}, session, 'sessions')
         if session_obj is None:
             session_obj = get_container('session', container['session'])
         project_id = session_obj['project']
 
     if project_id is None:
         raise APIStorageException('Failed to find project id in session obj')
-    update = {}
     project = metadata.get('project', {})
-    if update_timestamp:
-        update['$max'] = dict(timestamp=container['timestamp'])
-        project['timezone'] = dict(timezone=container.get('timezone'))
     if project.keys():
         project['modified'] = now
-        project_obj = _update_container({'_id': project_id}, update, project, 'projects')
+        project_obj = _update_container({'_id': project_id}, {}, project, 'projects')
 
 
 def merge_fileinfos(parsed_files, infos):
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -334,52 +334,53 @@ def dict_fileinfos(infos):
     return dict_infos
 
 
-def update_container_hierarchy(metadata, acquisition_id, level):
-    project = metadata.get('project')
-    session = metadata.get('session')
-    acquisition = metadata.get('acquisition')
+def update_container_hierarchy(metadata, cid, container_type):
+    c_metadata = metadata.get(container_type)
     now = datetime.datetime.utcnow()
-    if acquisition.get('timestamp'):
-        acquisition['timestamp'] = dateutil.parser.parse(acquisition['timestamp'])
-    acquisition['modified'] = now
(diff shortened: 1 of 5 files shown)


nagem:
Will update the endpoint documentation on next commit

api/dao/hierarchy.py
```diff
@@ -319,53 +319,59 @@ def dict_fileinfos(infos):
...
+    project_id = container.get('project') # for sessions
+    now = datetime.datetime.utcnow()
+
+    if container_type == 'acquisition':
+        update = {}
+        session = metadata.get('session', {})
+        session_obj = None
+        if update_timestamp:
+            update['$min'] = dict(timestamp=container['timestamp'])
+            session['timezone'] = dict(timezone=container.get('timezone'))
```
nagem: This min/max setting of the session and project timestamps was in the existing code. Unless someone has a reason to keep it, I'd like to remove it. @rentzso @gsfr

api/dao/hierarchy.py
```diff
@@ -319,53 +319,59 @@ def dict_fileinfos(infos):
     return dict_infos
 
 
-def update_container_hierarchy(metadata, acquisition_id, level):
-    project = metadata.get('project')
-    session = metadata.get('session')
-    acquisition = metadata.get('acquisition')
+def update_container_hierarchy(metadata, cid, container_type):
```
kofalt: Could you add a docstring with a few sentences describing what this does? I'm not sure if I've ever understood what (exactly) this function does without reverse-engineering it each time I read it :)

api/placer.py
```diff
@@ -188,17 +188,18 @@ class EnginePlacer(Placer):
 
     def check(self):
         self.requireTarget()
-        validators.validate_data(self.metadata, 'enginemetadata.json', 'input', 'POST', optional=True)
+        if self.metadata is not None:
+            log.debug('about to validate')
+            validators.validate_data(self.metadata, 'enginemetadata.json', 'input', 'POST', optional=True)
+            log.debug('validated')
```
nagem: Will clean up debug messages

api/dao/hierarchy.py
kofalt: Same comment applies to the other diff'd functions in this file

api/upload.py
```diff
@@ -134,7 +134,6 @@ def process_upload(request, strategy, container_type=None, id=None, origin=None,
     if placer.sse and not response:
         raise Exception("Programmer error: response required")
     elif placer.sse:
-        log.debug('SSE')
```
kofalt: :tada:

kofalt:
Looks good so far. Ping me when changes are done

kofalt:
```
  File "webapp2.py", line 570, in dispatch
  File "./api/upload.py", line 236, in engine
    return process_upload(self.request, Strategy.engine, container_type=level, id=cid, origin=self.origin)
  File "./api/upload.py", line 141, in process_upload
    return placer.finalize()
  File "./api/placer.py", line 219, in finalize
    self.obj = hierarchy.update_container_hierarchy(self.metadata, bid, self.container_type)
  File "./api/dao/hierarchy.py", line 333, in update_container_hierarchy
    _update_hierarchy(c_obj, container_type, metadata, update_timestamp)
  File "./api/dao/hierarchy.py", line 358, in _update_hierarchy
    session_obj = _update_container({'_id': container['session']}, update, session, 'sessions')
  File "./api/dao/hierarchy.py", line 341, in _update_container
    return_document=pymongo.collection.ReturnDocument.AFTER
  File "pymongo/collection.py", line 2155, in find_one_and_update
    sort, upsert, return_document, **kwargs)
  File "pymongo/collection.py", line 1937, in __find_and_modify
    allowable_errors=[_NO_OBJ_ERROR])
  File "pymongo/collection.py", line 205, in _command
    read_concern=read_concern)
  File "pymongo/pool.py", line 211, in command
    read_concern)
  File "pymongo/network.py", line 100, in command
    helpers._check_command_response(response_doc, msg, allowable_errors)
  File "pymongo/helpers.py", line 196, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)

OperationFailure: command SON([('findAndModify', u'sessions'), ('query', {'_id': ObjectId('5772dc4c669becae4325338f')}), ('new', True), ('update', {'$set': {u'subject.sex': u'male', u'subject.firstname_hash': None, u'timestamp': u'1970-01-01T00:00:00+00:00', u'subject.lastname': u'', 'modified': datetime.datetime(2016, 6, 28, 20, 23, 21, 805863), u'subject.firstname': u'', u'subject.lastname_hash': None, u'subject.age': 0.0, 'timezone.timezone': None, u'operator': None}, '$min': {'timestamp': datetime.datetime(1970, 1, 1, 0, 0)}}), ('upsert', False)]) on namespace scitran.$cmd failed: exception: Cannot update 'timestamp' and 'timestamp' at the same time
```

> Cannot update 'timestamp' and 'timestamp' at the same time

Mongo does not approve :persevere:

api/dao/hierarchy.py
gsfr: The UI renders the session timestamp, so I suggest we keep setting it. It is currently defined to be equal to the oldest acquisition timestamp.

The project timestamp is not used by the UI, as far as I know, and also what we are doing doesn't make a ton of sense, so we could get rid of that part.

nagem: > It is currently defined to be equal to the oldest acquisition timestamp.

Thanks, that's helpful information. A user can set the timestamp, would we want to overwrite a user's input? Unfortunately we wouldn't be able to differentiate between user input and machine input at this time (as we were discussing), so the question is more "is overwriting a session's timestamp with metadata ever okay?"

gsfr: After offline conversation, we conclude that the engine placer will only do as it is told and not have side effects, such as updating session timestamps without being explicitly told to do so. Reaper uploads are a different story, though. Side effect will remain there.

nagem:
Non-explicit timestamp setting removed. Will rerun a few tests using the engine and then this is ready for merge.

kofalt:
I'll give this a test today.

kofalt:
This update looks to have a bug that removes or overwrites container files on upload. An acquisition that had a dicom zip lost that file:

```json
    "files": [
        {
            "name": "1_1_nifti.nii.gz", 
            "origin": {
                "id": null, 
                "type": "unknown"
            }, 
            "type": "nifti"
        }
    ], 
```

nagem:
Fixed issue with metadata for files overwriting what's there during `update_hierarchy` step, which also fixed missing origin. Added a test for that specific situation.

Reruning the engine in a newly bootstrapped environment is also successful.

kofalt:
Latest changes work; LGTM

nagem:
Thanks for reviewing, merging now.

PR #358, opened 2016-06-28
Fix issue updating persisted config

### LINKED ISSUE #355: Changes in default config should be reflected in db config
Changes [here](https://github.com/scitran/core/blob/master/api/config.py#L23-L58) should be reflected in the config persisted to the db. A key error bug has led the team to believe that is not happening. More investigation needed.

nagem:
Fixes #355.

Python `dict.update()` does a shallow update, so any changes the default config other than at the top level were being overwritten by the database version of the config.

Some slight refactoring was necessary in order to have a util class that was free from any local imports.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -5,6 +5,7 @@ import logging
 import pymongo
 import datetime
 import elasticsearch
+from . import util
 
 
 logging.basicConfig(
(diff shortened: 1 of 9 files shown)


nagem:
@ryansanford This fix will make deployment easier, but unfortunately it comes with a necessary refactoring change that touches a lot of the api functionality. I've tested the change locally without issue, but a once-over of any use cases that involve checking permissions would be necessary before shipping this update.

nagem:
Any questions, talk to @kofalt in my absence as I updated him with some alternatives if this would need to much testing before deployment.

nagem:
This has been rebased to resolve merge conflicts. When CI is green it is ready to land, pylint should give errors if there was a mistake in the refactor.

PR #360, opened 2016-07-07
Metadata from upload endpoints should only set null fields

### LINKED ISSUE #359: Metadata from uploads should only set null fields
If metadata is sent to an upload endpoint (for example, `api/engine`), it should only update fields that are `null`. This is to prevent the engine or other upload paths from clearing out user edited information.

nagem:
Closes #359

Notes about efficiency/concurrency:
 - Pulling the doc into python, doing the update in code, and then using one db `update` was considered but not chosen as there is no (current) way in our architecture to lock a document
 - Using an unordered bulk request limits network traffic to mongo and also allows mongo to make its own optimizations with the update set
 - If an installation is using WiredTiger (available in mongo 3.0), document-level locks are available and will be used. Without WiredTiger, mongo will use a collection-level lock for this update. 
 - If the update includes metadata and the metadata field on the target container does not exist or is empty, the entirety of the metadata field is set at once for efficiency (thanks for the idea @ryansanford). This is to address DICOM headers coming in from gear output causing hundreds of update calls for each metadata field.

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -358,9 +358,21 @@ def _update_container(query, update, set_update, container_type):
 
 def update_container_nulls(base_query, update, container_type):
     coll_name = container_type if container_type.endswith('s') else container_type+'s'
-    update_dict = util.mongo_dict(update)
+    cont = config.db.[coll_name].find(base_query)
+    if cont is None:
+        raise APIStorageException('Failed to find {} object using the query: {}'.format(container_type, base_query))
+
     bulk = config.db[coll_name].initialize_unordered_bulk_op()
 
+    if (update.get('metadata') is not None and
+       (cont.get('metadata') is None or cont.get('metadata',{}).keys == 0)):
+        # If we are trying to update metadata fields and
+        # the container metadata does not exist or is empty,
+        # metadata can all be updated at once for efficiency
+        m_update = update.pop('metadata')
+        bulk.find(base_query).update_one({'$set': {'metadata': m_update}})
+
+    update_dict = util.mongo_dict(update)
     for k,v in update_dict.items():
         q = {}
         q.update(base_query)
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -182,83 +182,52 @@ def _find_or_create_destination_project(group_id, project_label, timestamp):
         )
     return project
 
-
-def _create_session_query(session, project, type_):
-    if type_ == 'label':
+def _create_query(cont, cont_type, parent_type, parent_id, upload_type):
+    if upload_type == 'label':
         return {
-            'label': session['label'],
-            'project': project['_id']
+            'label':        cont['label'],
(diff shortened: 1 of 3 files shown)


nagem:
This currently only supports engine upload but I wanted to get a PR open for discussion while I tie up the other upload endpoints that modify existing containers.

nagem:
Reaper upload endpoints (`api/upload/label` and `api/upload/uid`) now only update null values as well.

The old `reaper` upload method has been removed.

nagem:
This is ready to land after a code review.

api/dao/hierarchy.py
```diff
@@ -375,7 +335,33 @@ def _update_hierarchy(container, container_type, metadata):
...
+def _update_container_nulls(base_query, update, container_type):
+    coll_name = container_type if container_type.endswith('s') else container_type+'s'
+    cont = config.db[coll_name].find_one(base_query)
+    if cont is None:
+        raise APIStorageException('Failed to find {} object using the query: {}'.format(container_type, base_query))
+
+    bulk = config.db[coll_name].initialize_unordered_bulk_op()
+
+    if (update.get('metadata') is not None and
+       (cont.get('metadata') is None or cont.get('metadata',{}).keys == 0)):
```
rentzso: this condition could be replaced with 
```update.get('metadata') and not cont.get('metadata')```
as both `None` and `{}` are falsy in python.

File: api/dao/hierarchy.py [hunk 2]
```diff
@@ -375,7 +335,33 @@ def _update_hierarchy(container, container_type, metadata):
...
+        # If we are trying to update metadata fields and the container metadata does not exist or is empty,
+        # metadata can all be updated at once for efficiency
+        m_update = util.mongo_sanitize_fields(update.pop('metadata'))
+        bulk.find(base_query).update_one({'$set': {'metadata': m_update}})
```
rentzso: shouldn't the method return now?

rentzso: Oh I see now. You are popping the metadata field... 👍

api/dao/hierarchy.py
```diff
@@ -375,7 +335,33 @@ def _update_hierarchy(container, container_type, metadata):
...
+
+    update_dict = util.mongo_dict(update)
+    for k,v in update_dict.items():
+        q = {}
+        q.update(base_query)
+        q['$or'] = [{k: {'$exists': False}}, {k: None}]
+        u = {'$set': {k: v}}
+        log.debug('the query is {} and the update is {}'.format(q,u))
+        bulk.find(q).update_one(u)
+    bulk.execute()
```
rentzso: Didn't know the mongo bulk feature. It's very nice! 👍

rentzso:
LGTM pending one small comment

nagem:
Fixed the unnecessarily complex if statement. Running a data refresh and engine run locally to finish testing, afterwards I'll merge when CI is green.

PR #366, opened 2016-07-14
Restructure jobs

### LINKED ISSUE #362: Restructure job schema
`job` objects stored in Mongo have an `inputs` field structured as a dictionary. The arbitrary keys make it difficult to search the dictionary for inputs referencing a specific container, resulting in inefficient javascript queries like [this one](https://github.com/scitran/core/blob/master/api/jobs/queue.py#L188-L197). 

Restructuring the `inputs` field to be a list of inputs and setting a `type` field as the current dictionary key (for example, 'dicom') will allow us to remove this slow filter and replace it with an indexed lookup.

nagem:
Closes #362.

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -43,9 +43,8 @@ def inflate_job_info(analysis):
 
     # Inflate files from job inputs, add to analysis file array
     files = analysis.get('files', [])
-    for i in getattr(job, 'inputs', []):
-        i.pop('input')
-        fileref = create_filereference_from_dictionary(i)
+    for i in getattr(job, 'inputs',{}):
+        fileref = create_filereference_from_dictionary(job.inputs[i])
         contref = create_containerreference_from_filereference(fileref)
         file_ = contref.find_file(fileref.name)
         if file_:
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -30,7 +30,7 @@ def inflate_job_info(analysis):
 
     # If the job currently tied to the analysis failed, try to find one that didn't
     while job.state == 'failed' and job._id is not None:
-        next_job = config.db.jobs.find_one({'previous_job_id': job['_id']})
+        next_job = config.db.jobs.find_one({'previous_job_id': job._id})
         if next_job is None:
             break
         job = Job.load(next_job)
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -112,19 +112,20 @@ class Job(object):
         Flatten struct to map
         """
 
-        d = self.__dict__
+        # Don't modify the job obj
+        d = copy.deepcopy(self.__dict__)
 
-        if d.get('inputs', None):
+        if d.get('inputs'):
             for x in d['inputs'].keys():
                 d['inputs'][x] = d['inputs'][x].__dict__
-
-        if d.get('destination', None):
-            d['destination'] = d['destination'].__dict__
-
-        if d['destination'] is None:
-            d.pop('destination')
-        if d['inputs'] is None:
+        else:
             d.pop('inputs')
+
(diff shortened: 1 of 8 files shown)


nagem:
A bit more work is needed, the update to `containerutil.py` to remove `inflate_job_info` was not committed by mistake.

api/jobs/jobs.py
```diff
@@ -12,6 +12,48 @@
...
+    for i in getattr(job, 'inputs',{}):
+        fileref = create_filereference_from_dictionary(job.inputs[i])
+        contref = create_containerreference_from_filereference(fileref)
+        file_ = contref.find_file(fileref.name)
+        if file_:
+            file_['input'] = True
+            files.append(file_)
+
+    analysis['files'] = files
+    return analysis
```
kofalt: Could you move this function to live with the analysis code instead? 
It's calling Job functions, rather than being one. Similar to `get_jobs` elsewhere in the diff

api/jobs/jobs.py
```diff
@@ -83,10 +125,14 @@ def load(cls, e):
         # Don't modify the map
         d = copy.deepcopy(e)
 
-        if d.get('inputs', None):
-            inputs = d['inputs']
-            for x in inputs.keys():
-                inputs[x] = create_filereference_from_dictionary(inputs[x])
+        if d.get('inputs'):
```
kofalt: I take it the `, None` that you removed from this `get` call is unnecessary?
My memory of these map semantics is hazy.

kofalt:
LGTM so far.

api/jobs/jobs.py
```diff
@@ -83,10 +125,14 @@ def load(cls, e):
...
+            input_dict = {}
+
+            for i in d['inputs']:
```
rentzso: Can you elaborate a bit on what this piece of code is doing?
If  two values from `d['inputs']` have the same `input` key, one will override the other. Is it possible/acceptable?

nagem: That is true, but if we always interact with jobs through the Job class instead of directly with the mongo object we won't have an issue, as it is a map/dict. Mongo has the option of a unique index key, but it does almost the exact opposite of what we want:

https://docs.mongodb.com/manual/core/index-unique/#unique-constraint-across-separate-documents

nagem: Using this we could have multiple inputs of `input` 'dicom' in one list but it wouldn't be allowed in any other document.

nagem:
I'm running into a tricky pass-by-ref encoder bug when there is more than one input to a job referencing the same container. Short description: when the custom encoder is iterating over the jobs returned in `get_jobs`, it is hitting a job whose inputs are already converted to a map. 

I'm going to give it a weekend off and fix it on Monday to preserve my own sanity 😎.

nagem:
The bug has been fix, `job.map()` needs to make a copy of `self.__dict___` if the intended usage is to return a map of a job, and not convert the job object to a map.

api/jobs/jobs.py
nagem: I'm not sure why we have it everywhere, as `None`, `''`, `{}` and `[]` are all falsey in Python. I'm going to assume I might have put it a few places before I knew better :).

api/jobs/jobs.py
nagem: I realize I didn't explain what the intent was. With this conversion we can maintain a job structure in Mongo that makes lookups on inputs efficient but can also have a job structure via a Python obj that maintains uniqueness of input keys and allows us to restructure the mongo schema without a bit refactor.

nagem:
Retested with existing data, rebootstrap, running the engine and analyses. Merging now.

nagem:
@ryansanford Note: this does have a database upgrade script.

PR #368, opened 2016-07-18
Build context

lmperry:


lmperry:
This resolves the autobuild failure by placing the Dockerfile in the root of the repo.

ryansanford:
@lmperry @gsfr I support the movement of the Dockerfile. I'll update the file `docker/README.md` to reflect the new path.

Any objections to creating a symlink from the old location to make the transition easier? The progression in my head is...
1. Add symlink and merge.
2. Update Dev environment for new location
3. Give devs a week to merge this into or rebase scitran/core feature branches, then update their dev environment.
4. Pull out the symlink

For Flywheel to move to a model where we consume docker images rather than the github repo:
- I prefer a model where images are only posted to dockerhub when CI tests pass.
- I'd hate to lose the injection of commit hash, and build time into the image. This is currently handled via BUILD vars defined in the Dockerfile.

I'll throw together a branch tomorrow to address the above bullets via travis. Let's talk this afternoon if that direction is not amenable.

gsfr:
@ryansanford Good plan! Let's discuss how we split tasks between Travis and DockerHub. CI doing only CI - fast - seems desirable. Image builds should be contingent on successful CI, but not necessarily part of it.

ryansanford:
Thanks for the input @gsfr . I won't do anything with travis until we talk. In the mean time I'll add the symlink and change to docker/README.md

gsfr:
Is this ready to land?

ryansanford:
This is ready to land. Look good to you @gsfr ?

gsfr:
👍

PR #369, opened 2016-07-19
Fix propagation bug related to missing s

### LINKED ISSUE #367: Permissions not propagated when moving a session from one project to another
If you move a session from one project to another, the permissions of the target project are not assigned to the session's acquisitions, thus making that session appear empty when viewed.

cc @ryansanford @gsfr

nagem:
Fixes #367.

There was a mistype between `session` and `sessions` in the propagation query for acquisitions.

I also fixed the propagation tests that should have caught this. The tests were showing a false positive due to `public` not being a propagated property from the session level.


nagem:
Will merge when CI is green

PR #372, opened 2016-07-20
Store inputs on analyses objects

nagem:
To aid in searching analyses objects, we will store the job inputs as file objects on the analyses object.

### DIFF
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -257,20 +257,22 @@ class AnalysesStorage(ListStorage):
             job = Job.load(next_job)
         if job._id != analysis['job']:
             # Update analysis if job has changed
+            # Remove old inputs and replace with new job inputs
+            # (In practice these should never change)
+            files = analysis.get('files', [])
+            files[:] = [x for x in files if x.get('output')]
+
+            for i in getattr(job, 'inputs',{}):
+                fileref = job.inputs[i]
+                contref = containerutil.create_containerreference_from_filereference(job.inputs[i])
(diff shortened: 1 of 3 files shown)


kofalt:
I'm not thrilled by duplication, but Megan's convinced me for now. Diff tentatively looks good but I defer to @rentzso for review.

If we merge this, I would like to see a ticket filed that tracks either the removal of this duplication or the conclusion that we'll keep duplicating in this & other cases. My understanding is that empowering the mongo connector to use the core (import via requirements.txt? merge repos?) could avoid this in the long run.

nagem:
@kofalt another option you and I discussed was the possibility of the mongo-es replicator calling the API to get the correct format it needs.

rentzso:
To clarify, one of the reasons for this change is that it gives a more consistent and self-contained format to each analysis in the database. 
This enable tools like mongo connectors to read directly the mongo oplog, and to replicate asynchronously but almost instantly the state of the database without any additional load on the API (@megan calling the API to replicate all the records could be ok, but I think it would be better to avoid any additional overload, and also someone will have to write all the code, almost from scratch, to export data from mongo to elasticsearch).

I am not sure that the merge solution will be very practical either: let's say that we need another tool that reads from the oplog. This tool code will also need to be merged to core, and so on...

Data duplication seems the lesser evil, but @kofalt feel free to open a ticket to investigate on this and on the possible implications of data duplication.

rentzso:
LGTM!

nagem:
Retested the db update script, everything looks good. Merging now.

PR #377, opened 2016-07-27
Check permissions when downloading containers

nagem:
Adds permissions checks to container downloads. Uses simple approach to pass over containers where the user does not have view rights. Also passes over containers requested but not found instead of failing. If at the end of the preflight stage no containers are found, a `404` is returned.

Discussion around a more usable reaction to missing or non-permitted containers has been on-going and will continue after this PR.

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -7,10 +7,9 @@ import datetime
 import cStringIO
 
 from . import base
-from . import validators
-
-from . import util
 from . import config
+from . import util
+from . import validators
 
 log = config.log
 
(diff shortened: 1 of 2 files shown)


rentzso:
LGTM

nagem:
Going to fix broken test as well as beef up what tests we have for download as part of this ticket.

gsfr:
👍

nagem:
Fixed up false positive test, will add more tests as separate ticket to get this one through. Merging now.

PR #384, opened 2016-07-29
Add file type for Philips PAR/REC format

lmperry:
Having Philips PAR/REC files in the system with their own format `parrec` will allow specific classifiers and converters to be used on them.


nagem:
LGTM!

PR #386, opened 2016-08-02
Add initial support for user API keys

nagem:
**Request new/replacement API key**
A user will not have an API key until they request one. If the user wishes to revoke their existing API key, they may do so by requesting a new one at this endpoint.

***Request***
```
POST /api/users/self/key HTTP/1.1
(empty request body)
```
***Response***
```
{
  "key": "uCmHDCnNdej9WbLiv_T0VnrbWzMP_-02czad6M7f2A_3cdx4l2V9RbRO"
}
```

**Use API key in a request**
Authorization header expects `scitran-user` and the API key delimitated by a ` `(space) character. The API key information will only be returned via the `api/users/self` endpoint, never by `api/users/` or `api/users/<uid>` even with super-user requests.

***Request***
```
GET /api/users/self HTTP/1.1
Authorization: scitran-user uCmHDCnNdej9WbLiv_T0VnrbWzMP_-02czad6M7f2A_3cdx4l2V9RbRO
```
***Response***
`last_used` will be `null` when the key has not been used yet. 
```
{
  "api_key": {
    "last_used": "2016-08-08T16:36:11.548000+00:00",
    "key": "uCmHDCnNdej9WbLiv_T0VnrbWzMP_-02czad6M7f2A_3cdx4l2V9RbRO",
    "created": "2016-08-08T16:27:15.539000+00:00"
  },
  "firstname": "Example",
 ...
```

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -198,6 +198,7 @@ routes = [
     webapp2_extras.routes.PathPrefixRoute(r'/api/users', [
         webapp2.Route(r'/self',                                 userhandler.UserHandler, handler_method='self', methods=['GET']),
         webapp2.Route(r'/self/avatar',                          userhandler.UserHandler, handler_method='self_avatar', methods=['GET']),
+        webapp2.Route(r'/self/key',                             userhandler.UserHandler, handler_method='generate_api_key',methods=['POST']),
         webapp2.Route(_format(r'/<_id:{user_id_re}>'),          userhandler.UserHandler, name='user'),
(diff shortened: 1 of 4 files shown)


gsfr:
LGTM!

gsfr:
@nagem: Can we get away without a DB update script and just let the UI handle a missing `api_key` field?

nagem:
That was my plan. I'm writing up some documentation on how to use this endpoint and then will merge.

PR #393, opened 2016-08-11
remove "id":"#" from schemas/input/download.json

coltonlw:
RAML API Workbench threw an error parsing the $ref's in download.json schema.  I believe the reason is the documents "id" parameter is set to "#" .  "id" needs to be a valid URI.    "id" does not need to be specified, and if id is not specified a ref that starts with "#" will reference the current document, so setting "id":"#" seems wrong to me.   http://json-schema.org/latest/json-schema-core.html#anchor27   EDIT:  From that link:  "This URI MUST be normalized, and SHOULD NOT be an empty fragment (#) or the empty URI."

### DIFF
--- a/api/schemas/input/download.json
+++ b/api/schemas/input/download.json
@@ -1,5 +1,4 @@
 {
-    "id": "#",
     "$schema": "http://json-schema.org/draft-04/schema#",
     "definitions": {
         "filterDefinition": {


rentzso:
LGTM 

Just tested some downloads without that line and it worked. 
At the time I worked on this, I had some issues with json schemas references and, I believe, that was one of the suggested solutions.

@nagem please take a look at this also as your ongoing changes could be impacted.

nagem:
@rentzo the download.json schema was not working* with my changes but I noticed it was also not properly filtering on master either. Looks like we needed to add an `additionalProperties: false` to `filterDefinition` and that fixed the problem. It is now working as expected on my branch. 

@coltonlw LGTM, also made this change in my branch that I hope to merge today, too. 

*it was letting filtering options other than `+` and `-` through

nagem:
PR #399 which also had this change has merged, going to close this one.

Issue #226, opened 2016-03-30
JSON schema improvements


kofalt:
Two goals:

1. Use a single source of truth
1. Eliminate need for network loopback.

Filed per request from team discussion, although this sat in my notes for awhile.

kofalt:
Loopback continues to be a problem in my dev environment :disappointed: 
Will be happy to see this fixed.

coltonlw:
@kofalt Would you consider this completed?

nagem:
>Use a single source of truth 

I think we have the structure for that in place with the schema hierarchy @coltonlw added.
Seems pretty done to me.

kofalt:
:+1:

PR #399, opened 2016-08-12
Remove api loopback request for schema ref resolution

nagem:
Addresses half of #226.

Also made a small bug fix to `download.json`.


coltonlw:
I know user schema ref's the avatar schema, it would be nice if we had some tests for these endpoints and you could wait a week or so until we do to merge this.  Otherwise LGTM

nagem:
It'll be nice when I don't have to manually test these, though 😄

coltonlw:
:octocat: yes it will, hopefully soon

rentzso:
LGTM!

Issue #394, opened 2016-08-11
Some endpoints do not validate input 


coltonlw:
This issue is to track endpoints which need to have input validation with jsonschema added:
- [ ] Jobs
- [ ] Gears
- [ ] Rules
- [ ] Search
- [x] POST /api/engine   (strategy "analysis")   AnalysisPlacer:  https://github.com/scitran/core/blob/master/api/placer.py#L505

@ryansanford

PR #400, opened 2016-08-12
AnalysisJobPlacer should not require metadata

nagem:
Not all gears produce metatdata


kofalt:
LGTM!

nagem:
In this ticket I also corrected an "error" caused by #399 where we were previously silently not loading a missing schema `analysis.json` but now erroring when it was not found. The long term fix would be adding an `analysis.json` schema, tracked by #394.

PR #403, opened 2016-08-15
Clear schema path from db to support PR #399

nagem:
Relates to PR #399.

The precedence order for config is env vars > config stored in db > default config, and should not change. In #399, we changed the `persistent.schema_path` config field to be an absolute path rather than a relative path. This PR adds a DB update script that removes the path in the db config to allow the default to persist to the db config, rather than getting overwritten by the old (assumed default*) value already there. 

*By the nature of this config field, a user entered value should not exist here.

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -11,7 +11,7 @@ from api import config
 from api.dao import containerutil
 from api.jobs.jobs import Job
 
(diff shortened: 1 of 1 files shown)


ryansanford:
@nagem 

LGTM. Infrequent upgrade issue still persists in our environment, but we have a documented workaround.

PR #411, opened 2016-08-17
Add 'now' flag to jobs

### LINKED ISSUE #109: Add 'now' flag to job entries
Proposed by @ryansanford as a preemptive production support feature.

Add a boolean `now` flag to job objects.
Toggling this flag will cause the job to run as soon as possible.

Change the job modification (post/put) route to only allow superusers to manipulate the field.

Change the `next` route to make two queries - first, check for any jobs with a `now` flag, second, the normal query that exists today. Critically, the `now` query should sort by the `modified` timestamp in the *opposite* order the normal query uses - that is, return a job that has **most** recently been modified and has a `now` flag. This contrasts with the normal query which is FIFO.

Should be set on endpoint tracked by #319.

nagem:
Closes #109

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -13,7 +13,7 @@ from .. import config
 log = config.log
 
 class Job(object):
-    def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previous_job_id=None, created=None, modified=None, state='pending', request=None, id_=None, config_=None):
+    def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previous_job_id=None, created=None, modified=None, state='pending', request=None, id_=None, config_=None, now=False):
         """
         Creates a job.
 
@@ -46,14 +46,14 @@ class Job(object):
 
         # TODO: validate inputs against the manifest
 
-        now = datetime.datetime.utcnow()
+        time_now = datetime.datetime.utcnow()
 
         if tags is None:
             tags = []
         if created is None:
-            created = now
+            created = time_now
         if modified is None:
-            modified = now
+            modified = time_now
 
         if destination is None and inputs is not None:
             # Grab an arbitrary input's container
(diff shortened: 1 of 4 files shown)


api/jobs/jobs.py
```diff
@@ -46,14 +46,14 @@ def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previou
 
         # TODO: validate inputs against the manifest
 
-        now = datetime.datetime.utcnow()
+        time_now = datetime.datetime.utcnow()
```
kofalt: Why was this renamed?

api/jobs/jobs.py
```diff
@@ -78,6 +78,7 @@ def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previou
         self.request         = request
         self.id_             = id_
         self.config          = config_
+        self.now             = now
```
kofalt: This looks to be using the old name?

kofalt:
Could we remove `now` from `map()` if `now` is False?

kofalt:
Should we add an index to the `now` flag? Would that first query result in a table scan?

(I assume we already have an index on the `modified` flag but I did not check.)

kofalt:
LGTM after comments. Did not test.

nagem:
Added a compound index that will support the `api/jobs/next` endpoint. @ryansanford let me know if you want to do any of your own testing, otherwise ready to merge.

ryansanford:
@nagem I'm testing now...

ryansanford:
@nagem Jobs without the now flag set are not being returned by `jobs/next` .  Also, new jobs from the UI and resulting from the dicom2nii gear don't appear to have that flag set in the db.

kofalt:
@ryansanford I wouldn't expect jobs queued by the UI and rules to have a now flag, since the intention is that it's only used for administrative overrides.

nagem:
@ryansanford bug fixed, I had removed the flag if it was `False` due to @kofalt's request, but my jobs used for testing still had the flag.

ryansanford:
LGTM

PR #417, opened 2016-08-19
Fix malformed timestamps

nagem:
This should cover all our bases for malformed timestamps in acquisitions and sessions. Amends db update `9` which only covered blank strings.

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -355,6 +355,8 @@ def _update_hierarchy(container, container_type, metadata):
         session_obj = None
         if session.keys():
             session['modified'] = now
+            if session.get('timestamp'):
+                session['timestamp'] = dateutil.parser.parse(session['timestamp'])
             session_obj = _update_container_nulls({'_id': container['session']},  session, 'sessions')
         if session_obj is None:
             session_obj = get_container('session', container['session'])
--- a/bin/database.py
+++ b/bin/database.py
@@ -436,7 +436,10 @@ def upgrade_schema():
         while db_version < CURRENT_DATABASE_VERSION:
             db_version += 1
             upgrade_script = 'upgrade_to_'+str(db_version)
-            globals().get(upgrade_script)()
+            globals()[upgrade_script]()
+    except KeyError as e:
+        logging.exception('Attempted to upgrade using script that does not exist: {}'.format(e))
+        sys.exit(1)
     except Exception as e:
         logging.exception('Incremental upgrade of db failed')
         sys.exit(1)
--- a/bin/database.py
+++ b/bin/database.py
@@ -11,7 +11,7 @@ from api import config
 from api.dao import containerutil
 from api.jobs.jobs import Job
 
-CURRENT_DATABASE_VERSION = 14 # An int that is bumped when a new schema change is made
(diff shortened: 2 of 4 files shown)


ryansanford:
@nagem String versions of the timestamp are still making their way into the session collection. Looks like this happens when the mr-classifier gear runs on an acquisition belonging to a session without an existing timestamp. There must be code that causes an acquisition timestamp update to trickle up to the session if the session timestamp is missing? Oddly, the acquisition timestamp is being set to a proper ISODate() format.

bin/database.py
```diff
@@ -387,36 +431,12 @@ def upgrade_schema():
...
-        if db_version < 13:
-            upgrade_to_13()
-        if db_version < 14:
-            upgrade_to_14()
 
+    try:
+        while db_version < CURRENT_DATABASE_VERSION:
+            db_version += 1
+            upgrade_script = 'upgrade_to_'+str(db_version)
+            globals().get(upgrade_script)()
```
nagem: @coltonlw @gsfr I finally got sick of this and made it dynamic, please let me know if I'm using the wrong kind of python voodoo 😄

api/dao/hierarchy.py
```diff
@@ -325,6 +325,8 @@ def _update_hierarchy(container, container_type, metadata):
         session_obj = None
         if session.keys():
             session['modified'] = now
+            if session.get('timestamp'):
+                session['timestamp'] = dateutil.parser.parse(session['timestamp'])
```
nagem: @ryansanford - I found the culprit! Thanks for narrowing the search down for me.

gsfr: Awesome! Thanks for tracking this down, @nagem & @ryansanford.

bin/database.py
gsfr: `globals()` seems to be the way to go for functions in the current module. Using the `globals()[upgrade_script]()` notation will produce a slightly more intuitive `KeyError` rather than a `TypeError: 'NoneType' object is not callable`. Might we worth catching either one of those explicitly and printing an appropriate message in order to expedite future debugging.

ryansanford:
LGTM -- I re-tested the Db update and confirm session timestamps are being set correctly via mr-classifier gear workflow.

coltonlw:
Adding some unit tests would be awesome

nagem:
I added some basic tests for `database.py`, will merge when ci green.

test/unit_tests/test_db_upgrade.py
```diff
@@ -0,0 +1,51 @@
...
+def test_all_upgrade_scripts_ran(database_mock_setup):
+    with pytest.raises(SystemExit):
+        database.upgrade_schema()
+    for i in range(1, CDV):
+        script_name = 'upgrade_to_{}'.format(i)
+        assert getattr(database, script_name).called
+
+@patch('bin.database.get_db_version', Mock(return_value=CDV-4))
+def test_necessary_upgrade_scripts_ran(database_mock_setup):
+    with pytest.raises(SystemExit):
```
nagem: I couldn't find a great way to check which actual sys.exit() code is returned, but ensuring the scripts are called gets us most of the way there.

PR #422, opened 2016-08-22
Make schema_path no longer user configurable

coltonlw:
Having schema_path be user configurable causes problems during changes, such as  PR #403.  This PR hardcodes schema_path, as there isn't a valid use case for changing it.


nagem:
LGTM

PR #425, opened 2016-08-22
Fix markdown files

coltonlw:
Change markdown files to unix encoding
Remove comment for dillinger.io from README

@gsfr


gsfr:
LGTM.

PR #428, opened 2016-08-23
Add update-only flag to upload endpoint

nagem:
Adds new query param for `/api/upload/uid` called `update-only`:

```
  queryParameters:
    update-only:
      type: boolean
      required: false
      description: If true, it will only upload to containers that already exist
```

Endpoint will return status code `404` if containers do not exist. For more info see raml docs in diff.


nagem:
@coltonlw I added you to the review as well so you can review any raml and abao changes I made as part of this PR.

test/integration_tests/abao/abao_test_hooks.js
```diff
@@ -106,6 +107,11 @@ hooks.before("GET /download -> 404", function(test, done) {
     done();
 });
 
+hooks.before("POST /upload/label -> 404", function(test, done) {
```
coltonlw: 🎉 👍

test/integration_tests/abao/abao_test_hooks.js
```diff
@@ -48,6 +48,7 @@ hooks.skip("POST /download -> 200");
 hooks.skip("GET /download -> 200");
 hooks.skip("POST /upload/label -> 200");
 hooks.skip("POST /upload/uid -> 200");
+hooks.skip("POST /upload/uid -> 404");
```
coltonlw: Can you add a comment on why this was skipped?  The others apply to the comment above about testing file fields, which now should probably say postman something

kofalt:
Rebased, testing now :)

nagem:
This has been reworked into a new endpoint. Query param is no longer required, there is a new [json schema]() and [raml doc](). The new endpoint is `/api/upload/uid-match`.

test/integration_tests/abao/abao_test_hooks.js
nagem: In order to cause a 404, we'll need to send metadata via a file form so I thought it lined up well enough.

kofalt:
Tested; LGTM

PR #433, opened 2016-08-25
Pin newman to 3.0.1

coltonlw:
Newly released newman 3.1.0 broke file fields

See also: https://github.com/postmanlabs/newman/issues/570


ryansanford:
LGTM

PR #434, opened 2016-08-25
add commit message guidelines

coltonlw:
Same changes form previous PR but manually fixed commit history

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,11 @@
 - Add docstrings to all functions with a one-line description of its purpose.
 
 ### Format
-Ensure that `./test/lint.sh api` exits without errors.
+- Ensure that `./test/lint.sh api` exits without errors.
 
+### Commit Messages
+1. The subject line should be a phrase describing the commit and limited to 50 characters
+1. The subject line should be as specific as possible, e.g. "Add new API resource for user api keys" not "Updated authentication.
 
 ## Adding an API Resource
 ### Design Resource


nagem:
LGTM

gsfr:
Does Github understand the *colons* when referencing issues? [The documentation](https://help.github.com/articles/closing-issues-via-commit-messages/) doesn't mention them.

CONTRIBUTING.md
```diff
@@ -12,8 +12,19 @@
...
+1. The subject line should be a phrase describing the commit, do not use punctuation.
+1. The subject line should be limited to 50 characters, capitalized, and not use punctuation.
+1. The subject line should use present tense / imperative mood  e.g. "Fix bug in foo.py", NOT "fixed" or "fixes".
+1. Following the subject is an optional detailed description, separated by a blank line.
+1. The description should be limited to 72 characters per line and use punctuation.
+1. The description should be broken into paragraphs limited to 4 sentences each.
+1. The description should state what you changed and why, less important is how.
+1. Following the description include a line containing "Resolves: #xxx, #yyy" for any issues fixed by the commit, separated by a blank line.
+1. Finally include a line containing "See also: #xxx, #yyy" for any other issues.
+1. These steps were adapted from [7 Rules of a Great Commit Message](http://chris.beams.io/posts/git-commit/#seven-rules).  Visit for more info.
```
gsfr: Let's drop the last sentence. Everyone should know the benefit of links. ;-)

kofalt:
@gsfr Is this a good place to discuss the merit of the guidelines, or has the decision already been made?

gsfr:
@kofalt No decision has been made. Let's discuss.

Personally, I would just reference "the 7 rules" and be done with it. All I really care about is the format of the first line. Do whatever works best below that.

coltonlw:
Yea I mean if we think it's unnecessary then let's not have commit message guidelines.  I was just trying to document our existing process for consistency but I can see how it's a little pedantic

kofalt:
For everyone's context, here are some recent commits from [Gunnar](https://github.com/scitran/core/commits?author=gsfr), [Megan](https://github.com/scitran/core/commits?author=nagem), [Renzo](https://github.com/scitran/core/commits?author=rentzso), and [myself](https://github.com/scitran/core/commits?author=kofalt). I won't mind guidelines as long as the majority of these would pass muster and we're pretty [relaxed](http://slinky.me/uploads/pic/8/slinky_me_54b2bdbb16d6a.gif) about it.

FWIW, I will never see anything following a newline except for my ~twice yearly usage of `git bisect`. So if you want to type a paragraph down there totally go for it :+1: I just won't see it. 

The only thing I actually care about is that I type a sentence or two, on the first line, and I will never notice or care about whichever arbitrary character-count cutoff someone's using today (this one asks for 50). This approach also works well with the Github PR view. The most important part of the message goes first so that, however you're viewing the graph today, you'll have something useful.

I'm a happy camper  as long as my PRs will be accepted despite having an apostrophe in the commit log :)

gsfr:
I continue to think that a firm (not hard) 50 character limit on the first line is nice as it makes for a more readable log.

Beyond that, I agree with @kofalt. I also never look past the first line.

PR #436, opened 2016-08-27
Set config.log.propagate to False

coltonlw:
Currently our logging setup duplicates messages.  This is because propagate is set to True by default on python logger objects.  This causes messages sent to the logger to propagate up and in our case hit the root logger which is configured to also log to stderr.  https://docs.python.org/2/library/logging.html#logging.Logger.propagate

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -14,6 +14,7 @@ logging.basicConfig(
     level=logging.DEBUG,
 )
 log = logging.getLogger('scitran.api')
+log.propagate = False
 
 logging.getLogger('MARKDOWN').setLevel(logging.WARNING) # silence Markdown library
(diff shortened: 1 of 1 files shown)


coltonlw:
This was wrong because config.log actually has no handlers, basicConfig adds the handlers to the root logger and we need our messages to propagate up.  I have seen "duplicate" messages on startup which I'm looking into

gsfr:
If you launch uWSGI with multiple processes and/or threads, you are basically launching multiple instances of our app, which will each emit their own log messages. Is that maybe what you are seeing?

coltonlw:
Thats a great thing to bring up, both are set to 1.  I have found the issue (bug setting __config_persisted to True) I'll open another PR shortly.

PR #437, opened 2016-08-27
Set additionalProperties to false in schema defs

coltonlw:
When I separated the user schema into a user definition I added additionalProperties in the wrong place so it was not taking effect.  Everywhere properties is specified, additionalProperties must be specified in the same document or it will default to True.  This also required creating a separate user-output definition for the output schemas to share, because the keys "created" and "modified" are not shared by input schemas.  There is no way to add properties when referencing a schema.  For an explanation of that topic, see http://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties

### DIFF
--- a/api/base.py
+++ b/api/base.py
@@ -270,7 +270,7 @@ class RequestHandler(webapp2.RequestHandler):
         if isinstance(exception, webapp2.HTTPException):
             code = exception.code
         elif isinstance(exception, validators.InputValidationException):
-            code = 400
+            code = 422
             log.warn(str(exception))
         elif isinstance(exception, APIConsistencyException):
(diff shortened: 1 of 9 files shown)


coltonlw:
Following @nagem advice I updated the output schema to use additionalProperties instead of duplicating content of user schema for output.  Also changed base.py error handling to 422 on invalid json, instead of 400.  Re-enabled abao test for "POST /users -> 422" which would prevent this regression in the future (broken json schema / not setting additionalProperties to false)

coltonlw:
Closing this until I confirm changing 400 to 422 on invalid json won't cause some type of regression in the UI

PR #441, opened 2016-08-29
Schema defs additional properties

coltonlw:
Fix regression where additionalProperties was not properly set on user input schema.

Enable abao hook for POST /users -> 400 to test for this regression

Replaces #437


nagem:
LGTM, changing the test to expect a 400 is a nice midstep that gets us coverage now.

PR #450, opened 2016-09-01
Fix file origin for packfile upload

### LINKED ISSUE #449: Set packfile origin from ticket info
Since `packfile-end` is an SSE endpoint, frontend code cannot send headers (including the auth header needed to set file origin). The ticket has enough info to set it, and can be used instead.

nagem:
Fixes #449

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -10,11 +10,12 @@ import zipfile
 from . import config
 from . import encoder
 from . import files
-from .jobs import rules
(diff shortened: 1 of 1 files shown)


nagem:
@dpuccetti I can't assign you to the review but this should fix the frontend origin bug.

nagem:
It does really need much of a review so I'll probably merge when CI is green.

PR #457, opened 2016-09-01
Fix coll GET from pylint refactor

### LINKED ISSUE #454: collections/:collectionId Response Body Empty
The response has a 200 code both on my local environment and dev, but the response body is empty.

nagem:
Fixes #454


nagem:
@dpuccetti there was a minor regression from the pylint refactoring, it's fixed now. I'll merge when CI is green.

Issue #410, opened 2016-08-17
Remove insecure mode 


coltonlw:
Right now we run our integration tests in "insecure" mode which allows any user to impersonate any other user and also to gain root privileges without actual authorization.  To properly verify permissions we will need to run in "secure" mode and use different test accounts for different roles.  My opinion is that insecure mode should be removed altogether, the same thing can be accomplished simply by adding a test user with root permissions, and that doesn't leave us open to the pitfall where a production instance somehow gets insecure mode turned on.

https://github.com/scitran/core/blob/master/api/base.py#L26

https://github.com/scitran/core/blob/master/api/base.py#L56

https://github.com/scitran/core/blob/master/test/docker-compose.yml#L29

coltonlw:
Also remove debug links

ryansanford:
@coltonlw Check with Imad before proceeding. He has expressed appreciation in the past with the debug links. See if there's a way to accommodate his use case in a post-insecure-mode world.

Issue #451, opened 2016-09-01
Traceback swallowed by error handling code


coltonlw:
dispatcher method in api.py does not log traceback when it catches an error that is not an instance of webapp2.HTTPException
https://github.com/scitran/core/blob/master/api/api.py#L200
Suggest removing this except block, causing the error to be handled by [WSGIApplication._internal_error](https://github.com/GoogleCloudPlatform/webapp2/blob/master/webapp2.py#L1582).  Since _internal_error does not log a traceback with the exception, we need to add an error handler for code 500 to log the traceback and then return webob.exc.HTTPInternalServerError which is the same return value as _internal_error.  This is better than adding code to log the traceback in the dispatcher function of our router, since any exceptions thrown outside of the router code will also be logged

coltonlw:
This should be coordinated with #199 so that the traceback is logged to the request specific logger adapter.  Then when we get a traceback and a 500, we'll have contextual request info to aid in debugging.

coltonlw:
fixed by #482

PR #468, opened 2016-09-06
Remove insecure mode

coltonlw:
This removes insecure mode

Fixes: #410


nagem:
LGTM.

api/api.py
```diff
@@ -197,18 +196,12 @@ def dispatcher(router, request, response):
             response.headers['Content-Type'] = 'application/json; charset=utf-8'
     except webapp2.HTTPException as e:
         util.send_json_http_exception(response, str(e), e.code)
-    except Exception as e: # pylint: disable=broad-except
-        if config.get_item('core', 'debug'):
-            message = traceback.format_exc()
-        else:
-            message = 'Internal Server Error'
-        util.send_json_http_exception(response, message, 500)
```
kofalt: Is this diff related to the purpose of the PR?

Unless I'm missing something I feel like catching the base class is a good idea here rather than letting it go up the stack.
I suggesting leaving the catch but remove the inner if statement.

kofalt:
This looks great :+1: lots of branches removed.

@ryansanford Can you confirm we don't need insecure & debug for _anything_ anymore? I don't use it...

api/api.py
coltonlw: @kofalt I added a ref to issue #451, which could be addressed as a separate change but since this except  branch was clearly added to facilitate debug mode I removed it.  I chose to let it go up the stack because without debug mode, the behavior would always be "log the error and send a message saying 'Internal Server Error'" to the client, which we can do just by letting the exception bubble up.  Also it seems to me having 500's return JSON is not a good idea because other times we may not catch an error and thus the body may or may not be JSON encoded, which could cause issues for the requester.  Status code 500 defined as "A generic error message, given when an unexpected condition was encountered and no more specific message is suitable"

coltonlw: I guess whether or not 500's should be JSON is debatable, from description of 5xx status codes:  "Except when responding to a HEAD request, the server SHOULD include an entity containing an explanation of the error situation"    I think in the case of 500 specifically JSON should not be included since this is the code used for an uncaught exception, compared to other more specific 5xx codes purposefully generated by the server.

kofalt: As long as there is there is defined behaviour for `Exception` bubbling up I'm fine with it :+1:

coltonlw: webapp2 doesn't define a default handler for any error codes, so it will call [WSGIApplication._internal_error](https://github.com/GoogleCloudPlatform/webapp2/blob/master/webapp2.py#L1582) which logs to the root logger with level error, however it only logs the exception and not a traceback.  I'm going to removed #451 from the list of issues this PR fixes, and update it to include an action item for adding a 500 handler to log the traceback.

gsfr:
It was my understanding that we were going to remove *insecure mode* and *debug links*, but not *debug mode*, which prints stack traces, IIRC.

I strongly suggest that we keep this PR limited to the first two items and discuss the last one separately.

Just adding a couple of comments on what's mentioned above:
- Given that we are open source, printing a stack trace in prod, doesn't reveal any secrets. Whether it's good form or not is debatable. It certainly makes debugging easier.
- I vote that the API never ever returns anything but JSON.

coltonlw:
@gsfr I agree leaving 500 errors JSON for the sake of consistency makes sense.  Right now we may have some 500's which are not json-encoded, #451 will address that in addition to logging the traceback.  Thus, I reverted my changes to that try-except block

I also restored debug mode, I know other frameworks have had security issues where debug mode has a remote code execution vulnerability and thus having a way to enable it in production carries some risk.  That's definitely not what this PR was about so I've reverted that change as well.

gsfr:
Per offline discussion, revert change in `api/api.py` and then we should be good to go.

coltonlw:
@gsfr @nagem Can I get a once over and final sign-off from you two before I merge this?

gsfr:
Thanks, @coltonlw. LGTM, but untested.

nagem:
I can test tomorrow but just looking it over it LGTM2.

PR #472, opened 2016-09-06
Fix bug passing parameters within userauth.py

nagem:


gsfr:
Are we good to go here? Looks like a trivial change to me.

rentzso:
LGTM

coltonlw:
@gsfr Megan's waiting on me to submit a test for this issue, specifically the abao test to validate the response of the list users endpoint against the raml output schema which is how we discovered the bug.  I plan to submit that shortly, before I leave today

coltonlw:
LGTM

Issue #473, opened 2016-09-07
Extend python jsonschema validator for additionalProperties to default to False


coltonlw:
In jsonschema additionalProperties defaults to true, see [section 5.4.4.3](http://json-schema.org/latest/json-schema-validation.html#anchor64) of the spec

This is rather unfortunate as by default input and output schemas for the api will allow arbitrary additional keys in json objects, unless we remember to set additionalProperties to false.  I recommend we extend the additionalProperties validator for python jsonschema to default additionalProperties to false.

https://python-jsonschema.readthedocs.io/en/latest/creating/#jsonschema.validators.extend

https://github.com/Julian/jsonschema/blob/master/jsonschema/_validators.py#L21

coltonlw:
jsonschema package sets the value of additionalProperties to an empty dictionary if it is not present, so one route would be to change the order of the if statements such that

```python
if not aP and extras:
```

comes first.  This branch will be hit when additionalProperties is an empty dict since those evaluate to False.

coltonlw:
Note:  Make sure to write a unit test that our change in default behavior is taking place

coltonlw:
This might not work because setting additionalProperties to false in the wrong place disallows all properties

PR #476, opened 2016-09-07
Fix user list output schema

coltonlw:
@nagem Diff is showing a bunch of stuff that got added in master, if you rebase your branch the only change is to add additionalProperties to user list output schema, which was the test that caught this originally when I ran it in a feature branch.  Also #473 was created to prevent further issues related to additionalProperties in the schemas.


coltonlw:
Closing while I fix whatever is failing the build

PR #478, opened 2016-09-07
Fix user list output schema

coltonlw:
This improves the user schemas to add proper validation of output when listing users or the /self endpoint


coltonlw:
@nagem The failing test on push actually confirms this is working correctly, failing before your fix, passing after.  Ready to merge

nagem:
LGTM

Issue #199, opened 2016-03-15
Use key to identify log entries from the same request.


ryansanford:
On a busy system, when an error occurs within scitran/core, it can difficult and time consuming to identify which log entries relate to the same request.

Suggest enhancement to create a key for each request, and include it in every log entry in a way that is easy to grep out, like `(request_key:abc123def456)`. Implementation should avoid request_key collisions between threads for a sane number of requests.

For reference, my uwsgi log file with 1986753 lines is 138423095 bytes in size.

nagem:
I'm thinking a couple of random alphanumerics mixed with a date-time stamp would keep us unique enough.

gsfr:
Wondering if webapp2 might already have something in store for us. Might be worth checking `self` and `self.request`.

kofalt:
I've seen this used before to great effect: https://github.com/newobj/nginx-x-rid-header

Works out of the box; just log the header :)

gsfr:
I poked at webapp2 requests for a sec while looking into #200. Doesn't look like webapp2 offers a unique request id out of the box. Would be easy enough to generate one using the uuid module and store in on the request (`self`).

@kofalt's proposal looks simple, but locks us into nginx, which I'd prefer to avoid. Sure, it looks like the obvious choice today, but we thought the same thing about uwsgi at some point. Now, we're less sure.

I don't understand the multi-component remark. What's the core/engine interaction for web requests?

kofalt:
Just thinking out loud. We currently route directly from the engine to the API, but we could put everything through nginx if we wanted. Food for thought.

The industry pattern is to attach these tracers at the outer layer, so I'd be happier to see them added there, at our rev-proxy. If we were taking bets, my money would be on nginx being the last currently-existing component to be switched out, plus it's trivial to generate guids in other rev-proxy products.

If we're planning to stick with New Relic in the longer term (my read on that situation is "no", due to the cost), it might be worth checking out [their plugin](https://www.nginx.com/blog/nginx-plugin-for-new-relic) to see if they can enable this feature. It wasn't clear from what I saw.

gsfr:
@ryansanford Are you comfortable compiling nginx from source, which the suggested module requires?

ryansanford:
My suggestion is to define and document a header that all scitran/core logging will use as a key, and provide UUID generation on entry, or first log event if that header is not defined. That allows 3rd parties the flexibility to generate and use this request_key outside the scitran/core stack if they wish.

[...]

PR #481, opened 2016-09-08
[Do Not Merge]  [Example]  Add per-request logging context

coltonlw:
This adds a custom request object which has additional attributes "id" and "logger".  logger is a LoggerAdapter which adds the request id to each message string.  In the future this can be expanded to include additional per-request context information with each logging call, including extra keys when we log in JSON.  Also added a thread local request context which can be accessed globally via set_current_request and get_current_request

Example output:
```
2016-09-08 11:09:46 scitran.api base.py  105:INFO Initialized request request_id=0935954b-1473350986
```

Fixes: #199, #451

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -21,6 +21,7 @@ from .handlers import collectionshandler
 from .handlers import searchhandler
 from .handlers import schemahandler
 from .handlers import reporthandler
+from .request import SciTranRequest, set_current_request
 
 log = config.log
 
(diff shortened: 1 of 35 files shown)


coltonlw:
FYI pushed a one-line change to remove unused request_id attribute

coltonlw:
Talking to Megan, I'm going to make a new feature branch which does not have global request context / set_current_request + get_current_request functionality.  Most of the logger calls that do not already have access to the request are of level debug and so not useful in production anyway and not worth the complexity of the global request context.  I'll leave this open incase anyone still wants to review the implementation for curiosity sake

coltonlw:
Closing because the new PR will be opened from a different branch

PR #482, opened 2016-09-08
Add logging context for request including request id

coltonlw:
This adds a custom request object which has additional attributes "id" and "logger". logger is a LoggerAdapter which adds the request id to each message string. In the future this can be expanded to include additional per-request context information with each logging call, including extra keys when we log in JSON.  Also added a uwsgi log var for request_id so we can use that in our uwsgi logformat allowing us to match a request_id from python log messages to a message in the uwsgi access log.  uwsgi log var is optional so other webservers may be used

Example output:

```
2016-09-08 13:34:04  scitran.api  base.py  104:INFO Initialized request request_id=c88e9c98-1473359644
127.0.0.1 - - [08/Sep/2016:13:34:04 -0500] "GET /api/rules?root=true HTTP/1.1" 200 112 "-" "-" request_id=c88e9c98-1473359644
```

Fixes: #199, #451

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -1,6 +1,7 @@
 import json
 import sys
 import traceback
+
 import webapp2
 import webapp2_extras.routes
 
@@ -21,8 +22,14 @@ from .handlers import collectionshandler
 from .handlers import searchhandler
 from .handlers import schemahandler
 from .handlers import reporthandler
+from .request import SciTranRequest
 
 log = config.log
+
+try:
+    import uwsgi
+except ImportError:
+    uwsgi = None
 
 class Config(base.RequestHandler):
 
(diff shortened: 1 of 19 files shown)


gsfr:
I'm good to go with this approach per offline discussion, but @nagem still needs to sign off on both approach and implementation.

api/api.py
```diff
@@ -21,9 +22,15 @@
 from .handlers import searchhandler
 from .handlers import schemahandler
 from .handlers import reporthandler
+from .request import SciTranRequest
 
 log = config.log
```
nagem: Looks like you forgot one here.

nagem:
I like it, looks good!

api/api.py
coltonlw: Apparently we do use this logger in the new relic setup code

PR #483, opened 2016-09-09
Allow analysis files from session to be used as input in other analyses

nagem:
Only supported for session level analyses. Gear-based analyses are also (currently) only allowed at the session level so this doesn't restrict actions any further than they already are.

How the analyses are created hasn't change, for an example (using an analysis level file):

```
POST /api/sessions/57d1f3cbd73ca10011ff0ee5/analyses?job=true HTTP/1.1

{
    "analysis": {
        "label": "Megan Analysis 1"
    },
    "job" : {
        "gear": "gear-name",
        "inputs": {
            "not-a-dicom": {
                "type": "analysis",
                "id": "57d1fb632ab3a500241cbe78",
                "name" : "8892_1_1_localizer_montage.zip"
            }
        },
        "tags": ["analysis"]
    }
}
```

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -5,6 +5,7 @@ from ..auth import INTEGER_ROLES
 
 CONT_TYPES = ['acquisition', 'analysis', 'collection', 'group', 'project', 'session']
 
+
 def getPerm(name):
     return INTEGER_ROLES[name]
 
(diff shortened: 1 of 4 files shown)


lmperry:
LGTM! Looking forward to working with this.

kofalt:
Remove `self._cache` as per offline discussion then LGTM :+1:

nagem:
Fixed up a few things, ready to land when CI hits green.

api/dao/containerutil.py
```diff
@@ -44,33 +45,36 @@ class ContainerReference(object):
...
         if not isinstance(type, basestring):
             raise Exception('Container type must be of type str')
         if not isinstance(id, basestring):
             raise Exception('Container id must be of type str')
 
-        self.type = type
-        self.id   = id
+        self.type   = type
+        self.id     = id
+        self._cache = None
```
kofalt: This last line should be removed

Issue #404, opened 2016-08-15
Review skipped abao tests


coltonlw:
We are skipping some abao tests for various reasons, this issue tracks those tests so we can begin making the changes necessary to enable them.

Skipped tests are listed here: https://github.com/scitran/core/blob/master/test/integration_tests/abao/abao_test_hooks.js

nagem:
According to [this comment](https://github.com/cybertk/abao/issues/149#issuecomment-226778697), we will not be able to send file form data using abao, or at least not yet. 

Unfortunately since we expect the upload calls to be of this type, we cannot test metadata only, either.

coltonlw:
The majority of the remaining skipped tests are for Jobs, which just need a workflow test for creating a project and session and uploading data, so those objects are available in the API to enable creating a test job.  I'm working on that in a feature branch so I'll steal this issue.

PR #485, opened 2016-09-10
Add raml and schemas for groups and rules.  Update raml for jobs and …

coltonlw:
Adds RAML and schemas for jobs, gears, groups.  Complete rules output raml and schemas.

Changes groups handler to use separate schemas for new and update, which $ref the group input schema definition.  Tests added for invalid input to groups endpoint to verify that change.

### Review Checklist

…gears


gsfr:
Great. Thanks, @coltonlw. I think @nagem should do the review for this.

coltonlw:
Sounds good - no rush Megan hopefully the github notifications don't disturb your weekend

To make it clear how the tests for /groups 400 (invalid input) get enabled, they're turned on "automatically" by abao when a status 400 response for that endpoint is added to the raml.  In this case I'm using a [RAML trait](http://raml.org/developers/raml-200-tutorial#traits) called "validates-json-body" which adds a 400 response to any endpoint with that trait.  The trait definition is in api.raml

raml/schemas/definitions/group-output.json
```diff
@@ -0,0 +1,46 @@
...
+          "_id": {
+            "type": "string"
+          },
+          "site": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "access",
+          "_id"
```
nagem: `roles` on groups and `permissions` on projects/sessions/acquisitions will all follow the same schema, they just have different field names (`roles` vs `permissions`). We'll probably want to add a schema snippet for that.

coltonlw:
@nagem got that taken care of

coltonlw:
Blocking #404 , we need this first because it corrects job output schema.  Otherwise I have changes ready to resolve #404 and enable jobs test plus a postman workflow test that includes uploading a file to an acquisition.

nagem:
This LGTM but there is a lot to go over. Even if we did miss something, it's safe to merge. I'd say go ahead.

PR #488, opened 2016-09-12
Gear configuration

kofalt:
Closes #380, #381, #382.

Allows for suggesting files for a job's input, and for validating job configuration.

### Review Checklist

- Tests were added to cover all code changes
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md

PR #494, opened 2016-09-14
Gear config followup - Add Analysis File Suggestion

nagem:
`api/gears/<gear-id>/suggest/session/<session-id>` will now also return analyses with suggestion blocks in the file objects. This is the union of #483 and #488.

If we only want to return analyses file suggestions for certain gear categories, that can be accommodated. Returning for all gears currently. @lmperry, would a there be a usecase for running non-analyses gears on analysis output?

### Review Checklist

### DIFF
--- a/api/dao/containerstorage.py
+++ b/api/dao/containerstorage.py
@@ -24,10 +24,9 @@ class ContainerStorage(object):
         self.dbc = config.db[cont_name]
 
     def get_container(self, _id, projection=None, get_children=False):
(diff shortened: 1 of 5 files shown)


nagem:
@kofalt and I discussed these changes offline earlier this week. Tested, merging now.

lmperry:
@nagem: There should never be a use case that would involve a Utility gear consuming the output of an analysis gear.

PR #496, opened 2016-09-14
Add gear submission tests

### LINKED ISSUE #492: Only allow valid gears to be submitted to POST /gears
Now that we're using the `gear_tools` library, we can validate that a gear manifest is actually valid :+1:

kofalt:
Closes #492.

### Review Checklist

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -74,7 +74,20 @@ class GearHandler(base.RequestHandler):
         if _id != doc.get('name', ''):
             self.abort(400, 'Name key must be present and match URL')
 
-        upsert_gear(self.request.json)
+        try:
+            upsert_gear(self.request.json)
+        except ValidationError as err:
+            key = None
+            if len(err.relative_path) > 0:
+                key = err.relative_path[0]
+
+            self.response.set_status(400)
+            return {
+                'reason': 'Gear manifest does not match schema',
+                'error': err.message.replace("u'", "'"),
+                'key': key
+            }
+
         return { 'name': _id }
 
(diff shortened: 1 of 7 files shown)


kofalt:
I don't have time to fix this test before I head out today, but everything should be basically ready.

Ping @ryansanford for acceptance as this affects the infrastructure.

coltonlw:
@kofalt There are two tests failing, the first is the one that adds the dcm-convert gear.  I opened postman and executed that request against my local instance and got the response: 

```json
{
  "reason": "Gear manifest does not match schema",
  "key": null,
  "error": "'author' is a required property"
}
```

The second test is failing because it depends on that gear being added

kofalt:
CI is green :tada:

api/jobs/handlers.py
```diff
@@ -74,7 +74,20 @@ def post(self, _id):
...
 
-        upsert_gear(self.request.json)
+        try:
+            upsert_gear(self.request.json)
+        except ValidationError as err:
+            key = None
+            if len(err.relative_path) > 0:
+                key = err.relative_path[0]
+
+            self.response.set_status(400)
```
nagem: You use a 422 for a schema validation error in jobs/add - we should keep it consistent.

nagem:
Other than the comment about the error code, LGTM.

api/jobs/handlers.py
kofalt: Colton requested this be 400 for now until we go over this module with a full-verification pass :)

kofalt:
Merging this based on our conversation

Issue #493, opened 2016-09-14
Improve speed / caching of run-tests-docker.sh


kofalt:
`run-tests-docker.sh` works great, but does a bunch of pip/npm/docker cycling on each run.

Since this is something to run iteratively, it would be nice to jump straight to the actual tests, using a clean & already-built set of containers if they've been built before.

coltonlw:
@ryansanford I'm going to investigate this see how much of the native testing can use the docker test script instead, nice work you did there.

ryansanford:
@kofalt I believe PR #500 resolved this issue. Can you confirm?

kofalt:
Yes, it seems to work pretty well now :)

PR #498, opened 2016-09-14
Use run-tests-docker.sh instead of run-tests-osx.sh

coltonlw:
Upgrades run-tests-docker.sh so it can be used in place of run-tests-osx.sh

- Updates Dockerfile to use install-ubuntu.sh instead of duplicating commands
- Install pip from the python packaging authority so it can be updated without conflicting against apt
- Updates run-tests-docker.sh to re-use image between runs.  When re-using the image, tests run in 25 seconds or with the new option to disable linting 10 seconds
- Removes run-tests-osx.sh  in favor of run-tests-docker.sh
- Unifies testing process between osx, ubuntu, and travis, allowing ubuntu users easier and much faster local test runs
- Travis uses run-tests-docker.sh, increasing runtime to about 4.5 minutes (from 3 min).  This is OK because we now have a unified, cross-platform way for developers to locally run the tests in less than 30 seconds using the same process as the CI
- Image travis builds can now be used as public scitran-core image, a future changeset can add auto-push to dockerhub and an example docker-compose.yml

Fixes: #493 
### Review Checklist


ryansanford:
@coltonlw 
The approach to resolve #493 is 2-fold
1. Making the build of the docker image optional (accomplished by the -b in run-tests-docker.sh  with this PR)
2. Copy only the pip and npm execution requirements, and execute them in the Dockerfile before copying the whole repo ala  https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/add-or-copy

Unrelated to the resolution of #493, Docker is considered optional infrastructure for scitran/core, and is an async concern to Travis CI. Docker issues should not impact the latency of travis CI, nor impact the resulting pass/fail badge.

coltonlw:
@ryansanford You're right that copying the whole repo in does not take full advantage of the performance benefits of docker cache.  IMO those performance benefits are of limited value because developers will usually not rebuild the image to run tests.  I thought it was more valuable to have the Dockerfile use install-ubuntu.sh so we are testing that script and making it clear what the install process is.  When the Dockerfile starts to duplicate commands from the install script the processes diverge and I want to avoid that unless we have good reasoning.

Can you explain the second part of your comment a little more, specifically the part relating to travis CI?  I think we should use the docker test process, now that it allows for quick iterations without rebuilding the image, because it's a cross-platform solution and prevents us from maintaining separate processes for native ubuntu and osx testing.  If we are using the docker test process, why wouldn't we want the CI to run tests using docker?  The application can still be installed and run on native ubuntu, and that's part of the reasoning for wanting to use the install-ubuntu.sh script in the Dockerfile.

coltonlw:
@gsfr Had a great offline discussion with Ryan, I understand why the changes to travis and the removal of the run-tests-osx.sh script are not acceptable changes for the project.  I'm going to re-work the PR to revert most of the changes, leaving the -b option for run-tests-docker.sh per Ryan's request.

gsfr:
Thanks, @ryansanford and @coltonlw. Sounds good to me.

PR #499, opened 2016-09-14
Check permissions for upload endpoints

### LINKED ISSUE #423: Extend the API to support label uploads from users
Label uploads are only available to drone requests.
However, the [organizer](https://github.com/scitran/organizer) should use that route to upload user data.
A possible solutions is to provide a route to retrieve a drone API key for normal users (related to #413 I guess).
@nagem @gsfr I would like to know your thoughts about this.
Thanks!

nagem:
Closes #423

Will also modify existing python integration tests to cover expected `403`s.
### Review Checklist

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -139,7 +139,7 @@ class UIDPlacer(Placer):
         metadata_validator(self.metadata, 'POST')
 
         # If not a superuser request, pass uid of user making the upload request
-        targets = self.create_hierarchy(self.metadata, self.context.get('uid'))
+        targets = self.create_hierarchy(self.metadata, user=self.context.get('uid'))
 
         self.metadata_for_file = {}
 
--- a/api/placer.py
+++ b/api/placer.py
@@ -138,7 +138,8 @@ class UIDPlacer(Placer):
         metadata_validator = validators.from_schema_path(payload_schema_uri)
         metadata_validator(self.metadata, 'POST')
 
-        targets = self.create_hierarchy(self.metadata)
+        # If not a superuser request, pass uid of user making the upload request
+        targets = self.create_hierarchy(self.metadata, self.context.get('uid'))
 
         self.metadata_for_file = {}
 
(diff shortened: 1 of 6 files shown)


nagem:
For match uploads, we will always check that the user has read-write access to the session.

For upsert endpoints, if the session exists we check perms at the session level, otherwise we require read-write access to the group. @gsfr do we want to handle a situation where the user has access to an existing project but not the group? I left it out since it would cause a larger change to the flow we have (project upserting), but it can be added.

gsfr:
@nagem Thanks!

Yes, we must respect the project permissions and we have to account for the possibility that the project has different permissions than the group.

api/placer.py
```diff
@@ -138,7 +138,8 @@ def check(self):
         metadata_validator = validators.from_schema_path(payload_schema_uri)
         metadata_validator(self.metadata, 'POST')
 
-        targets = self.create_hierarchy(self.metadata)
+        # If not a superuser request, pass uid of user making the upload request
```
rentzso: I think it would be possible to check here user permissions.
This would keep permissions checking separate from database access.

nagem: Possibly .. it's hard because before this point, we only know of the uid or label and `hierarchy.py` does a lot of work to look these item up by this information as well as fuzzy match on group/project. If we wait until after running `create_hierarchy`, the hierarchy has already been created/modified. 

We could maybe expose some of the lookup functions?

rentzso: Yeah, that's a good idea. 
You could add a "get_target_object" method to each hierarchy uploader (in a similar way to how create_hierarchy is defined for them), and then compute permissions on it.
It could be a little convoluted but you will extract some complexity from the hierarchy creation/update part.

coltonlw:
@nagem You can use postman instead of python to test this, when you have a chance I can show you how

rentzso:
@nagem just pushed a small fix to the label upload flow.

PR #506, opened 2016-09-19
Add test/bin/inject_api_key.py as replacement for bootstrap-dev.py

coltonlw:
Adds test/bin/inject_api_key.py which replaces ones users api key

Removes bootstrap-dev.py in favor of inject_api_key.py

This is very similar to bootstrap-dev.py except it limits the concern to setting the api key for one already existing user   
Also inject_api_key.py lives under test/bin to make it clear this is for testing purposes only

### Review Checklist


coltonlw:
@ryansanford per our discussion this morning about adding users through the API for permissions propagation, etc I am submitting this PR    An advanced user running the API on their laptop for testing could docker exec test/bin/inject_api_key.py  to set a saved api key, allowing the use case of having a static api key for testing saved in postman environments, etc.

ryansanford:
LGTM

PR #514, opened 2016-09-26
Fix nodejs install for install-dev-osx.sh

coltonlw:
Sending the npm tarball directly to tar and using --strip-components causes npm to not be properly installed.  This means the test dependencies are installed using the system npm, polluting the systems node_modules.  Not sure why tar doesn't work, possibly due to symlink.  Reverting to temporary directory method with this PR.

### Review Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,5 @@
 /runtime
 bootstrap.json
 .cache
-.coverage
+.coverage*
 coverage.xml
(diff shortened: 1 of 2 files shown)


gsfr:
That does not make sense to me. Isn't it just that `$VIRTUAL_ENV` is undefined and we don't have `set -u` to save us?

Maybe we should turn on `set -u` after activating the virtualenv?

coltonlw:
It doesn't make sense to me either.  We do activate the virtualenv prior to this, on line 56.  set -u afterwords still a good idea.

coltonlw:
I guess the only answer is I had an old runtime/ folder, I removed it and re-ran on master and it worked /facepalm

PR #518, opened 2016-09-28
Session templates

### LINKED ISSUE #495: Session templates
Allow projects to specify a "session template" - an expected format for the acquisitions and their files.

Plan for implementation:
 - A project can have a singular `session_template`, in the format of a json schema
 - A boolean flag on each session will be `true` if its acquisitions are in compliance with the template
 - When files are added or removed from an acquisition, the session's compliance with that template will be reviewed and the flag will be updated

nagem:
Closes #495.

Will be writing extensive tests, just wanted to get the review of the code going now while they are being written.

### Review Checklist


nagem:
@dpuccetti included are the RAML docs for the two new endpoints. Let me know if anything there isn't clear. We had originally talked about blocking the update to a project template if there were sessions without something like a `force=true` flag (to prevent unexpected long waits when session count is in the thousands). I'm going to hold off on that until we see any actual problems, it's been calculating extremely fast in all my tests.

api/handlers/containerhandler.py
```diff
@@ -480,6 +481,37 @@ def get_groups_with_project(self):
...
+        result = permchecker(self.storage.exec_op)('PUT', _id=project_id, payload=payload)
+
+        if result.modified_count == 1:
+            sessions = self.storage.get_children(project_id, projection={'_id':1})
+            log.debug('sessions are {}'.format(sessions))
+            session_storage = self.container_handler_configurations['sessions']['storage']
+            for s in sessions:
+                log.debug('update session {}'.format(s))
+                session_storage.exec_op('PUT', s['_id'], payload={'project_has_template': True})
+            return {'modified': result.modified_count}
```
kofalt: It makes sense why this logic has to be in the handler, but could you add a code comment explaining why?

kofalt:
This looks great :+1:

api/handlers/containerhandler.py
```diff
@@ -480,6 +481,37 @@ def get_groups_with_project(self):
...
+        else:
+            self.abort(404, 'Could not find project {}'.format(project_id))
```
coltonlw: Would APINotFoundException be better here?

api/handlers/containerhandler.py
```diff
@@ -480,6 +481,37 @@ def get_groups_with_project(self):
...
+
+    def calculate_project_compliance(self, **kwargs):
+        project_id = kwargs.pop('cid')
+        self.config = self.container_handler_configurations['projects']
+        self.storage = self.config['storage']
+        return {'sessions_changed': self.storage.recalc_sesssions_compliance(project_id)}
```
coltonlw: Does recalc_sessions_compliance return the number of sessions changed?

nagem: It returns the ids of the sessions changed. The idea is that a cron job could clean up any sessions that were modified and did not receive the correct compliance marker. Logging the cron job output (session ids changed) could give us a few clues as to where we missed an endpoint that modifies a session or acquisition. 

Oh I just saw now that I haven't committed the bug fix that adds the return statement. Thanks for pointing that out.

api/handlers/containerhandler.py
```diff
@@ -480,6 +481,37 @@ def get_groups_with_project(self):
...
+        validators.validate_data(template, 'project-template.json', 'input', 'POST')
+        payload = {'template': json.dumps(template)}
+        payload['modified'] = datetime.datetime.utcnow()
+
+        permchecker = self._get_permchecker(container)
+        result = permchecker(self.storage.exec_op)('PUT', _id=project_id, payload=payload)
+
+        if result.modified_count == 1:
+            sessions = self.storage.get_children(project_id, projection={'_id':1})
+            log.debug('sessions are {}'.format(sessions))
```
nagem: I'll probably just remove it 😄. I try to clean up any debug statements I add during development.

api/dao/liststorage.py
```diff
@@ -105,7 +106,11 @@ def _delete_el(self, _id, query_params):
         update = {'$pull': {self.list_name: query_params} }
         log.debug('query {}'.format(query))
         log.debug('update {}'.format(update))
-        return self.dbc.update_one(query, update)
+        result =  self.dbc.update_one(query, update)
+        if self.list_name is 'files' and self.cont_name in ['sessions', 'acquisitions']:
+            session_id = _id if self.cont_name == 'sessions' else AcquisitionStorage().get_container(_id).get('session')
```
nagem: Yeah, I have a habit of using a ternary-type if even when it's hard to read ...

PR #521, opened 2016-09-29
Add dockerhub trigger for master branch and tags

### LINKED ISSUE #520: Trigger dockerhub builds
Dockerhub builds should be triggered for
- pushes/merges to master branch
- tag pushes

ryansanford:
Resolves #520 

- [x] Any push to master triggers dockerhub build.
- [x] Any tag push causes trigger where dockerhub image tag matches tag in this repo.

### Review Checklist

### DIFF
--- a/docker/build-trigger.sh
+++ b/docker/build-trigger.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+# Triggers an auto-build on dockerhub for the given source control reference.
+#
+# Example usage: ./build-trigger Tag 1.0.0 https://registry.hub.docker.com/u/scitran/reaper/trigger/11111111-2222-3333-4444-abcdefabcdef/
+function Main() {(
+  set -e
+
+  if [ $# -ne 3 ] ; then
+    >&2 echo "Usage: $( basename $0 ) <source-control-ref-type> <source-control-ref-value> <trigger-url>"
+    exit 1
+  fi
+
+  export SOURCE_CONTROL_REF_TYPE="${1}"
+  export SOURCE_CONTROL_REF_NAME="${2}"
+  export TRIGGER_URL="${3}"
+
+
+  if [ -z "${SOURCE_CONTROL_REF_TYPE}" ] ; then
+    >&2 echo "INFO: Source control reference type provided, skipping build trigger."
+    exit 0
+  fi
+
+  if [ -z "${SOURCE_CONTROL_REF_NAME}" ] ; then
+    >&2 echo "INFO: Source control tag name not provided, skipping build trigger."
+    exit 0
+  fi
+
+
+  # Trigger builds if branch is master, or for any tag.
+  EXEC_BUILD_TRIGGER=false
+
+  if [ "${SOURCE_CONTROL_REF_TYPE}" == "Branch" ] && [ "${SOURCE_CONTROL_REF_NAME}" == "master" ] ; then
+    EXEC_BUILD_TRIGGER=true
+  fi
+
+  if ! $EXEC_BUILD_TRIGGER && [ "${SOURCE_CONTROL_REF_TYPE}" == "Tag" ] ; then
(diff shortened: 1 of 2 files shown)


ryansanford:
@gsfr This is ready for review.

docker/build-trigger.sh
```diff
@@ -0,0 +1,53 @@
...
+  # Trigger builds if branch is master, or for any tag.
+  EXEC_BUILD_TRIGGER=false
+
+  if [ "${SOURCE_CONTROL_REF_TYPE}" == "Branch" ] && [ "${SOURCE_CONTROL_REF_NAME}" == "master" ] ; then
+    EXEC_BUILD_TRIGGER=true
+  fi
+
+  if ! $EXEC_BUILD_TRIGGER && [ "${SOURCE_CONTROL_REF_TYPE}" == "Tag" ] ; then
+    EXEC_BUILD_TRIGGER=true
+  fi
```
gsfr: I would simplify this to
```
if [ "${SOURCE_CONTROL_REF_TYPE}" == "Tag" ] ; then
```

docker/build-trigger.sh
```diff
@@ -0,0 +1,53 @@
...
+    TriggerDockerBuild
+    >&2 echo "INFO: A dockerhub build for ${SOURCE_CONTROL_REF_TYPE} '${SOURCE_CONTROL_REF_NAME}' has been triggered."
+  fi
+
+)}
+
+function TriggerDockerBuild() {(
+  TRIGGER_PAYLOAD="{\"source_type\": \"${SOURCE_CONTROL_REF_TYPE}\", \"source_name\": \"${SOURCE_CONTROL_REF_NAME}\"}"
+  curl -H "Content-Type: application/json" --data "${TRIGGER_PAYLOAD}" -X POST "${TRIGGER_URL}"
+)}
```
gsfr: I don't see the benefit of using two functions in this script that communicate via "locally-defined" global variables. Linear code would be much simpler.

docker/build-trigger.sh
```diff
@@ -0,0 +1,53 @@
...
+    exit 1
+  fi
+
+  export SOURCE_CONTROL_REF_TYPE="${1}"
+  export SOURCE_CONTROL_REF_NAME="${2}"
+  export TRIGGER_URL="${3}"
+
+
+  if [ -z "${SOURCE_CONTROL_REF_TYPE}" ] ; then
+    >&2 echo "INFO: Source control reference type provided, skipping build trigger."
```
gsfr: Shouldn't this say "Source control reference type NOT provided"? If so, then this is also wrong in scitran/reaper.

docker/build-trigger.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+# Triggers an auto-build on dockerhub for the given source control reference.
+#
+# Example usage: ./build-trigger Tag 1.0.0 https://registry.hub.docker.com/u/scitran/reaper/trigger/11111111-2222-3333-4444-abcdefabcdef/
+function Main() {(
```
gsfr: What's the purpose of the extra parenthesis around the function body? Is that to force a subshell? Which you then export your local variables out of?

docker/build-trigger.sh
ryansanford: Correct! Thanks.

docker/build-trigger.sh
ryansanford: Sure, I prefer this readability improvement.

docker/build-trigger.sh
ryansanford: Agreed.

When I started tweaking the script, I split out `TriggerDockerBuild()` and `Main()` because that was more clear to me while I was experimenting with the structure. After landing on the solution, I didn't take a critical eye to realize the abstractions were no longer carrying their own weight.

ryansanford:
@gsfr This is ready for review again.

PR #530, opened 2016-10-07
Add stats endpoint for colls/projs

nagem:
`api/projects` and `api/collections`, when called with the query param `stats=true` will add 3 new fields to each object in the response:
`session_count`
`subject_count`
`attachment_count`

### Review Checklist

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -37,6 +37,32 @@ def add_id_to_subject(subject, pid):
         subject['_id'] = bson.ObjectId()
     return subject
 
+def get_stats(cont, cont_type):
+    """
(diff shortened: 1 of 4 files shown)


api/dao/containerutil.py
```diff
@@ -37,6 +37,32 @@ def add_id_to_subject(subject, pid):
         subject['_id'] = bson.ObjectId()
     return subject
 
+def get_stats(cont, cont_type):
```
nagem: I would like to move this to `containerstorage.py`, as well as `add_id_to_subject()` above, but `session-templates` contains the rework. 

This will most likely merge first so I'll make the move in the rebase of `session-templates` when this lands.

PR #545, opened 2016-10-12
use site on permissions for label and uid uploads

rentzso:


nagem:
@rentzso and I just had an offline discussion. The `site` portion of user permissions is not entirely supported throughout the app. There are a few instances of user permissions lookup where `site` is assumed to be `local`, as upload endpoints were before this PR. 

Cross-site user auth is not currently supported, but to remain consistent until its fate has been decided, the user's site should be checked. I'll leave it to @rentzso to decide if we want to merge this PR, but the work itself LGTM.

I will open a ticket to remove all `site` logic from user permissions.

Issue #538, opened 2016-10-11
URL rule for /groups/{GroupId}/tags/{TagValue} should specify allowed methods


coltonlw:
https://github.com/scitran/core/blob/master/api/api.py#L155

I believe the allowed methods should be ["GET", "PUT", "DELETE"], notice POST is not included

coltonlw:
I think the same is true for the /{ContainerName}/{ContainerId}/tags/{TagValue} route:

https://github.com/scitran/core/blob/master/api/api.py#L158

PR #550, opened 2016-10-14
Fix api route method descrepencies

### LINKED ISSUE #535: URL rule for /api/groups/GroupId/roles should only allow POST
This URL rule only works for POST request, due to #534 , and it seems to be this route was only intended to allow POST method.  Suggested resolution is to specify "methods" argument when adding the url rule for this route

https://github.com/scitran/core/blob/master/api/api.py#L150

nagem:
Fixes #535, #538 

### Review Checklist


coltonlw:
The tests for these endpoints are in my latest RAML feature branch, otherwise LGTM

PR #564, opened 2016-10-20
Fix packfile subject code bug

### LINKED ISSUE #557: Sessions created via packfile upload always create new Subject ID's instead of matching
When new session is created via packfile upload, if subject has same code as existing subject in project, it gets a new subject ID

nagem:
Fixes #557 

### Review Checklist


nagem:
We might get a sad coveralls output since the database.py call structure is tested but not the individual update scripts.

PR #568, opened 2016-10-24
Add some more file types.

cgc:
This is probably not quite the right way to do this, but how can I get some new file types supported? I guess "support" is a bit ill-defined and may include 1. rendering a particular way in the UI (in simple descriptions, and also with specialized viewers) 2. permitting querying by file type. thanks!

And if we don't think the above matter, then what is the right type of blob data types like this?


api/filetypes.json
```diff
@@ -4,8 +4,11 @@
     "dicom":        [ ".dcm", ".dcm.zip", ".dicom.zip" ],
     "parrec":       [ ".parrec.zip", ".par-rec.zip" ],
     "gephysio":     [ ".gephysio.zip" ],
+    "log":          [ ".log" ],
```
nagem: It looks like the format here is to keep research/neuro file types up top, and more general file types below. Mind moving log down to the other list? Other than that, LGTM!

cgc: good call, just moved it. thanks @nagem

nagem:
Thanks! Merging..

Issue #334, opened 2016-06-06
Turn gears into first-class citizens


kofalt:
Break away from their singleton doc & turn into their own objects.

kofalt:
Rescoping this to just gears - it's pretty obvious that rules need an overhaul to the extent that we'll likely replace the current system entirely. It'd be a waste to move rules around when they're clearly going away :)

PR #572, opened 2016-10-26
Moved gears to their own table

kofalt:
Closes scitran/core#334


nagem:
Thanks for the changes, LGTM!

PR #581, opened 2016-10-27
Add multi-file-gear rule matching

kofalt:
This small change to gear rules allows a gear to have multiple files automatically selected by type. To use this feature, simply add a `match` key to the rule mapping gear input name to file type:

```json
"match": {
	"bval": "bval",
	"bvec": "bvec",
	"diffusion": "nifti"
}
```

### DIFF
--- a/api/jobs/rules.py
+++ b/api/jobs/rules.py
@@ -131,6 +131,12 @@ def queue_job_legacy(algorithm_id, input_):
     job = Job(algorithm_id, inputs)
     return job.insert()
 
+def find_type_in_container(container, type_):
+    for c_file in container['files']:
+        if type_ == c_file['type']:
+            return c_file
+    return None
+
 def create_jobs(db, container, container_type, file_):
     """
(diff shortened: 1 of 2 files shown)


gsfr:
Very cool. I'll leave it to @nagem to review the code. I'd like to see this merged asap, coordinated around SfN freeze.

kofalt:
Ryan, Megan and I have coordinated on the SfN freeze & have decided that non-SfN branches are clear to land on master. We'll branch off a release tag if we have more SfN core changes.

For review purposes: it should be noted that the same old behavior is left alone; this only does something new when the `match` key is present on a rule.

kofalt:
Meta: this may have been better implemented as using the gear manifest's own input specifications. That might be a good upgrade in the future, but this will serve for the high-priority use case.

nagem:
Code LGTM. We should open a ticket to amend the existing issue of TokenPlacer not calling the code that would create jobs.

Issue #578, opened 2016-10-27
Jobs endpoint 500's for a session with no acquisitions


coltonlw:
Impact: User interface error popups occur any time a session with no acquisitions is viewed

<img width="1279" alt="screen shot 2016-10-27 at 12 38 48 pm" src="https://cloud.githubusercontent.com/assets/20908947/19778431/596da522-9c42-11e6-8d1e-b9657efd5ff1.png">

TB:

```
2016-10-27 16:37:09      scitran.api                  base.py   288:ERRO Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
    return method(*args, **kwargs)
  File "./api/handlers/containerhandler.py", line 242, in get_jobs
    jobs = Queue.search(cont_array, states=states, tags=tags)
  File "./api/jobs/queue.py", line 197, in search
    type1 = containers[0].type
IndexError: list index out of range
 request_id=e45ec17c-1477586229

[pid: 19|app: 0|req: 106/417] 192.168.3.180 () {44 vars in 856 bytes} [Thu Oct 27 16:37:09 2016] GET /api/sessions/5812205ddad42900117f915e/jobs?root=true => generated 58 bytes in 14 msecs (HTTP/1.1 500) 3 headers in 130 bytes (2 switches on core 0)
```

This happens when a request is made to GET /api/sessions/{SessionId}/jobs  for a session which does not have any acquisitions.  

The error occurs when the search method is called with an empty container list (because there are no acquisitions aka containers for that session)
https://github.com/scitran/core/blob/bd0f068cfadbb38cce0ec3566513fd76d7b8ad8b/api/jobs/queue.py#L197

Suggest search method automatically return an empty list if it is called with an empty "containers" argument.

coltonlw:
@ryansanford can you prioritize this, could possibly impact demos?

nagem:
@coltonlw @ryansanford this is fixed in `add-job-info` as part of #552

nagem:
For context, @dpuccetti found this bug as we are allowing session creation in the UI for the first time. Currently there is no way in Flywheel to create a session without an acquisition (although the API supports it). It should not exist as an error in live sites, unless a user is POSTing directly to the API.

PR #582, opened 2016-10-27
Add jobs info

### LINKED ISSUE #552: Job endpoint enhancements for dashboard
1. [x] Store `origin` on job documents. Part of the job class; loaded during `/jobs/add` or `POST .../analyses/`.
2. [x]  Origin can be an optional `None` param on init; set to "system" or similar when not present. This is for jobs added by eg the rules settings.
3. [x] Add a `saved` string array to job object; this holds filenames of produced files. Set on engine & analysis uploads endpoints.
4. [ ] Save session information on analysis container references on the job.

cc @dpuccetti for frontend :thumbsup:

nagem:
Closes #552, #578.
### Review Checklist

### DIFF
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -88,7 +88,7 @@ class Job(object):
         self.config          = config_
         self.now             = now
         self.origin          = origin
-        self.outputs         = []
+        self.saved_files     = []
 
     @classmethod
     def load(cls, e):
@@ -176,7 +176,6 @@ class Job(object):
     def save(self):
         self.modified = datetime.datetime.utcnow()
         update = self.mongo()
-        config.log.debug(update)
         job_id = update.pop('id')
         result = config.db.jobs.replace_one({'_id': job_id}, update)
         if result.modified_count != 1:
--- a/api/jobs/jobs.py
+++ b/api/jobs/jobs.py
@@ -6,13 +6,14 @@ import bson
 import copy
 import datetime
 
+from ..types import Origin
 from ..dao.containerutil import create_filereference_from_dictionary, create_containerreference_from_dictionary, create_containerreference_from_filereference
 
 from .. import config
 
 
 class Job(object):
-    def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previous_job_id=None, created=None, modified=None, state='pending', request=None, id_=None, config_=None, now=False):
+    def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previous_job_id=None, created=None, modified=None, state='pending', request=None, id_=None, config_=None, now=False, origin=None):
         """
         Creates a job.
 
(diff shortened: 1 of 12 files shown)


nagem:
The `/jobs` endpoint doesn't return jobs run on analyses, only acquisitions. Let me know if you want that to change @dpuccetti.

api/jobs/jobs.py
```diff
@@ -162,6 +173,16 @@ def insert(self):
         result = config.db.jobs.insert_one(self.mongo())
         return result.inserted_id
 
+    def save(self):
+        self.modified = datetime.datetime.utcnow()
+        update = self.mongo()
+        config.log.debug(update)
```
kofalt: Do we want this log statement in there when we merge?

api/jobs/jobs.py
```diff
@@ -162,6 +173,16 @@ def insert(self):
...
+        job_id = update.pop('id')
```
kofalt: Is the id pop actually necessary?

api/jobs/jobs.py
```diff
@@ -162,6 +173,16 @@ def insert(self):
...
+        result = config.db.jobs.replace_one({'_id': job_id}, update)
+        if result.modified_count != 1:
+            raise Exception('Job modification not saved')
+        return {'modified_count': 1}
```
kofalt: Why return a manual map rather than the actual result?

api/types.py
```diff
@@ -5,5 +5,6 @@
     'user':    'user',    # An authenticated user
     'device':  'device',  # A connected device (reaper, script, etc)
     'job':     'job',     # Made on behalf of a job (downloading data, uploading results, etc)
-    'unknown': 'unknown', # Other or public
+    'system':  'system',  # Created by the system (a job initiated by a rule, etc)
+    'unknown': 'unknown'  # Other or public
```
kofalt: @dpuccetti Be aware that when this PR lands there will be a new origin type.

bin/database.py
```diff
@@ -506,6 +507,20 @@ def upgrade_to_18():
...
+
+    Add origin information to job object
+    """
+
+    update = {
+        '$set': {
+            'origin' : {'type': str(Origin.unknown), 'id': None}
+        }
+    }
+    config.db.jobs.update_many({'origin.': {'$exists': False}}, update)
```
kofalt: Do you need specifically `origin.`, or will `origin` work here?

api/jobs/jobs.py
```diff
@@ -78,6 +87,8 @@ def __init__(self, name, inputs, destination=None, tags=None, attempt=1, previou
         self.id_             = id_
         self.config          = config_
         self.now             = now
+        self.origin          = origin
+        self.outputs         = []
```
kofalt: I hate to do this to you, but I think `outputs` should be renamed before this lands - `inputs` and `outputs` (currently a singleton `destination` but not forever) are kinda reserved meaning. I'd take `saved` or `result-filenames` instead?

bin/database.py
nagem: Good catch!

nagem: But, surprisingly, `origin.` does work as intended.

api/jobs/jobs.py
nagem: It's an existing practice across the API. Not married to the idea or anything, but might as well keep it consistent for now.

api/jobs/jobs.py
nagem: Nope, thanks.

nagem:
Made changes requested by @kofalt, merging.

PR #585, opened 2016-11-01
Basic gear exchange format support

kofalt:
For compatibility with [flywheel/exchange](https://github.com/flywheel-io/exchange)


nagem:
Sorry this is late - LGTM. Only change I would make is explicitly skipping those endpoints in the aboa hooks rather than losing the documentation we have. We can fix it up when gears stabilize, I know there are a lot of changes coming.

PR #589, opened 2016-11-04
disable resolver cache

coltonlw:
### Review Checklist


ryansanford:
@coltonlw Please set the uwsgi threads back to `2` for this PR. Then it will LGTM.

Assigning to Megan for her review as well.

ryansanford:
My assumption is that coverage went down because the changed function is covered, and its lines decreased. If that's true, I don't see a reason for that coverage drop to hold up this PR.

nagem:
LGTM, ignoring the thread discussion. It's nice to have the option open to add threads, if a need arises. 

Regarding an offline discussion - this does add 10-15ms (if I remember correctly) to api calls that use json schemas. One option discussed was we could resolve the schemas and hold them in memory before the processes are forked.

PR #593, opened 2016-11-11
Server-side resolver

kofalt:
A first pass at the server-side resolver. The resolver can walk through a human-meaningful path and return the path walked through the data hierarchy, as well as valid children of the final node. This is great for statelessly exploring the hierarchy without incuring high latency / transfer overhead.

A heavily-abbreviated example for `scitran/Neuroscience`:

```json
{
	"children": [
		{
			"node_type": "session", 
			"label": "patient_1"
		}, 
		{
			"node_type": "session", 
			"label": "control_1"
		},
		{
			"node_type": "file",
			"name": "qa_report.png"
		}, 
		{
			"node_type": "file",
			"name": "spectrum.csv"
		}
	], 
	"path": [
		{
			"node_type": "group",
			"_id": "scitran"
		}, 
		{
			"node_type": "project",
			"label": "Neuroscience"
		}
	]
}
```

The path is ordered by the hierarchy; children are always consistently ordered.
Will abort if any part of the path is not authorized; will filter any unauthorized children.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -21,6 +21,7 @@ from .handlers import userhandler
 from .handlers import grouphandler
 from .handlers import containerhandler
 from .handlers import collectionshandler
+from .handlers import resolvehandler
 from .handlers import searchhandler
 from .handlers import schemahandler
 from .handlers import reporthandler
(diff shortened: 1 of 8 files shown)


nagem:
LGTM as a first pass - nice work.

Just want to make note of a situation that might get us in a bind - it is possible to be in a project's permissions list without having permission to the group. Discussion unrelated to this ticket is currently ongoing on how to handle this situation smoothly, any final decision there should eventually be reflected in this permissions checking workflow.

nagem:
Also, mind writing a test or two 😎?

PR #599, opened 2016-11-21
Add basic support for jwt LDAP auth

nagem:
Allows users to authenticate with a JWT token distributed by the ldap-jwt web service. Support for multiple auth types will be added as an enhancement when a plan to continue supporting environment variable config can be created that works with an array of auth config options. 

### Review Checklist


kofalt:
LGTM. Might want to have a gander at the LDAP library docs to make sure that `r.ok` is a sufficient check against all possible error returns.

nagem:
@dpuccetti you can authenticate the same as always - just pass the jwt in the `Authorization` header instead of the Google token.

ryansanford:
@nagem Can you add the new env variable to `sample.config` file?

nagem:
@kofalt The source code of the tiny node service we're using is pretty simple, I don't see any way we'd get bit by that `r.ok` logic. Even if it did I doubt the return would have a valid email.

nagem:
@ryansanford added.

ryansanford:
@nagem Dev and Prod environment changes to default `SCITRAN_AUTH_AUTH_TYPE` correctly are complete.

Suggest setting default in `api/config.py` to align with the related existing defaults.

After that, LGTM.

api/config.py
```diff
@@ -42,6 +42,7 @@
         'prefetch': False
     },
     'auth': {
+        'auth_type': 'google',
```
nagem: @ryansanford let me know if you meant something other than this in your latest comment.

PR #621, opened 2016-12-20
Batch gears

### LINKED ISSUE #604: Batch gear processing endpoint
`POST /batch` with a set of containers, in the form of:

```json
{
	"gear": "example-gear",

	"targets": [
		{
			"type": "session",
			"id": 3
		},
		{
			"type": "session",
			"id": 4
		}
	]	
}
```

At the present time, this endpoint will probably error out for any `type` other than `session`.

Returns:

```json
{
	"ticket": 3,

	"gear": "example-gear",

	"job-count": 20,

	"matched": [
		{
			"type": "acquisition",
			"id": 1
		}
	],

	"ambiguous": [
		{
			"type": "acquisition",
			"id": 5
		}
	],

	"unmatched": [
		{
			"type": "acquisition",
			"id": 2
		}
	]
}
```

Matched containers represent what containers jobs will run on, ambiguous containers have an ambiguous match where multiple files could satisfy one or more of the inputs, unmatched containers had no valid candidates for one or more inputs.

Under the hood, we will store all the job objects that would be created, the origin of the request, state of batch (pending / launched), on the batch object. Example:

```json

{
	"_id": 1, 

	"state": "pending",

	"origin": {
		"type": "user",
		"id": "example@example.example"
	},

	"proposal": [
		{
			"dicom": {
				"type": "acquisition",
				"id": 2,
				"name": "bob.dcm"
			},
			"bval": {
				"type": "acquisition",
				"id": 2,
				"name": "bob.bval"
			}
		},
		{
			"dicom": {
				"type": "acquisition",
				"id": 4,
				"name": "bob.dcm"
			},
			"bval": {
				"type": "acquisition",
				"id": 4,
				"name": "bob.bval"
			}
		}
	]
}
```

`POST /batch/<id>/run` will enqueue all the jobs on the batch object, return an array of job IDs, and change the batch object's state to launched.

The batch object will gain a `job-ids` string array.

`POST /batch/<id>/cancel` will set any created jobs with state `pending` to `failed`.

`GET /batch/<id>` will return the batch object with additional generated key `status` with form:

```json
{
	"status": {
		"total": 100,
		"pending": 30,
		"running": 30,
		"failed": 30,
		"complete": 10
	}
}
```

`GET /batch/<id>/jobs` (optional) will return an array of the job objects

`POST /batch/<id>/nuke` (future update) will delete the batch object and all data it generated (jobs and files).

nagem:
Closes #604.

### Review Checklist

### DIFF
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -2,10 +2,13 @@
 Batch
 """
 import bson
+import datetime
 
 from .. import config
 from ..dao import APINotFoundException
+from ..dao.containerstorage import AcquisitionStorage
 from ..dao.containerutil import create_filereference_from_dictionary, create_containerreference_from_filereference
+from ..dao.liststorage import AnalysesStorage
 from .jobs import Job
 from .queue import Queue
 from . import gears
@@ -100,6 +103,10 @@ def insert(batch_proposal):
     """
     Simple database insert given a batch proposal.
     """
+
+    time_now = datetime.datetime.utcnow()
+    batch_proposal['created'] = time_now
+    batch_proposal['modified'] = time_now
     return config.db.batch.insert(batch_proposal)
 
 def update(batch_id, payload):
@@ -107,8 +114,10 @@ def update(batch_id, payload):
     Updates a batch job, being mindful of state flow.
     """
 
+    time_now = datetime.datetime.utcnow()
     bid = bson.ObjectId(batch_id)
     query = {'_id': bid}
+    payload['modified'] = time_now
     if payload.get('state'):
         # Require that the batch job has the previous state
         query['state'] = BATCH_JOB_TRANSITIONS[payload.get('state')]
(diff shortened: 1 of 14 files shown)


api/jobs/batch.py
```diff
@@ -0,0 +1,182 @@
...
+    gear_name = batch_job.get('gear')
+    config_ = batch_job.get('config')
+    origin = batch_job.get('origin')
+
+    jobs = []
+    job_ids = []
+    for inputs in proposed_inputs:
+        for input_name, fr in inputs.iteritems():
+            inputs[input_name] = create_filereference_from_dictionary(fr)
+        # TODO support analysis gears (will have to create analyses here)
```
nagem: We'll need to create analyses here on the fly. I've got a plan for this, will add that functionality tomorrow.

api/jobs/batch.py
```diff
@@ -0,0 +1,182 @@
...
+            continue
+
+    update(batch_job['_id'], {'state': 'cancelled'})
+    return cancelled_jobs
+
+
+def get_stats():
+    """
+    Return the number of jobs by state.
+    """
```
nagem: I'm not sure if this is useful to you @dpuccetti but I'll add it before merging. It's a simple mongo aggregation.

api/jobs/batch.py
```diff
@@ -0,0 +1,182 @@
...
+    raise NotImplementedError()
+
+def resume():
```
nagem: `resume` and `delete` are prioritize for a next pass. They will not be implemented before merging.

api/jobs/handlers.py
```diff
@@ -271,3 +256,128 @@ def retry(self, _id):
...
+
+        container_ids = []
+        container_type = None
+
+        for t in targets:
+            if not container_type:
+                container_type = t.get('type')
+            else:
+                # Ensure all targets are of same type, may change in future
+                if container_type != t.get('type'):
```
nagem: Currently assumes the target files will be at this level (ie, these containers are acquisitions not sessions). Will add target resolver for other hierarchy levels.

nagem:
There is one "gotcha" with this design - if a job fails, we no longer have it's "current" job id in the array stored on the batch job. I have a few ideas for this, @kofalt maybe we can discuss later today.

api/util.py
```diff
@@ -171,6 +171,12 @@ def send_json_http_exception(response, message, code):
     response.headers['Content-Type'] = 'application/json; charset=utf-8'
     response.write(content)
 
+def send_custom_json_http_exception(response, errors, code):
+    response.set_status(code)
+    content = json.dumps({errors})
+    response.headers['Content-Type'] = 'application/json; charset=utf-8'
+    response.write(content)
```
kofalt: Any way we could constrain this to forcibly overlap with the above struct, `send_json_http_exception`?

I ask because this would add a lot of complexity to consuming the API from a static language. Right now, I can deserialize API errors into a struct and get useful things out. This change would mean errors are just an arbitrary json object :pensive: 

How about adding `custom=None` to the existing function, and I can consider that a wild west map?

```go
type ApiError struct {
	Message    string `json:"message"`
	StatusCode int    `json:"status_code"`
	
	// I would add this
	Custom map[string]interface `json:"custom"`
}
```

api/jobs/handlers.py
```diff
@@ -271,3 +256,128 @@ def retry(self, _id):
...
+            'state': 'pending',
+            'origin': self.origin,
+            'proposed_inputs': [c.pop('inputs') for c in matched]
+        }
+
+        batch.insert(batch_proposal)
+        batch_proposal.pop('proposed_inputs')
+        batch_proposal['not_matched'] = results['not_matched'],
+        batch_proposal['ambiguous'] = results['ambiguous'],
+        batch_proposal['matched'] = matched
```
nagem: I remember when it's separate - these are not stored on the batch object in mongo, only returned so the caller can display/do something intelligent with the containers that match/didn't match/were ambiguous. The batch job doesn't need to know this information.

nagem:
There is an issue with the aboa tests generated from the raml - Docker fails with an unhelpful error. I'll add those as a follow up PR as to not hold up these changes.

PR #625, opened 2016-12-29
Improve and organize the routes table!

kofalt:
Something we've been needing for awhile, I think :slightly_smiling_face: 

Don't bother with the diff, the [result](https://github.com/scitran/core/blob/better-routes/api/api.py) is what you're after.

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -106,8 +106,8 @@ endpoints = [
 
         # Users
 
-        route( '/users',                   UserHandler, 'get_all', m=['GET']),
-        route( '/users',                   UserHandler,            m=['POST']),
+        route( '/users',                   UserHandler, h='get_all', m=['GET']),
+        route( '/users',                   UserHandler,              m=['POST']),
         prefix('/users', [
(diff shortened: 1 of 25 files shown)


kofalt:
> coverage decreased (-0.05%)

api/api.py
```diff
@@ -7,210 +7,266 @@
...
+        # Search
+
+        route('/search',                     SearchHandler, h='advanced_search', m=['POST']),
+        route('/search/files',               SearchHandler, h='get_datatree',    m=['GET']),
+        route('/search/<cont_name:{cname}>', SearchHandler,                      m=['GET']),
+
+
+        # Users
+
+        route( '/users',                   UserHandler, 'get_all', m=['GET']),
```
nagem: Just missing the `h` here, otherwise looks good! Because python it will work without it, though 😉.

nagem:
Merge it if you see it today, otherwise I'll merge it tonight so we can bug bash it tomorrow.

api/api.py
kofalt: Fixed lol

Issue #639, opened 2017-01-16
Provide build info with docker images


ryansanford:
Docker images provided by scitran should be able to report basic build information, like github commit hash. Potentially make this available via `api/version` endpoint.

PR #640, opened 2017-01-16
Load build info into db with dockerized version

ryansanford:
Example resolution of #639 

@nagem Let's discuss this approach of loading build info into the DB vs. doing something where the code serves this same info directly. The disadvantage to persisting this in the database, is that it can be possible for it to be out of sync with the code serving it.

### Review Checklist

### DIFF
--- a/docker/build-info-load.py
+++ b/docker/build-info-load.py
@@ -1,0 +1,12 @@
+"""
+Loads the build info into the version document in the database.
+"""
+import json
+import os
+import pymongo
+db_uri = os.getenv('SCITRAN_PERSISTENT_DB_URI')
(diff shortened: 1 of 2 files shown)


kofalt:
Couple comments: 

1. All other things being equal I'd rather the API determine its own build info for this purpose (`git rev-parse HEAD`) rather than read off a file produced by something else.
1. Failure (not having `version.json`, or in my case not having `git`) should not block the API. Maybe it could try json, then git, then register an object with dummy values.
1. This adds another hard assumption that there is one effective API node. I don't know how many of those we really want to rack up.

ryansanford:
Abandoning this approach and POC. Closing...

PR #644, opened 2017-01-19
Add Access Logging

### LINKED ISSUE #397: Access logging
In our effort to becoming HIPPA compliant, we have to start logging user access to data.

In a first iteration, let's implement logging of the following "view" events:
- [ ] bulk file download
- [ ] individual file download
- [ ] file viewing
- [ ] retrieval of subject information

To keep us from having to log each session that is returned by session list queries, we should consider stripping non-essential information from those session lists, including the subject information.

I'm inclined to start with logging to a rotating log file, not the DB.

Open to hearing further thoughts and suggestions.

nagem:
Closes #397 

### Review Checklist


ryansanford:
@nagem 
Requests should fail when logging to the file fails. You can reproduce this issue by using the `log-remap` branch of fly/fly after mounting the attached osx image.
[logs.dmg.zip](https://github.com/scitran/core/files/724306/logs.dmg.zip)



```
uwsgi_1           | 2017-01-23 16:16:15      scitran.api           liststorage.py   125:DEBU projection {'files.$': 1}
uwsgi_1           | Traceback (most recent call last):
uwsgi_1           |   File "/usr/lib/python2.7/logging/__init__.py", line 875, in emit
uwsgi_1           |     self.flush()
uwsgi_1           |   File "/usr/lib/python2.7/logging/__init__.py", line 835, in flush
uwsgi_1           |     self.stream.flush()
uwsgi_1           | IOError: [Errno 28] No space left on device
uwsgi_1           | Logged from file base.py, line 371
uwsgi_1           | 2017-01-23 16:16:15   scitran.access                  base.py   371:INFO {"access_type": "download_file", "context": {"acquisition": {"id": "58823ab85fc0430011009e4c", "label": "8403_1_1_localizer"}, "group": {"id": "scitran", "label": "Scientific Transparency"}, "project": {"id": "58823aa25fc043001a009e3f", "label": "Neuroscience"}, "session": {"id": "58823ab85fc0430011009e4b", "label": "patient_1"}}, "origin": {"id": null, "type": "unknown"}, "request_method": "GET", "request_path": "/api/acquisitions/58823ab85fc0430011009e4c/files/8403_1_1_localizer.dicom_3Plane_Loc_SSFSE_20141113162516_1b.nii.gz", "timestamp": "2017-01-23T16:16:15.639316+00:00"}
uwsgi_1           | [pid: 17|app: 0|req: 31/131] 172.19.0.1 () {44 vars in 1081 bytes} [Mon Jan 23 16:16:15 2017] GET /api/acquisitions/58823ab85fc0430011009e4c/files/8403_1_1_localizer.dicom_3Plane_Loc_SSFSE_20141113162516_1b.nii.gz?ticket=147dc803-2951-4c04-a4dc-dbcf3dade8e4 => generated 2621839 bytes in 467 msecs (HTTP/1.1 200) 4 headers in 221 bytes (9265 switches on core 1)


```

nagem:
Did the request succeed? In my testing if there was an uncaught error in the logging the request would return a 500. The download endpoint does function differently than the others, though.

ryansanford:
@nagem 
The `view_container` and `download_file` requests did succeed, yes. I didn't test all access types, assuming they would behave the same.

ryansanford:
@nagem 

With the latest iteration, a `delete_file` still results in the file being deleted from the container, even though a 500 response is returned.

downloading files and viewing containers seem to be behaving correctly in these situations.

nagem:
@ryansanford that's expected, that fix is still in a local branch on mine. I'll commit it up now.

api/api.py
```diff
@@ -193,6 +196,12 @@ def prefix(path, routes):
 
         # Sessions
 
+        prefix('/sessions', [
+            route('/<cid:{cid}>/jobs',          ContainerHandler, h='get_jobs',     m=['GET']),
+            route('/<cid:{cid}>/subject',       ContainerHandler, h='get_subject',  m=['GET']),
+        ]),
+
+        route('/sessions/<cid:{cid}>/jobs',    ContainerHandler, h='get_jobs', m=['GET']),
```
kofalt: Are these new routes entirely? They don't feel like it, but I can't see where they moved from.

api/config.py
```diff
@@ -96,6 +98,15 @@ def apply_env_variables(config):
...
 log.debug(str(db))
 
+log_db = pymongo.MongoClient(
+    __config['persistent']['db_log_uri'],
+    j=True, # Requests only return once write has hit the DB journal
+    connectTimeoutMS=__config['persistent']['db_connect_timeout'],
+    serverSelectionTimeoutMS=__config['persistent']['db_server_selection_timeout'],
+    connect=False, # Connect on first operation to avoid multi-threading related errors
+).get_default_database()
+log.debug(str(log_db))
```
kofalt: Want this log statement, or just here for dev?

api/config.py
```diff
@@ -212,6 +223,8 @@ def initialize_db():
         upsert=True
     )
 
+    log_db.access_log.create_index('origin.id')
```
kofalt: Thoughts on creating an index on the timestamp as well? Otherwise most log-grep operations will be very inefficient.

I had a look through any concerns about using mongo as a time series, and mostly TLDR'd out, but there didn't seem to be anything big. Most common facet is that folks suggest combining multiple events into one document, say one per minute, and so reduce per-second logging by a factor of 60. 

:face_with_rolling_eyes: <-- emoji Github declines to render because they are fascists

So yeah. Index?

api/dao/hierarchy.py
```diff
@@ -67,12 +67,58 @@ def add_file(self, fileinfo):
 # TODO: already in code elsewhere? Location?
 def get_container(cont_name, _id):
     cont_name += 's'
-    _id = bson.ObjectId(_id)
+    if cont_name != 'groups':
+        _id = bson.ObjectId(_id)
```
kofalt: I'm grumpy about this but I can't reasonably ask for anything else rn :pensive:

api/dao/hierarchy.py
```diff
@@ -67,12 +67,58 @@ def add_file(self, fileinfo):
...
 
     return config.db[cont_name].find_one({
         '_id': _id,
     })
 
+def get_parent_tree(cont_name, _id):
+    """
+    Give a contanier and an id, returns that container and its parent tree.
```
kofalt: s/Give/Given

api/handlers/containerhandler.py
```diff
@@ -64,7 +65,7 @@ class ContainerHandler(base.RequestHandler):
             'parent_storage': containerstorage.ProjectStorage(),
             'storage_schema_file': 'session.json',
             'payload_schema_file': 'session.json',
-            'list_projection': {'info': 0, 'analyses': 0},
+            'list_projection': {'info': 0, 'analyses': 0, 'subject.firstname': 0, 'subject.lastname': 0},
```
kofalt: Are you sure this is the right layer to exclude personal data? I guess if it's a policy change on all container endpoints, it makes sense, it just surprised me.

api/handlers/listhandler.py
```diff
@@ -474,6 +484,22 @@ def post(self, cont_name, list_name, **kwargs):
...
+        self.log_user_access(AccessType.delete_file, cont_name=cont_name, cont_id=_id)
+        try:
+            result = keycheck(storage.exec_op)('DELETE', _id, query_params=kwargs)
+        except APIStorageException as e:
+            self.abort(400, e.message)
+        if result.modified_count == 1:
+            return {'modified': result.modified_count}
+        else:
+            self.abort(404, 'Element not removed from list {} in container {} {}'.format(storage.list_name, storage.cont_name, _id))
+
```
kofalt: Is this part of the same branch because it was handy to do at the same time? Did it move from somewhere else?

api/handlers/listhandler.py
```diff
@@ -782,6 +808,23 @@ def download(self, cont_name, list_name, **kwargs):
...
```
kofalt: Same question here

api/web/base.py
```diff
@@ -208,6 +210,35 @@ def validate_oauth_token(self, access_token, timestamp):
...
+        Return succcess boolean if user successfully authenticates.
+
+        Used for access logging.
+        Not required to use system as logged in user.
+        """
+
+        if not self.uid:
+            self.abort(400, 'Only users may log in.')
+
+        return {'success': True}
```
kofalt: Presumed to be a stub rn :slightly_smiling_face:

api/web/base.py
```diff
@@ -310,6 +341,39 @@ def handle_exception(self, exception, debug):
 
         util.send_json_http_exception(self.response, str(exception), code, custom=custom_errors)
 
+    def log_user_access(self, access_type, cont_name=None, cont_id=None):
```
kofalt: I'm worried about the logging parameter taking optional arguments. For the `access_type` that require it, should this function throw if these opt args are missing?

api/web/base.py
```diff
@@ -310,6 +341,39 @@ def handle_exception(self, exception, debug):
...
+
+        log_map = {
+            'access_type':      access_type.value,
+            'request_method':   self.request.method,
+            'request_path':     self.request.path,
+            'origin':           self.origin,
+            'timestamp':        datetime.datetime.utcnow()
+        }
+
+        if access_type not in [AccessType.user_login, AccessType.user_logout]:
```
kofalt: Before this line, do a type check on `access_type` to confirm that it's truly an enum (else throw).

api/web/request.py
```diff
@@ -29,3 +41,30 @@ def get_request_logger(request_id):
...
+            cont_id = None
+
+            if access_type not in [AccessType.user_login, AccessType.user_logout]:
+
+                cont_name = kwargs.get(cont_kwarg)
+                cont_id = kwargs.get(cont_id_kwarg)
+
+                # Only log view_container events when the container is a session
+                if access_type is AccessType.view_container and cont_name not in ['sessions', 'session']:
+                    return result
```
kofalt: Why does this decorator appear to do nothing for non-sessions? If that's a policy decision, could you add a doc line on why?

bin/database.py
```diff
@@ -635,6 +635,21 @@ def dm_v2_updates(cont_list, cont_name):
...
+    scitran/core PR #397
+
+    Add access logging
+    """
+    config.db.singletons.find_one_and_update(
+        {'_id': 'config', 'persistent.db_log_uri': {'$exists': False}},
+        {'$set': {
+            'persistent.db_log_uri': config.get_item('persistent', 'db_log_uri'),
+            'core.access_log_enabled': config.get_item('core', 'access_log_enabled')
+        }})
```
kofalt: Dumb question - this should have the affect of inserting default values in the database, correct?

Does @ryansanford prefer that method? Will you have pre-populated values before the upgrade runs?

test/bin/run-integration-tests.sh
```diff
@@ -48,6 +48,7 @@ PYTHONPATH="$( pwd )" \
     python test/bin/inject_api_key.py admin@user.com \
     "XZpXI40Uk85eozjQkU1zHJ6yZHpix+j0mo1TMeGZ4dPzIqVPVGPmyfeK"
 
+PYTHONPATH="$( pwd )" \
```
kofalt: What was breaking without this line?

api/api.py
nagem: `/subject` is, `/jobs` was moved here.

kofalt:
Trying out the new review flow again. Not a huge fan of how it shows review as negative, but interested to see how it feels when we're done.

api/config.py
nagem: I was following what we had above for the other MongoClient connection. It does help in dev to ensure that the connections were created properly.

api/config.py
nagem: I actually talked to Ryan about not indexing anything but the `_id` as Mongo will most likely only be a temporary holder for the log messages.

api/handlers/containerhandler.py
nagem: It's only excluding on the `get_all` handler endpoint. I might be confused about what surprised you - did you expect it to be explicit in the container storage layer?

api/web/base.py
nagem: Yeah it's just there to log. If (when) we move to proper session tokens this will be fleshed out.

api/web/request.py
nagem: It is a policy decision, I can add a comment.

test/bin/run-integration-tests.sh
nagem: We couldn't import the `api` package in the integ tests. I needed to in order to test that the log messages were logged properly.

nagem:
@ryansanford I think `true` is the correct call, especially because there would be no way to do an integration test of the access logging if the functionality was off. Also, looking at the code, the check for whether logging is turned on or off has been lost. Replacing it now.

test/bin/run-integration-tests.sh
ryansanford: @nagem One option is to handle the log db uri the same way as the scitran db uri. Not rely on `config` but rather a separate var that is passed in ala https://github.com/scitran/core/blob/master/test/integration_tests/python/conftest.py#L30

I'd prefer the 2 db uris are handled consistently.

nagem: Nice catch Ryan I didn't see we were doing that. I can use that instead.

ryansanford: @nagem Sweet. Once that's in I can hook up the parameter passing from `run-tests-ubuntu.sh` unless you want to tackle that as well with your change.

api/handlers/listhandler.py
nagem: It's just a logging version of the base delete method here: https://github.com/scitran/core/pull/644/files/c7c042b41b7c19977e02f3da32016107604066fd#diff-2b6de8dffa100181ea2c9b85e6e5f50cR166

`listhandler.py` has been relatively untouched for over a year, it might be time to see how we can optimize it after our growth over the last year.

api/web/base.py
nagem: Although we'll fail when it gets to access_type.value :)

nagem: Not in a reasonable or helpful way, mind you.

ryansanford:
General comment... I know this PR will be landing later than we all hoped, and some of that delay is on me for the latency with my part of the review. I'm really happy with the current direction, and believe the delay will be more than worth it. 🥂

api/config.py
kofalt: Are you certain adding more indexes is a bad idea? If mongo will truly be temporary then it shouldn't be a problem. OTOH, if ya'll have already decided this then let me know :)

api/handlers/containerhandler.py
kofalt: Um, maybe? Dropping arbitrary-length info & analyses fields make sense to me here, adding subject fname / lastname seems more like a policy decision that doesn't jive at this layer.

But, you're a lot more native to this area than I am - wanna just document why those keys are dropped & go ahead with it being there?

api/web/base.py
kofalt: Ah. So when this lands, clients will be expected to call the login endpoint... as a courtesy, but it won't do anything beyond log the event?

nagem: Exactly. I mean it will also give them a way to check that their method of auth works without having to make a dummy call to /sessions or something.

nagem: Depending on policy this could be changed to log whenever a user has an auth_token (eventually session token) inserted into the db but for now it's just a 1:1 for a presumed button on a frontend client.

kofalt:
Hmm.

So, reviews cannot accept individual threads, which is how I've seen it in every other major review product. So it's not immediately obvious which parts are resolved and which are outstanding - I can only accept / dismiss the whole thing.

In lieu of that, I propose that:

1. The submitter will comment on every thread - either to answer a question, or to post updates
2. The reviewer will :+1: that comment if they agree no further action is required.

And we use that to check off items. Once everything is :+1: ed, the reviewer can hit approve.

We don't have to be strict about it, but this is an example of a large review that we'll definitely see again. Thoughts?

nagem:
Thanks for adding those changes, @ryansanford. Excited to see it go green 👍

test/bin/run-integration-tests.sh
ryansanford: FYI -- We pulled out the access log URI, but this line is still required for validating the access type. I think that's fine.

ryansanford:
LGTM. Nice work. @nagem

nagem:
@ryansanford Do we have any final opinions on what kind of indexes we want on this data? Currently I have an index on `origin.id` which is the user's id but we could add one on `timestamp` or remove it entirely.

ryansanford:
@nagem My opinion is that index choices are out of scope for this PR, as scitran/core currently doesn't implement any reads. It's up to system implementers define their own indexes for whatever queries they want to support.

In other words, just the default _id index for this PR.

bin/database.py
nagem: After discussion this was out of scope. I'm going to remove the id index for now.

api/config.py
nagem: After discussion this was out of scope. I'm going to remove the id index for now.

bin/database.py
nagem: Discussed offline, but our logic in config.py is limited and could definitely be improve. It uses python's dictionary `.update()`, which is naive when dealing with sublevel maps. Any new default sublevel keys will get overwritten by what's in the database. I'll add a comment here and in `config.py` for clarity.

api/web/base.py
nagem: Addressed.

api/dao/hierarchy.py
nagem: Addressed.

api/web/request.py
nagem: Comment added.

test/bin/run-integration-tests.sh
nagem: This has been changed.

api/web/base.py
kofalt: Looks like this check was added

kofalt:
LGTM!

PR #648, opened 2017-02-01
Only log once per ticket for download

nagem:
### Review Checklist

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -214,6 +214,9 @@ def initialize_db():
     db.gears.create_index('name')
     db.batch.create_index('jobs')
 
+    if __config['core']['access_log_enabled']:
+        log_db.access_log.create_index('ticket_id')
+
     # Must be kept in sync with jobs/rules.py
     db.singletons.update({"_id" : "rules"}, {
             '$setOnInsert': {
--- a/api/config.py
+++ b/api/config.py
@@ -215,7 +215,7 @@ def initialize_db():
     db.batch.create_index('jobs')
 
     if __config['core']['access_log_enabled']:
-        log_db.access_log.create_index('ticket_id')
+        log_db.access_log.create_index('context.ticket_id')
 
     # Must be kept in sync with jobs/rules.py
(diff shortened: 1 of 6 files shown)


nagem:
This is ready to review and merge.

ryansanford:
@nagem 

I'm getting 6 `download_file` events for 30 downloads from a single ticket. Think it's due to multiple request connections on the requester end?

nagem:
@ryansanford Probably, there is a decent chunk of logic between when we get the ticket from the db and when we log. I could do a recheck of whats in the db right before attempting to log but unfortunately there is not an easy way to make sure it's perfect.

nagem:
I thought of another option - I could set the ticket id on the log message and then use $setOnInsert to make it reasonably atomic. We'd need an index on ticket_id to make it reasonably fast.

ryansanford:
@nagem I'm still getting 6 logging events with the same test. If you want to try and reproduce, it's subject 468, and I'm viewing the dicoms for acquisition 468_7_1_t1 in the testdata dataset.

nagem:
@ryansanford this has been modified to ensure only one log per ticket id. The downside being there must be an index for ticket id.

kofalt:
Code in its current state gets a :+1: at a high speed pass :slightly_smiling_face:

ryansanford:
Current commit behaves great. Only 1 entry per ticket now.

`ticket_id` at the root level feels odd to me, given that it's only valid for downloads. Objections to moving it under the `context` object @nagem ?

Otherwise, LGTM

nagem:
Thanks for reviewing! I'll move it to context. One additional change I'm going to make per my discussion yesterday with mgmt is adding `subject.id` and `subject.code` to the context, since most searching will be done from the context of "Who looked at this subject's data?" and it would be better if we didn't have to tease out the sessions that matter from a separate search.

ryansanford:
Functionality looks good to me from with the latest commit. Feel free to merge after CI failures are addressed.

Issue #607, opened 2016-12-02
User sometimes gets unusual 500 error after initial request when "logging in" on frontend


nagem:
This is not easy to reproduce, and is confusing as the only db call to the `authtokens` collection is a `replace_one` call with `upsert=true`.
```
Traceback (most recent call last):
  File "./api/api.py", line 217, in dispatcher
    rv = router.default_dispatcher(request, response)
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1278, in default_dispatcher
    return route.handler_adapter(request, response)
  File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 1101, in __call__
    handler = self.handler(request, response)
  File "./api/handlers/containerhandler.py", line 82, in __init__
    super(ContainerHandler, self).__init__(request, response)
  File "./api/base.py", line 52, in __init__
    self.uid = self.authenticate_user(access_token)
  File "./api/base.py", line 141, in authenticate_user
    config.db.authtokens.replace_one({'_id': access_token}, update, upsert=True)
  File "/usr/local/lib/python2.7/dist-packages/pymongo/collection.py", line 785, in replace_one
    bypass_doc_val=bypass_document_validation)
  File "/usr/local/lib/python2.7/dist-packages/pymongo/collection.py", line 710, in _update
    _check_write_command_response([(0, result)])
  File "/usr/local/lib/python2.7/dist-packages/pymongo/helpers.py", line 300, in _check_write_command_response
    raise DuplicateKeyError(error.get("errmsg"), 11000, error)
DuplicateKeyError: E11000 duplicate key error collection: scitran.authtokens index: _id_ dup key: { : "<google auth token>" }
```

nagem:
Fixed in #652

PR #652, opened 2017-02-03
Support multiple auth providers

### LINKED ISSUE #650: Support multiple auth providers

nagem:
Closes #650, #607 
### Review Checklist

### DIFF
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -97,12 +97,12 @@ class WechatOAuthProvider(AuthProvider):
 
     def validate_code(code):
         payload = {
-            'client_id':        self.config['client_id']
-            'client_secret':    self.config['client_secret']
-            'code':             code,
-            'grant_type':       'authorization_code'
+            'appid':        self.config['client_id']
+            'secret':       self.config['client_secret']
+            'code':         code,
+            'grant_type':   'authorization_code'
         }
-        r = requests.post(self.config['token_url'], data=payload)
+        r = requests.post(self.config['token_url'], params=payload)
         if not r.ok:
             raise APIAuthProviderException('User code not valid')
 
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -35,7 +35,7 @@ class AuthProvider(object):
         else:
             raise NotImplementedError('Auth type {} is not supported'.format(auth_type))
 
-    def validate_code(self, code):
+    def validate_code(self, code, **kwargs):
         raise NotImplementedError
 
     def ensure_user_exists(self, uid):
@@ -60,7 +60,7 @@ class JWTAuthProvider(AuthProvider):
     def __init__(self):
         super(JWTAuthProvider,self).__init__('ldap')
 
-    def validate_code(self, code):
+    def validate_code(self, code, **kwargs):
         uid = self.validate_user(code)
         return code, None, uid
 
@@ -88,7 +88,7 @@ class GoogleOAuthProvider(AuthProvider):
     def __init__(self):
         super(GoogleOAuthProvider,self).__init__('google')
 
-    def validate_code(self, code):
+    def validate_code(self, code, **kwargs):
         payload = {
             'client_id':        self.config['client_id'],
             'client_secret':    self.config['client_secret'],
@@ -170,7 +170,7 @@ class WechatOAuthProvider(AuthProvider):
     def __init__(self):
         super(WechatOAuthProvider,self).__init__('wechat')
 
-    def validate_code(self, code, registration_code=None):
+    def validate_code(self, code, **kwargs):
         payload = {
             'appid':        self.config['client_id'],
             'secret':       self.config['client_secret'],
@@ -185,6 +185,7 @@ class WechatOAuthProvider(AuthProvider):
         config.log.debug(response)
         openid = response['openid']
 
+        registration_code = kwargs.get('registration_code')
         uid = self.ensure_user_exists(openid, registration_code=registration_code)
 
         return {
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -62,7 +62,12 @@ class JWTAuthProvider(AuthProvider):
 
     def validate_code(self, code, **kwargs):
         uid = self.validate_user(code)
-        return code, None, uid
+        return {
+            'access_token': code,
+            'uid': uid,
+            'auth_type': self.auth_type,
+            'expires': datetime.datetime.utcnow() + datetime.timedelta(hours=1)
+        }
 
     def validate_user(self, token):
         r = requests.post(self.config['verify_endpoint'], data={'token': token})
@@ -75,12 +80,6 @@ class JWTAuthProvider(AuthProvider):
         self.ensure_user_exists(uid)
         self.set_user_gravatar(uid, uid)
 
-        return {
-            'access_token': token,
-            'uid': uid,
-            'auth_type': self.auth_type,
-            'expires': datetime.datetime.utcnow() + datetime.timedelta(hours=1)
-        }
 
 
 class GoogleOAuthProvider(AuthProvider):
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -1,0 +1,114 @@
+import requests
+from . import APIAuthProviderException, APIUnknownUserException
+from .. import config, util
+
+
+log = config.log
+
+
+AuthProviders = util.Enum('AuthProviders', {
+    'google'    : GoogleAuthProvider,
+    'ldap'      : JWTAuthProvider,
+    'wechat'    : WechatAuthProvider
+})
+
+class AuthProvider(object):
+    """
+    This class provides access to mongodb collection elements (called containers).
+    It is used by ContainerHandler istances for get, create, update and delete operations on containers.
+    Examples: projects, sessions, acquisitions and collections
+    """
+
+    def __init__(self, auth_type):
+        try:
+            self.config =  config.get_auth(auth_type)
+        except KeyError:
+            raise NotImplementedError('Auth type {} is not supported by this instance'.format(auth_type))
+
+    @staticmethod
+    def factory(auth_type):
+        """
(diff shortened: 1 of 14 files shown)


ryansanford:
@nagem 
Can you provide an example of the config file contents when specifying multiple auth providers? Thanks.

nagem:
```
{
    'google': {
            'client_id': '1052740023071-n20pk8h5uepdua3r8971pc6jrf25lvee.apps.googleusercontent.com',
            'id_endpoint': 'https://www.googleapis.com/plus/v1/people/me/openIdConnect',
            'auth_endpoint': 'https://accounts.google.com/o/oauth2/auth',
            'verify_endpoint': 'https://www.googleapis.com/oauth2/v1/tokeninfo'
    },
    'ldap': {
            'auth_endpoint': '<ldap_endpoint>',
            'verify_endpoint': '<ldap_endpoint>'
    }
}
```

nagem:
@ryansanford I didn't have an example of the ldap config available but it would look like above, with each auth type having it's own key.

ryansanford:
@nagem 
Hitting below error now with google login. Looks like `SCITRAN_SITE_REDIRECT_URL` env var isn't making it into my db.

```
uwsgi_1           | 2017-03-08 16:33:35      scitran.api                  base.py   386:DEBU from None None POST /api/login {} request_id=e3562fc5-1488990815
uwsgi_1           | 2017-03-08 16:33:35      scitran.api                  base.py   173:DEBU {u'auth_type': u'google', u'code': u'4/PEto-mRdnRKM-llE01lcsu5rS9GzEjilND3kBNVlcbI'}
uwsgi_1           | 2017-03-08 16:33:35      scitran.api                  base.py   316:ERRO Traceback (most recent call last):
uwsgi_1           |   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
uwsgi_1           |     return method(*args, **kwargs)
uwsgi_1           |   File "./api/web/request.py", line 52, in log_user_access_from_request
uwsgi_1           |     result = handler_method(self, *args, **kwargs)
uwsgi_1           |   File "./api/web/base.py", line 184, in log_in
uwsgi_1           |     token_entry = auth_provider.validate_code(payload['code'], registration_code=registration_code)
uwsgi_1           |   File "./api/auth/authproviders.py", line 97, in validate_code
uwsgi_1           |     'redirect_uri':     config.get_item('site', 'redirect_url')
uwsgi_1           |   File "./api/config.py", line 291, in get_item
uwsgi_1           |     return get_config()[outer][inner]
uwsgi_1           | KeyError: 'redirect_url'
uwsgi_1           |  request_id=e3562fc5-1488990815
```

ryansanford:
@nagem 

Ldap login results in this error.
```
uwsgi_1           | 2017-03-08 16:52:04      scitran.api                  base.py   386:DEBU from None None POST /api/login {} request_id=ee5d0c34-1488991924
uwsgi_1           | 2017-03-08 16:52:04      scitran.api                  base.py   173:DEBU {u'auth_type': u'ldap', u'code': u'redacted'}
uwsgi_1           | 2017-03-08 16:52:04      scitran.api                  base.py   316:ERRO Traceback (most recent call last):
uwsgi_1           |   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
uwsgi_1           |     return method(*args, **kwargs)
uwsgi_1           |   File "./api/web/request.py", line 52, in log_user_access_from_request
uwsgi_1           |     result = handler_method(self, *args, **kwargs)
uwsgi_1           |   File "./api/web/base.py", line 193, in log_in
uwsgi_1           |     token_entry['_id'] = session_token
uwsgi_1           | TypeError: 'tuple' object does not support item assignment
uwsgi_1           |  request_id=ee5d0c34-1488991924
uwsgi_1           | [pid: 21|app: 0|req: 8/22] 172.19.0.1 () {52 vars in 918 bytes} [Wed Mar  8 16:52:04 2017] POST /api/login => generated 82 bytes in 383 msecs (HTTP/1.1 500) 3 headers in 130 bytes (2 switches on core 0)
```

nagem:
I can fix the second bug no problem. The first one probably needs some config files updated to set that new key on config. The redirect_url is just the main site url.

dpuccetti:
@nagem We were hoping to fit one more feature into this PR.  It would be great to have an endpoint that cleared a wechat user's open_id and issued a new registration_code.  This would allow the user to tie their account to a new wechat account if necessary.

api/config.py
```diff
@@ -61,7 +66,20 @@
 
 def apply_env_variables(config):
     # Overwrite default config values with SCITRAN env variables if available
+
+    # Load auth config from file if available
+    if 'SCITRAN_AUTH_CONFIG_FILE' in os.environ:
+        auth_config = config['auth']
+        file_path = os.environ['SCITRAN_AUTH_CONFIG_FILE']
+        with open(file_path) as config_file:
```
kofalt: Is this feature strictly relevant to the PR, or tacked on as a convenience?

api/config.py
```diff
@@ -262,11 +280,14 @@ def get_config():
     return __config
 
 def get_public_config():
+    auth = copy.deepcopy(__config.get('auth'))
+    for values in auth.itervalues():
+        values.pop('client_secret', None)
```
kofalt: This is blacklisting when the below return whitelists. Thoughts on making this section a whitelist also for safety?

api/dao/dbutil.py
```diff
@@ -0,0 +1,31 @@
...
+from .. import config
+from . import APIStorageException
+
+
+
+
+def fault_tolerant_replace_one(coll_name, query, update, upsert=False):
+    """
+    Mongo does not see replace w/ upsert as an atomic action:
+    https://jira.mongodb.org/browse/SERVER-14322
```
kofalt: :tada:

api/handlers/userhandler.py
```diff
@@ -23,7 +23,12 @@ def __init__(self, request=None, response=None):
     def get(self, _id):
         user = self._get_user(_id)
         permchecker = userauth.default(self, user)
-        result = permchecker(self.storage.exec_op)('GET', _id, projection={'api_key': 0} or None)
+        projection = {'api_key': 0}
+        if self.user_is_admin:
+            projection['wechat.openid'] = 0
+        else:
+            projection['wechat'] = 0
```
kofalt: Dumb question: will this cause `wechat` objects to be returned for users that don't have a `wechat` field?

api/auth/authproviders.py
```diff
@@ -0,0 +1,246 @@
...
+        if email and uid:
+            gravatar = util.resolve_gravatar(email)
+            if gravatar is not None:
+                timestamp = datetime.datetime.utcnow()
+                # Update the user's gravatar if it has changed.
+                config.db.users.update_one({'_id': uid, 'avatars.gravatar': {'$ne': gravatar}}, {'$set':{'avatars.gravatar': gravatar, 'modified': timestamp}})
+
+
+
+class JWTAuthProvider(AuthProvider):
```
kofalt: Will this provider be in use when using google auth? Exchanging google token for a JWT?

kofalt:
Preliminarily LGTM. Some comments & questions.

api/config.py
nagem: Relevant to the PR. How we load in environment variable config options now works because they are all known key-values we can iterate over. Adding dynamic keys ('google', 'ldap', etc), we can't use the same logic. Using a file to load the config at startup time is the easiest way to support that.

api/config.py
nagem: It's a blacklist inside of a white list, L I S T C E P T I O N.

The white list for fields inside an auth config would be a lot longer than the blacklist, and is not standard for each type of config. I figured this was more clear, but I see your point.

api/handlers/userhandler.py
nagem: Nope. I don't like that we have to explicitly call out wechat in the userhandler methods but it was the simplest way to solve the current problem. A more robust solution (if we support more auth types that need registration) is having a more generic `registration` map.

api/auth/authproviders.py
nagem: It's for LDAP. We use a third party service that shuttles username/password to an LDAP server and creates a JWT if successful. I was considering that other authtypes that don't follow OAuth2 could use the same mechanism so the API doesn't have to deal with username/password flows. In that case they would also use this JWTAuthType provider.

api/config.py
kofalt: Yeah, up to you.
Personally, I would rather a whitelist be long and incomplete than a blacklist be incomplete.

api/handlers/userhandler.py
kofalt: Okay. I was going to propose a "private fields" map where we could stash that somewhere not these methods, but ¯\\_(ツ)_/¯

api/auth/authproviders.py
kofalt: Ah, yes.

api/config.py
kofalt: Given that, what are your thoughts on `update` below this diff? Do we want this file to be completely authoritative, at which point you can't modify the the config from the DB? Or, leave it as is? Are you comfortable using `update`?

ryansanford:
@nagem 

I'm getting this error when logging in via ldap. Using the latest commit on this branch.
```
uwsgi_1           | 2017-03-13 17:52:59      scitran.api                  base.py   320:ERRO Traceback (most recent call last):
uwsgi_1           |   File "/usr/local/lib/python2.7/dist-packages/webapp2.py", line 570, in dispatch
uwsgi_1           |     return method(*args, **kwargs)
uwsgi_1           |   File "./api/web/request.py", line 52, in log_user_access_from_request
uwsgi_1           |     result = handler_method(self, *args, **kwargs)
uwsgi_1           |   File "./api/web/base.py", line 197, in log_in
uwsgi_1           |     token_entry['_id'] = session_token
uwsgi_1           | TypeError: 'tuple' object does not support item assignment
uwsgi_1           |  request_id=cf46897d-1489427578
```

Issue #654, opened 2017-02-07
Transition from ObjectIDs to strings


kofalt:
We've talked about this forever, but I don't think we've ever filed it.

* [ ] acquisitions
* [x] authtokens
* [ ] batch (?)
* [x] devices
* [ ] downloads (?)
* [ ] gears
* [x] groups
* [ ] jobs
* [ ] projects
* [ ] sessions
* [x] singletons
* [x] sites
* [ ] uploads (?)
* [x] users

PR #655, opened 2017-02-08
User enabled gears, phase 1

### LINKED ISSUE #645: Initial (hidden) gear versioning
This ticket will handle the backend tracking of versions of gears.
The API will not (yet) acknowledge the distinction.

* [ ] `POST /api/gears`
Will now check for a gear by the given name and version, and fail if that combination already exists. Adds a created timestamp (if we don't do that already). Returns the gear ID. 

* [ ] `GET /api/gears`
Same as before, but will only return the most recently created gear for each gear name.

* [ ] `GET /api/gears/{id}`
Same as before, but will only return the most recently created gear for each gear name.
*Frontend Change to use id rather than name*

* [ ] `GET  /api/gears/{id}/suggest/{container_type}/{container_id}`
Same as before, but will use the most recently created gear for the given gear name.
*Frontend Change to use id rather than name*

* [ ] `GET  /api/gears/{id}/invocation`
Same as before, but will use the most recently created gear for the given gear name.

* [ ] `POST /api/jobs/add`
Endpoint now takes a gear ID, rather than a gear name.

* [ ] `GET /api/jobs/next`
On-the-fly formula generation now looks up gear by ID rather than name.

kofalt:
Allows for versioned gears to be stored, without disturbing existing endpoints too much.

Closes #645, which has the details of which endpoints changed.

@ryansanford This has two implications on the infra:

1. Gears should not be registered if they already exist
1. The database update script will take an extra a one-time one-second pause per existing gear document, for irritating reasons that aren't worth getting into

### Review Checklist

### DIFF
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -287,13 +287,16 @@ class AnalysesStorage(ListStorage):
         if 'analysis' not in tags:
             tags.append('analysis')
 
-        gear_name = job['gear']
+        gear_id = job['gear_id']
 
         destination = create_containerreference_from_dictionary({'type': 'analysis', 'id': analysis['_id']})
-        job = Job(gear_name, inputs, destination=destination, tags=tags, config_=job.get('config'), origin=origin)
+
+        job = Job(gear_id, inputs, destination=destination, tags=tags, config_=job.get('config'), origin=origin)
         job_id = job.insert()
+
         if not job_id:
             raise APIStorageException(500, 'Job not created for analysis {} of container {} {}'.format(analysis['_id'], cont_name, cid))
+
         result = self._update_el(cid, {'_id': analysis['_id']}, {'job': job_id}, None)
         return { 'analysis': analysis, 'job_id':job_id, 'job': job}
 
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -134,8 +134,11 @@ def run(batch_job):
     if not proposal:
         raise APIStorageException('The batch job is not formatted correctly.')
     proposed_inputs = proposal.get('inputs', [])
-    gear_name = batch_job.get('gear')
-    gear = gears.get_gear_by_name(gear_name)
+
+    gear_id = batch_job['gear_id']
+    gear = gears.get_gear(gear_id)
+    gear_name = gear['gear']['name']
+
     config_ = batch_job.get('config')
     origin = batch_job.get('origin')
     tags = proposal.get('tags', [])
@@ -155,9 +158,9 @@ def run(batch_job):
         if gear.get('category') == 'analysis':
 
             # Analysis gear, must create analysis on session
-            job = {
+            job_map = {
                 'config':   config_,
-                'gear':     gear_name,
+                'gear_id':  gear_id,
                 'inputs':   inputs,
                 'tags':     tags
             }
@@ -165,7 +168,7 @@ def run(batch_job):
             # Create analysis
             acquisition_id = inputs.values()[0].get('id')
             session_id = acq_storage.get_container(acquisition_id, projection={'session':1}).get('session')
-            result = an_storage.create_job_and_analysis('sessions', session_id, analysis, job, origin)
+            result = an_storage.create_job_and_analysis('sessions', session_id, analysis, job_map, origin)
             job = result.get('job')
             job_id = result.get('job_id')
 
@@ -175,7 +178,8 @@ def run(batch_job):
             for input_name, fr in inputs.iteritems():
                 inputs[input_name] = create_filereference_from_dictionary(fr)
             destination = create_containerreference_from_filereference(inputs[inputs.keys()[0]])
-            job = Job(gear_name, inputs, destination=destination, tags=tags, config_=config_, origin=origin)
+
+            job = Job(gear_id, inputs, destination=destination, tags=tags, config_=config_, origin=origin)
             job_id = job.insert()
 
         jobs.append(job)
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -2,9 +2,11 @@
 Gears
 """
 
-# import jsonschema
+import bson.objectid
+import datetime
 from jsonschema import Draft4Validator, ValidationError
 import gear_tools
+import pymongo
 
 from .. import config
 from .jobs import Job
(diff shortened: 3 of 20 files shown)


test/integration_tests/python/test_gears.py
```diff
@@ -39,3 +39,145 @@ def test_gear_add(as_admin):
...
+    assert r.ok
+
+    r = as_admin.get('/gears/multi-version')
+    assert r.ok
+    assert r.json()['gear']['name']    == 'multi-version'
+    assert r.json()['gear']['version'] == '0.0.1'
+
+    r = as_admin.get('/gears?fields=all')
+    assert r.ok
+    matched = filter(lambda x: x['gear']['name'] == 'multi-version', r.json())
```
nagem: this can be a list comprehension so we don't have to turn off pylint warnings:
``` 
matched = [x for x in r.json if x['gear']['name'] == 'multi-version']
```

test/integration_tests/python/test_gears.py
```diff
@@ -39,3 +39,145 @@ def test_gear_add(as_admin):
...
+    assert r.ok
+
+    r = as_admin.get('/gears/multi-version')
+    assert r.ok
+    assert r.json()['gear']['name']    == 'multi-version'
+    assert r.json()['gear']['version'] == '0.0.2'
+
+    r = as_admin.get('/gears?fields=all')
+    assert r.ok
+    matched = filter(lambda x: x['gear']['name'] == 'multi-version', r.json())
```
nagem: Same here.

test/integration_tests/python/test_gears.py
```diff
@@ -39,3 +39,145 @@ def test_gear_add(as_admin):
...
+    assert not r.ok
+
+    r = as_admin.get('/gears/multi-version')
+    assert r.ok
+    assert r.json()['gear']['name']    == 'multi-version'
+    assert r.json()['gear']['version'] == '0.0.2'
+
+    r = as_admin.get('/gears?fields=all')
+    assert r.ok
+    matched = filter(lambda x: x['gear']['name'] == 'multi-version', r.json())
```
nagem: and here

api/jobs/gears.py
```diff
@@ -13,32 +15,43 @@
...
 
-    if fields is None:
-        return config.db.gears.find()
-    else:
-        fields.append('gear.name')
+    cursor = config.mongo_pipeline('gears', pipe)
 
-    for f in fields:
-        projection[f] = 1
+    return map(lambda x: x['original'], cursor)
```
nagem: This one would be :
```
matched = [x['original'] for x in cursor] 
```

api/jobs/gears.py
```diff
@@ -142,11 +160,21 @@ def insert_gear(doc):
         })
         return job.insert()
 
-def remove_gear(name):
-    config.db.gears.remove({'gear.name': name})
+
+def remove_gear(_id):
+    config.db.gears.remove({"_id": _id})
```
nagem: This is [deprecated](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.remove), you probably want `delete_one` instead.

api/jobs/gears.py
```diff
@@ -119,12 +132,17 @@ def insert_gear(doc):
     if doc.get("invocation-schema"):
         del(doc["invocation-schema"])
 
+    now = datetime.datetime.utcnow()
+
+    doc['created']  = now
+    doc['modified'] = now
+
     config.db.gears.insert(doc)
```
nagem: You could set `doc['_id'] = str(bson.ObjectId()) before insertion so you don't have to worry about casting it as a string elsewhere.

api/dao/liststorage.py
```diff
@@ -290,7 +291,8 @@ def create_job_and_analysis(self, cont_name, cid, analysis, job, origin):
         gear_name = job['gear']
 
         destination = create_containerreference_from_dictionary({'type': 'analysis', 'id': analysis['_id']})
-        job = Job(gear_name, inputs, destination=destination, tags=tags, config_=job.get('config'), origin=origin)
+        gear_id = str(get_gear_by_name(gear_name)['_id'])
```
nagem: Why don't we require a gear id here yet? Is this a midstep until all clients support gear versioning?

api/jobs/batch.py
```diff
@@ -175,7 +175,8 @@ def run(batch_job):
             for input_name, fr in inputs.iteritems():
                 inputs[input_name] = create_filereference_from_dictionary(fr)
             destination = create_containerreference_from_filereference(inputs[inputs.keys()[0]])
-            job = Job(gear_name, inputs, destination=destination, tags=tags, config_=config_, origin=origin)
+            gear_id = str(gears.get_get_by_name(gear_name)['_id'])
```
nagem: Same here.

api/jobs/gears.py
nagem: This one I feel pretty adamantly about. It doesn't really fix our problem to remember to put bson.ObjectId(<id>) everywhere if we have to put str(<id>) in the opposite situations instead.

test/integration_tests/python/test_gears.py
```diff
@@ -39,3 +39,145 @@ def test_gear_add(as_admin):
         }
     })
     assert r.ok
+
```
nagem: Thanks for writing tests 😸

api/jobs/gears.py
kofalt: Agreed that it doesn't fix our problem - it's exactly as bad as `b.OID` everywhere. I would argue that it adds less work to #654 though.

kofalt: I could hide the typing in the map & mongo functions, and do a check type on init?

nagem: Why not just go entirely to setting the _id as a string then?

test/integration_tests/python/test_gears.py
nagem: Discussed offline, I think there are some instances where using a `map` or `filter` would be advantageous. Other instances, it's a case of style choice. Allowing them is fine.

api/jobs/gears.py
kofalt: The job docs are still an ID, and I didn't want to hack on #654 until we make a branch just for that.

ryansanford:
@kofalt 
Can you make sure to update/add relevant RAML docs? This link should display current snapshot from github on this branch: https://scitran.github.io/core/api/v1/index.html?branch=gear-versioning-p1

ryansanford:
@kofalt 
Is it intentional these 2 GETs now both return the whole gear doc? I'm testing with postman against the latest commit.

/api/gears?fields=all
/api/gears

kofalt:
Yes, I'm surprised you caught that :slightly_smiling_face: 

I removed that field because Dan never ended up using it, and it only adds implementation burden with this new setup. I'll update the docs.

PR #656, opened 2017-02-08
allow the info field to be null in a request

rentzso:
with this change it is possible to delete the info field completely

@nagem @gsfr was there a reason not to allow this?

### DIFF
--- a/raml/schemas/definitions/container.json
+++ b/raml/schemas/definitions/container.json
@@ -6,7 +6,7 @@
         "public":       {"type": "boolean"},
         "archived":     {"type": "boolean"},
         "label":        {"type": "string", "minLength": 1, "maxLength": 256},
-        "info":     {"type": "object"},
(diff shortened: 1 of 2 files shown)


nagem:
Is setting info to `{}` a problem? We also have a supported query param on `PUT` endpoints (`replace_metadata`) that allows you to completely replace the key with `{}`. Is that enough to satisfy the problem you have?

nagem:
It easier for clients if we have useful null values. Otherwise client code would have to check to see if there is a key, check to see if that key is `null`, and then display it. We currently overlay a list of defaults so clients can depend on some basic keys being available with useful nulls [here](https://github.com/scitran/core/blob/master/api/dao/containerstorage.py#L29).

rentzso:
From MATLAB it looks like it is not possible to set info in the json to `{}`.
But `replace_metadata` with something else should work for most use cases.

PR #659, opened 2017-02-09
Remove invalid regex on user ID

kofalt:
The ID regex could not accept uppercase letters. This isn't correct - the [OWASP](https://www.owasp.org/index.php/OWASP_Validation_Regex_Repository) has a better regex, but let's just follow the `email` field conventions until we change user IDs.

### Review Checklist


nagem:
LGTM

PR #685, opened 2017-03-09
Allow for gear optional inputs

kofalt:
This brings our gears library up to the latest version, which will allow for the config-verification of gears with optional inputs. 

Ref flywheel-io/gears#11,  flywheel-apps/example-gear#2.


lmperry:
LGTM

PR #686, opened 2017-03-10
add more logging to tmpdir cleanup and missing file error

cgc:


kofalt:
Can't see any harm in this, and it works fine in my tests. Merging!

PR #688, opened 2017-03-13
add missing parent containers on search results

### LINKED ISSUE #540: Project missing in analysis search results
Projects are not always returned when searching for analyses in the additional data attached to the analyses results.
Some investigation is needed.

CC @lmperry @nagem

rentzso:
closes #540

### Review Checklist

### DIFF
--- a/api/handlers/searchhandler.py
+++ b/api/handlers/searchhandler.py
@@ -5,6 +5,8 @@ from ..web import base
 from .. import config
 from .. import util
 from ..search import pathparser, queryprocessor, es_query, es_aggs
+from ..dao import APINotFoundException
+from ..dao.containerstorage import ContainerStorage
 
 log = config.log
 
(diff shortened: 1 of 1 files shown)


nagem:
I have concerns about mixing mongo results and elastic results as they will not always be in sync, but failures are manageable (not finding a parent because it was deleted, for example). I think this will work for now. It will slow results, as there are n*hierarchy height calls to mongo where n is the number of results.

rentzso:
@nagem I agree... 
Actually I planned to cache the results of the mongo queries.
So that we get a parent container only once (so for n acquisitions in the same session, you will retrieve the session only once, same for projects).

But it seems I have forgot to actually implement it. I'll do that as it is very simple.

I also agree about mixing ES and Mongo, but for such a simple operation Mongo seemed to be more straightforward (avoiding to have to define a new method getContainerFromES). With caching it shouldn't be a big hit on performance, but let me work on it a little bit more.

nagem:
@rentzso that caching plan sounds great 👍

rentzso:
@nagem I pushed a new commit. Could you give it a look? Thanks!

nagem:
Changes look good! 👍

Issue #690, opened 2017-03-14
Make schema validation errors a superset of standard errors


kofalt:
We have a few spots where we return client-parsable information about a validation error:

```python
return {
	'reason': 'Gear manifest does not match schema',
	'error': err.message.replace("u'", "'"),
	'key': key
}
```

The issue is that this does not overlap with the standard api error struct:

```python
{
	"message": "Something went wrong",
	"status_code": 500
}
```

It is hugely convenient if all errors are a superset of the same base struct.

I propose we keep message useful, and add the extra fields like so:

```python
message = err.message.replace("u'", "'")

return {
	'message':  'Gear manifest does not match schema on key ' + key ': ' + message,
	'status_code': 400,

	'reason': 'Gear manifest does not match schema',
	'error': message,
	'key': key
}
```

This keeps `message` a useful all-in-one, while still breaking out the client-friendly keys afterwards.

kofalt:
As an update, we'd probably just let the jsonschema.ValidationError bubble up and handle it in the standard error dispatch.

PR #691, opened 2017-03-14
Quick fix for validation struct

kofalt:
Ref #690. This is a quick fix to move some server-side stuff along, and we can solve #690 properly later.


nagem:
LGTM

PR #703, opened 2017-03-16
Increase coverage of api/jobs/handlers.py

ambrussimon-invenshure:
Closes #683

8 LOC left in api/jobs/handlers.py (97%)
- 5L Permission handling (that'll be faster to handle globally in #702): 244, 351, 354, 416-417
- 2L Checks that seemed unreachable (maybe obsolete or useful for compatibility?): 166, 222
- 1L That I would rather solve later with additional fixture(s) (need a session w/o acquisitions): 340

### Review Checklist

- Tests were added to cover all code changes
- Documentation was added / updated
- Code and tests follow standards in CONTRIBUTING.md

PR #706, opened 2017-03-16
Load job result before returning on /jobs/next

### LINKED ISSUE #705: /jobs/next does not use object load
Results in different job structure than `/jobs/<id>`. Relates #704, #420.

kofalt:
Fixes #705.


nagem:
LGTM, assuming it doesn't cause any tests to fail.

kofalt:
@ambrussimon-invenshure Just a head's up - this was trivial to fix, so I went and did it. Apologies if this messes up #703 - you may want to either rebase or merge in master.

PR #707, opened 2017-03-20
Fix bug when user attempts to log in with expired token in Authorization header

nagem:
Ignore `Authorization` header flow when serving results to `/api/login`.

### Review Checklist

### DIFF
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -66,7 +66,7 @@ class JWTAuthProvider(AuthProvider):
             'access_token': code,
             'uid': uid,
             'auth_type': self.auth_type,
-            'expires': datetime.datetime.utcnow() + datetime.timedelta(hours=1)
+            'expires': datetime.datetime.utcnow() + datetime.timedelta(days=14)
         }
 
     def validate_user(self, token):
@@ -128,7 +128,6 @@ class GoogleOAuthProvider(AuthProvider):
 
         response = json.loads(r.content)
         return {
-            'refresh_token': response['refresh_token'],
             'access_token': response['access_token'],
             'expires': datetime.datetime.utcnow() + datetime.timedelta(seconds=response['expires_in'])
         }
@@ -232,7 +231,7 @@ class WechatOAuthProvider(AuthProvider):
                     'attempted_user':   user['_id']
                 }
                 config.log_db.access_log.insert_one(log_map)
-                raise APIUnknownUserException('Another user is already registred with this Wechat OpenID.')
+                raise APIUnknownUserException('Another user is already registered with this Wechat OpenID.')
             update = {
                 '$set': {
                     'wechat.openid': openid
@@ -245,7 +244,7 @@ class WechatOAuthProvider(AuthProvider):
         else:
             user = config.db.users.find_one({'wechat.openid': openid})
         if not user:
-            raise APIUnknownUserException('User {} will need to be added to the system before managing data.'.format(user['_id']))
+            raise APIUnknownUserException('No user associated with this WeChat OpenID. Please use a valid registration link upon first sign in.')
         if user.get('disabled', False) is True:
             raise APIUnknownUserException('User {} is disabled.'.format(user['_id']))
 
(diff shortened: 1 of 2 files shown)


api/auth/authproviders.py
```diff
@@ -66,7 +66,7 @@ def validate_code(self, code, **kwargs):
             'access_token': code,
             'uid': uid,
             'auth_type': self.auth_type,
-            'expires': datetime.datetime.utcnow() + datetime.timedelta(hours=1)
+            'expires': datetime.datetime.utcnow() + datetime.timedelta(days=14)
```
nagem: Since JWT has no refresh flow, I set this to be the default of our server that issues JWTs.

api/web/base.py
```diff
@@ -184,6 +191,8 @@ def log_in(self):
 
         config.db.authtokens.insert_one(token_entry)
 
+        self.set_origin(False)
```
kofalt: What does this line do?

api/web/base.py
```diff
@@ -173,6 +178,8 @@ def log_in(self):
         token_entry = auth_provider.validate_code(payload['code'], registration_code=registration_code)
         timestamp = datetime.datetime.utcnow()
 
+        self.uid = token_entry['uid']
```
kofalt: What was broken without this line?

api/web/base.py
```diff
@@ -27,6 +27,11 @@ def __init__(self, request=None, response=None): # pylint: disable=super-init-no
         """Set uid, source_site, public_request, and superuser"""
         self.initialize(request, response)
 
+        # If user is attempting to log in through `/login`, ignore Auth here:
+        # In future updates, move login and logout handlers to class that overrides this init
+        if self.request.path == 'api/login':
+            return
```
kofalt: Ah. As an update to my "What was broken without this line?" comment, are these two pieces of code linked? Basically bypass legacy logic, and only `uid` was missing?

api/auth/authproviders.py
kofalt: I'm concerned about this line. Won't we cache any JWT session for two weeks without verifying?
Why can't we respect the JWT expiry, if specified?

api/auth/authproviders.py
```diff
@@ -128,7 +128,6 @@ def refresh_token(self, token):
 
         response = json.loads(r.content)
         return {
-            'refresh_token': response['refresh_token'],
```
kofalt: Why was this removed?

nagem:
Tested changes with client that requests `access_type=offline` to ensure Google auth requests get a refresh token. Will merge when green.

api/auth/authproviders.py
nagem: Google doesn't give us a refresh_token when requesting a new access_token like Wechat does.

api/web/base.py
nagem: Legacy wouldn't be the right word. This logic is used to AuthX a request when the header contains auth information (session token/drone info/API key). For login/out endpoints, we should bypass this logic as we aren't trying to evaluate the authx via the headers.

api/auth/authproviders.py
nagem: That is the JWT expiry for our server we are using. JWTs inherently can't be "refreshed". New work would need to be done to support a flow that "re-checks" rather than refreshes. That can be addressed in a separate PR. This change is to prevent users from being logged out every hour when their JWT token really lasts 14 days.

nagem: Google doesn't give us a refresh_token when requesting a new access_token like Wechat does.

kofalt: I'm all for leaving the refresh logic alone - my question was specifically about why we can't respect the JWT expiry instead of setting our own duration?

kofalt: Cool. Can you add either a comment talking about why this duration is set, or add a ticket to make the value configurable?

api/web/base.py
nagem: In order to set the origin we would need to set self.uid

api/web/base.py
nagem: Set the origin after the uid is known.

api/auth/authproviders.py
nagem: I'll add a ticket to make this easier to understand/request that our JWT service gives us an expiry timestamp.

PR #716, opened 2017-03-22
Ported tests from fixture api_as_admin to as_admin

ambrussimon-invenshure:
Context: [comments](https://github.com/scitran/core/compare/port-tests-from-fixture-api_as_admin-to-as_admin?expand=1#diff-7efe753e7372f603a4a51a7844e67980L3) [in](https://github.com/scitran/core/compare/port-tests-from-fixture-api_as_admin-to-as_admin?expand=1#diff-7efe753e7372f603a4a51a7844e67980L24) `tests/integration_tests/python/basics.py` indicated that [RequestsAccessor](https://github.com/scitran/core/compare/port-tests-from-fixture-api_as_admin-to-as_admin?expand=1#diff-4cfc2d319d5834744d4ada14c9d00b20L44) is obsolete. I was adding new tests with the preferred new fixtures from `basics.py` that are not relying on RequestsAccessor.

The fixture schism started getting messy and hindering test cov progress. Ported all old api_as_* fixture usages.

### Review Checklist

### DIFF
--- a/test/integration_tests/python/basics.py
+++ b/test/integration_tests/python/basics.py
@@ -1,85 +1,34 @@
-# New test fixtures!
-#
-# The intention is to slowly build these up until we like them, then port the tests to them and replace parts of conftest.py
+import os
 
 import pytest
 import requests
 
-# The request Session object has no support for a base URL; this is a subclass to embed one.
-# Dynamically generated via a pytest fixture so that we can use the upstream fixtures.
-@pytest.fixture(scope="module")
-def base_url_session(base_url):
-    class BaseUrlSession(requests.Session):
-        def request(self, method, url, params=None, data=None, headers=None, cookies=None, files=None, auth=None, timeout=None, allow_redirects=True, proxies=None, hooks=None, stream=None, verify=None, cert=None, json=None):
 
-            url = base_url + url
+class BaseUrlSession(requests.Session):
+    def __init__(self, *args, **kwargs):
+        super(BaseUrlSession, self).__init__(*args, **kwargs)
+        self.base_url = os.environ.get('BASE_URL', 'http://localhost:8080/api')
 
-            return super(BaseUrlSession, self).request(method, url, params, data, headers, cookies, files, auth, timeout, allow_redirects, proxies, hooks, stream, verify, cert, json)
-
-    return BaseUrlSession
+    def request(self, method, url, **kwargs):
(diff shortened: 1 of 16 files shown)


kofalt:
I initially added the `global _apiAsAdmin` caching because I was pretty sure that otherwise, one session object would be created per injection. This wouldn't be great because it wouldn't reuse open connections (though that removed comment should probably say TCP/SSL instead of HTTP... whatever).

Are you confident this approach will only create one session object for `as_admin` and one session object for `as_user`, regardless of how many tests run? If so, LGTM, if not, let me know if you think it would be smart to change that.

(Also note a small conflict caused by the previous PR landing)

ambrussimon-invenshure:
The original `global _apiAsAdmin` caching tied the object to the module's lifecycle, which is _roughly_ equivalent to pytest's `scope='session'`. The fixture previously being `'module'` scoped was a bit misleading for me, as even though pytest tried to re-inject it new for every module that was using it, the same cached instance got served without relying on the pytest scoping mechanism.

The fixture is now session scoped, and I verified that only one session is ever created and used by adding some debug lines

```diff
 @pytest.fixture(scope='session')
-def as_admin():
+def as_admin(request):
     session = BaseUrlSession()
     session.headers.update({'Authorization': 'scitran-user XZpXI40Uk85eozjQkU1zHJ6yZHpix+j0mo1TMeGZ4dPzIqVPVGPmyfeK'})
     session.params.update({'root': 'true'})
+    print('ADMIN SESSION ENTER: %s' % session)
+    def finalizer():
+        print('ADMIN SESSION EXIT: %s' % session)
+    request.addfinalizer(finalizer)
     return session
```

The option -s to the py.test call, then grepping for the output:

```
ambrus@raftel ~/workspace/scitran/core $ test/bin/run-tests-docker.sh -L | grep 'ADMIN SESSION'
ADMIN SESSION ENTER: <basics.BaseUrlSession object at 0x7fd073795210>
ADMIN SESSION EXIT: <basics.BaseUrlSession object at 0x7fd073795210>
ambrus@raftel ~/workspace/scitran/core $
```

Please let me know if this covers all concerns!

(PS.: Thanks for the heads up on the conflict, rebased!)

kofalt:
It does! LGTM, merge as desired.

kofalt:
@ambrussimon-invenshure Go ahead and delete PR branches after they land, if you're done with them.

PR #723, opened 2017-03-24
Optionally join in gear name on job docs in containerhandler

kofalt:
This addition is useful for knowing the name of the gear when asking for container origin joins.
Does not join the whole gear doc in, though that could happen if desired.

Following the same pattern as before, just add `join=origin_job_gear_name` to your request.

Before:
```
$ 'api/sessions/<id>/acquisitions?join=origin' \
  | jq -r '.[]."join-origin".job | .[] | .gear_id + " " + .gear_name'

58d542de31b1540016f9aefe 
58d542de31b1540016f9aefe 
58d542de31b1540019f9aeff 
58d542de31b1540016f9aefe
```

After:
```
$ 'api/sessions/<id>/acquisitions?join=origin&join=origin_job_gear_name' \
  | jq -r '.[]."join-origin".job | .[] | .gear_id + " " + .gear_name'

58d542de31b1540016f9aefe dcm2niix
58d542de31b1540016f9aefe dcm2niix
58d542de31b1540019f9aeff qa-report-fmri
58d542de31b1540016f9aefe dcm2niix
```

Tagging in @ryansanford for any performance concerns.
One additional DB call per job involving the container.

### Review Checklist

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -131,6 +131,9 @@ class ContainerHandler(base.RequestHandler):
                 Origin.job.name:    {}
             }
 
+        # Cache looked-up gears if needed
+        cached_gears = {}
+
         for f in result.get('files', []):
             origin = f.get('origin', None)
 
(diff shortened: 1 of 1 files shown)


ryansanford:
@kofalt 
Thoughts on pulling a gear_id/gear_name dict upon the first job when this join is requested, then doing in-memory lookups against that? I would have no performance concerns in that case. We would probably want a covering index for that dict building query.

With the current implementation, I'd want some well thought out testing to understand the impact.

kofalt:
We could definitely do that, this right now will redundantly load gears if multiple jobs came from the same gear.

kofalt:
This is now implemented.

nagem:
LGTM, as mentioned offline we won't get a lot out of the cache because it's only cache for each acquisition. Improvements could be made for performance by improving the `join_origin()` method to iterate over all acquisitions instead of each individually. If we see huge performance hits, this is where to look ;).

kofalt:
Hoorayyyyy

PR #726, opened 2017-03-27
increase coverage of api/handlers/containerhandler.py

### LINKED ISSUE #722: Increase coverage of api/handlers/containerhandler.py

ambrussimon-invenshure:
Closes #722 

```
Name                            Stmts   Miss  Cover   Missing
------------------------------------------------------------------
api/handlers/containerhandler.py  346     22    94%   99-100, 102, 135, 147, 204, 265, 268, 278, 337-338, 340, 382, 435-436, 441, 459-460, 465, 540, 549, 553
```

Found+fixed 3 minor bugs:
- GET /session/{sessionId}/jobs?join=containers - pop permissions from dict instead of _id
- GET /{containerName}?info=true - pop info from projection.copy() (otherwise server state and subsequent GETs change!)
- GET /{containerName}?count=true - DataBase has no get method

Missing coverage:
- 99-100: I _think_ it's unreachable due to api.py routes. TBD! More like this below.
- 102: Couldn't reach, need more info.
- 135,147: Effort. Also need info on how one can have missing/different origin via the api.
- 204: Effort - would be nice to have the updated data_builder.
- 265,268: Possibly unreachable due to api routes?
- 278: Effort - data_builder
- 337,338: Possibly unreachable due to api routes?
- 340: Effort - data_builder
- 382: Couldn't figure out how to, need help.
- 435,436: Possibly unreachable due to api routes?
- 465,540: Couldn't figure out how to, need help.
- 549: Possibly unreachable due to api routes?
- 553: Couldn't figure out how to, need help.

### Review Checklist

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -90,7 +90,7 @@ class ContainerHandler(base.RequestHandler):
         _id = kwargs.get('cid')
         self.config = self.container_handler_configurations[cont_name]
         self.storage = self.config['storage']
-        container= self._get_container(_id)
+        container = self._get_container(_id)
 
         permchecker = self._get_permchecker(container)
         try:
@@ -234,9 +234,9 @@ class ContainerHandler(base.RequestHandler):
         if join_cont:
             # create a map of analyses and acquisitions by _id
             containers = dict((str(c['_id']), c) for c in analyses+acquisitions)
-            for c in containers:
+            for container in containers.itervalues():
                 # No need to return perm arrays
-                c.pop('permissions', None)
+                container.pop('permissions', None)
             response['containers'] = containers
 
         return response
@@ -245,7 +245,7 @@ class ContainerHandler(base.RequestHandler):
         self.config = self.container_handler_configurations[cont_name]
         self.storage = self.config['storage']
 
-        projection = self.config['list_projection']
+        projection = self.config['list_projection'].copy()
         if self.is_true('info'):
             projection.pop('info')
             if not projection:
@@ -307,7 +307,7 @@ class ContainerHandler(base.RequestHandler):
     def _add_results_counts(self, results, cont_name):
         dbc_name = self.config.get('children_cont')
         el_cont_name = cont_name[:-1]
-        dbc = config.db.get(dbc_name)
+        dbc = config.db[dbc_name]
         counts =  dbc.aggregate([
             {'$match': {el_cont_name: {'$in': [res['_id'] for res in results]}}},
             {'$group': {'_id': '$' + el_cont_name, 'count': {"$sum": 1}}}
--- a/test/integration_tests/python/test_containers.py
+++ b/test/integration_tests/python/test_containers.py
@@ -200,7 +200,7 @@ def test_get_container(with_hierarchy, as_user, as_public):
     assert 'join-origin' in r.json()
 
 
-def test_get_session_jobs(with_hierarchy, with_gear, as_user, db):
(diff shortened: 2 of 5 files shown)


api/handlers/containerhandler.py
```diff
@@ -245,7 +245,7 @@ def get_all(self, cont_name, par_cont_name=None, par_id=None):
         self.config = self.container_handler_configurations[cont_name]
         self.storage = self.config['storage']
 
-        projection = self.config['list_projection']
+        projection = self.config['list_projection'].copy()
```
kofalt: Good catch.

test/integration_tests/python/test_jobs.py
```diff
@@ -94,3 +94,5 @@ def test_jobs(with_hierarchy, with_gear, with_invalid_gear, as_user, as_admin):
 
     r = as_user.post('/jobs/' + job1_id + '/retry')
     assert r.ok
+
+    db.jobs.remove()
```
kofalt: This is probably a dumb comment, but - this ingrains a pretty strong assumption that we won't be running unit tests in parallel. I'm guessing there are assumptions like that all over the tests; curious what your thoughts are.

kofalt:
Looks great to me, but let's hold for Megan since containerhandler & similar is her wheelhouse.

test/integration_tests/python/test_jobs.py
ambrussimon-invenshure: No, it's a very good point! I'll limit the scope of the `db.job.remove()`s I introduced. Sadly I can confirm that some assumptions like that are sprinkled here and there, though I think it's only a few.

I was thinking the DataBuilder could get some new features:
- store global db state (set of ids for all collections?) on init
- more flexibility for creation (enable users/gears/jobs + autocreate parent containers)
- keep track of resources created with the data_builder instance
- automatic deletion of resources (and any children thereof, even ones not created with this DataBuilder)
- assertions after deletion that the db state is unchanged

The global db state assertions could provide a unified way of making sure the tests aren't leaving anything behind and at the same time not deleting anything they didn't create. What do you think?

kofalt: That sounds like a great idea. Let's leave this code as-is for this PR, it sounds like we have a good path forward. I don't think running stuff in parallel is hugely important, but I gotta say doing it for the SDK tests feels very nice :slightly_smiling_face:

test/integration_tests/python/test_containers.py
```diff
@@ -126,3 +126,140 @@ def test_project_template(with_hierarchy, data_builder, as_user):
...
+            }
+        }
+    })
+    assert r.ok
+
+    r = as_user.get('/sessions/' + data.session + '/jobs', params={'join': 'containers'})
+    assert r.ok
+
+    # TODO figure out better/generalized way to handle state cleanup
+    # make sure other tests start clean w/o existing jobs (eg for /jobs/next to work as expected)
```
nagem: I wouldn't be opposed to adding the removal of all jobs relating to this gear to the `with_gear` fixture's `teardown_db` method.

nagem:
Great work as always, @ambrussimon-invenshure!

I didn't see anything off the bat, just a suggestion of where we could put job clearing. A couple of these tests are for old params to endpoints that aren't being used by any client currently and could probably be removed, but it's great to have them until we have time to sort through and remove the "dead weight". 

I'll look through your list of uncovered endpoints and see if I can add any context/other ideas, but for now feel free to merge this!

PR #735, opened 2017-04-04
Improve integration test flexibility

### LINKED ISSUE #730: Integration test pain points
I tried to identify some testing pain points.
Improving upon them would increase testing flexibility, readability and speed of progress.

- fixture scope / resource lifecycle visibility
    + caused the states.py issue
    + not always clear when creation or removal happens
    + not always clear when modification might impact others
    + incomplete cleanup error traceability
- scope / lifecycle flexibility
    + with_ tied to databuilder lifecycle (module)
    + cleanup clutter within individual tests
    + needed for balance between # requests and speed
- envvar traceability

ambrussimon-invenshure:
Closes #730 

- revised data_builder
    + reduced scope to `function` (removes any inter-dependence between tests, getting the tests closer to parallel execution; it leads to more requests, but didn't run into any performance issues.)
    + supports gears and users (with api keys)
    + auto-creates parent containers (eg if test only needs to reference an acquisition, it may explicitly create only that, parents will be implicitly added)
    + auto-deletes any resource created with the data_builder (explicitly or implicitly)
    + enables tests to recursively delete a resource (eg uid upload test creates hierarchy under a group via an upload request; which can be cleaned up calling a recursive delete on the group)
    + as a consequence the with_* fixtures were replaced w/ pure data_builder usage
- revised requests accessors (as_*)
    + got rid of the external bootstrappers and now inject users using the new as_drone accessor
    + user and admin used to have the same access key
    + separated as_admin and as_root for now
- cleanup tasks
    + passing in and loading env vars without renaming (eg SCITRAN_CORE_DRONE_SECRET)
    + remove unused/sort imports
    + remove unused logger setups
    + switched from bunch to attrdict

PS: Sorry about the ginormous diff :|

### Review Checklist


nagem:
I looked over the diff again today and didn't see anything we haven't already discussed. Great work! I'm excited to have this merged so I can use it to write some tests for some PRs I have open 😉.

Since tests are passing I don't have any worry about unseen issues in the mammoth diff. Go ahead and merge when you're ready!

kofalt:
Just stopping by to say this is solid work :+1:

PR #740, opened 2017-04-05
Change API key format and allow for colon-separated key preambles

### LINKED ISSUE #668: Remove dashes from API key rune set
Dashes in API keys prevent double-click selecting, which in practice irritates the heck out of me copying them from the web form. Let's not have a user mistakenly only copy part of their key & get confused :slightly_smiling_face:

kofalt:
The current API key format takes some entropy and base64 encodes it. This ends up being longer than it needs to be, and has some formatting issues (#668 among other things).

Instead, generate an alphanum string directly, using the same entropy source.
Additionally, allow for arbitrary preambles to be added to a key, separated by `:`.
This can hold, for example, connection information for passing between connected products.

Closes #668.

### Review Checklist

### DIFF
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -75,7 +75,6 @@ class AuthProvider(object):
         }
         dbutil.fault_tolerant_replace_one('refreshtokens', query, refresh_doc, upsert=True)
 
-
 class JWTAuthProvider(AuthProvider):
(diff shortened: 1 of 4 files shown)


nagem:
LGTM, glad this will allow the CLI auth/uri config to be easier to use.

PR #743, opened 2017-04-10
add roles & permissions to session projection to enable has_access

ambrussimon-invenshure:
During testing I found uid-match uploads weren't working due to a minor bug, where has_access always failed because it was called with a session object that was missing the roles/permissions attrs due to a projection.

### Review Checklist


nagem:
LGTM! Thanks for fixing this.

PR #744, opened 2017-04-10
Gear name on job tabs and database upgrade speed improvements

### LINKED ISSUE #734: Job gear-name tags, and DB upgrade speed improvements
1. Add code to `/jobs/next` to look up the gear name and add it to the job's tags
1. Write a migration script to add the gear name to any job doc that doesn't have it
1. Investigate options for increased DB upgrade speed:
    1. Javascript query that is handed off to the database for it to chew on
    1. Fetch multiple jobs from cursor at a time, batch update
    1. Python workers to run a closure over a cursor in parallel
1. Write a synthetic script that tries one or more of these strategies & logs the wall-clock of each.
1. Hand script to @ryansanford and see how things perform on production-like data sets.
1. Choose a winning strategy and set it up for this & future upgrades :sunglasses: 

Ref scitran/core#378.

kofalt:
This patch has two components:

1) New jobs will now have the gear name added as a tag on the job

~2) A synthetic upgrade to test speed improvements~

3) An upgrade to add the gear name to all existing job tags.

~Once the upgrade is confirmed faster (via @ryansanford), and general approach approved (via @nagem), then the upgrade will be swapped out for a real one.~

The new upgrade strategy is a slap-happy approach to the worst upgrade times we see right now, where a table cursor is iterated over and the individual upgrade order does not matter.

In this case we can just throw multiprocessing at the problem.
Time is measured in seconds for this benchmark.

With ~ 13,000 documents:

```
Serial   5.110
Parallel 1.520
```

With ~ 126,932 documents:

```
Serial   47.740
Parallel 13.940
```

The pool is currently set to num-v-cpu-cores, for no discernible reason. It could be made tunable if desired.
Closes #734.

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -60,9 +60,12 @@ def getMonotonicTime():
 def process_cursor(cursor, closure):
     """
     Given an iterable (say, a mongo cursor) and a closure, call that closure in parallel over the iterable.
-    Call order is undefinied. Currently launches N python process workers, where N is the number of vcpu cores.
+    Call order is undefined. Currently launches N python process workers, where N is the number of vcpu cores.
 
     Useful for upgrades that need to touch each document in a database, and don't need an iteration order.
+
+    Your closure MUST return True on success. Anything else is logged and treated as a failure.
+    A closure that throws an exception will fail the upgrade immediately.
     """
 
     begin = getMonotonicTime()
@@ -78,14 +81,21 @@ def process_cursor(cursor, closure):
     results = [pool.apply_async(closure, (document,)) for document in cursor]
 
     # Read the results back, presumably in order!
+    failed = False
     for res in results:
         result = res.get()
         if result != True:
+            failed = True
             logging.info('Upgrade failed: ' + str(result))
 
     logging.info('Waiting for workers to complete')
     pool.close()
     pool.join()
+
+    if failed is True:
+        msg = 'Worker pool experienced one or more failures. See above logs.'
+        logging.info(msg)
+        raise Exception(msg)
 
     end = getMonotonicTime()
     elapsed = end - begin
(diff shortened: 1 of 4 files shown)


kofalt:
@ryansanford reports ~55% speed increase :tada: 
@nagem reports approach acceptance :tada: 

Swapping the synthetic upgrade for a real one.

kofalt:
New upgrade is work :tada: 

```
db.getCollection('jobs').find({}, {"tags": 1})
```

```
/* 1 */
{
    "_id" : ObjectId("58ee9636acddae001beb1678"),
    "tags" : [ 
        "dcm2niix"
    ]
}

/* 2 */
{
    "_id" : ObjectId("58ee9636acddae001beb1679"),
    "tags" : [ 
        "dicom-mr-classifier"
    ]
}
```

bin/database.py
```diff
@@ -51,6 +53,44 @@ def confirm_schema_match():
...
+    # Note that this creates an array of n multiprocessing.pool.AsyncResults, where N is table size.
+    # Memory usage concern in the future? Doesn't seem to be an issue with ~120K records.
+    # Could be upgraded later with some yield trickery.
+    results = [pool.apply_async(closure, (document,)) for document in cursor]
+
+    # Read the results back, presumably in order!
+    for res in results:
+        result = res.get()
+        if result != True:
+            logging.info('Upgrade failed: ' + str(result))
```
nagem: If an upgrade doesn't succeed isn't that a level of failure that should fail the upgrade?

bin/database.py
```diff
@@ -51,6 +53,44 @@ def confirm_schema_match():
...
+def getMonotonicTime():
+    # http://stackoverflow.com/a/7424304
+    return os.times()[4]
+
+def process_cursor(cursor, closure):
+    """
+    Given an iterable (say, a mongo cursor) and a closure, call that closure in parallel over the iterable.
+    Call order is undefinied. Currently launches N python process workers, where N is the number of vcpu cores.
+
+    Useful for upgrades that need to touch each document in a database, and don't need an iteration order.
```
nagem: Maybe just add a comment that the expected return value of the closure is either `True` or a `str` error?

kofalt:
Comments addressed

PR #748, opened 2017-04-12
Adapt to new reaper options (--secret, --key)

gsfr:
This is the start of getting core aligned with the new option parsing capabilities of reaper.
scitran/reaper#54

@ryansanford Please add additional commits here as needed.

**Note that `@options` needs to be removed from `bin/run-dev-osx.sh:184`, and `docker/bootstrap-data.sh:28` needs to be reverted to `1.0.3` or a more recent release.**

### DIFF
--- a/bin/install-dev-osx.sh
+++ b/bin/install-dev-osx.sh
@@ -86,5 +86,5 @@ fi
 
 # Install testing dependencies
 echo "Installing testing dependencies"
-pip install -r "test/integration_tests/requirements-integration-test.txt"
+pip install --no-cache-dir -r "test/integration_tests/requirements-integration-test.txt"
 npm install test/integration_tests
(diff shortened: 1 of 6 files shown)


ryansanford:
@gsfr The docker change looks and works well.

bin/install-dev-osx.sh
```diff
@@ -86,5 +86,5 @@ fi
 
 # Install testing dependencies
 echo "Installing testing dependencies"
-pip install -r "test/integration_tests/requirements-integration-test.txt"
+pip install --no-cache-dir -r "test/integration_tests/requirements-integration-test.txt"
```
kofalt: Is this just to suppress some warnings?

kofalt:
LGTM, one quick question.

bin/install-dev-osx.sh
gsfr: It's actually just to align with another `pip` command above, to ensure consistent behavior.

gsfr:
I'm all set here, bar the two noted pre-merge changes, once scitran/reaper#54 has landed.

@kofalt Could you please take another look. To get things right in the venv on the Mac, I had to make a small change in `run-integration-tests.sh`. Seems harmless. [Famous last words...]

PR #754, opened 2017-04-13
Up result size for api/search/fields

nagem:
Return a max of 1000 buckets for fields aggregation instead of Elastic's default of 10.

In the future we could make this an option for clients via a query string. 
### Review Checklist


nagem:
Small change, tested with 20 session tags, works as expected. Will merge when CI is green.

PR #768, opened 2017-04-24
fix AnalysisHandler.download when ticket is None

ambrussimon-invenshure:
### Review Checklist


nagem:
👍

api/handlers/listhandler.py
```diff
@@ -776,6 +776,7 @@ def download(self, cont_name, list_name, **kwargs):
         permchecker, storage, _, _, _ = self._initialize_request(cont_name, list_name, _id)
         filename = kwargs.get('name')
         ticket_id = self.get_param('ticket')
+        ticket = None
```
nagem: This line specifically fixes an issue we ran into today so nice work 😉

nagem:
Merging since we need the fix.

PR #770, opened 2017-04-26
Move uid upload logic to 'reaper' class, modify existing uid logic to always follow group/project

nagem:
The old `UID` placer logic would ignore the project/group information in the metadata if it found the session via `UID` in a separate project to support moving sessions during data acquisition. That logic has been moved to a new endpoint at `api/upload/reaper`. The existing endpoint `api/upload/uid` has been changed to always follow what the metadata specifies for project/group.

### Review Checklist

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -84,10 +84,10 @@ endpoints = [
 
         # General-purpose upload & download
 
-        route('/download',                              Download, h='download',              m=['GET', 'POST']),
-        route('/upload/<strategy:label|uid|uid-match>', Upload,   h='upload',                m=['POST']),
-        route('/clean-packfiles',                       Upload,   h='clean_packfile_tokens', m=['POST']),
-        route('/engine',                                Upload,   h='engine',                m=['POST']),
+        route('/download',                                      Download, h='download',              m=['GET', 'POST']),
+        route('/upload/<strategy:label|uid|uid-match|reaper>',  Upload,   h='upload',                m=['POST']),
+        route('/clean-packfiles',                               Upload,   h='clean_packfile_tokens', m=['POST']),
+        route('/engine',                                        Upload,   h='engine',                m=['POST']),
 
 
         # Top-level endpoints
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -418,7 +418,7 @@ def _get_targets(project_obj, session, acquisition, type_, timestamp):
     return target_containers
 
 
-def find_existing_hierarchy(metadata, user=None, site=None):
+def find_existing_hierarchy(metadata, type_='uid', user=None, site=None):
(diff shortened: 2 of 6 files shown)


gsfr:
Thanks, @nagem. Seems to work fine.

@ambrussimon-invenshure We added a new upload endpoint to respect session and acquisition UIDs in a more intuitive way. The current `/upload/uid` endpoint is now `/upload/reaper` and the new one has taken over `/upload/uid`.

The reaper endpoint will find an existing session by UID, even if it is not in the referenced group and project, while the UID endpoint will always respect the given information. (Users sometimes move a session to a different project, while data is still being acquired, and still expect the session to remain intact. That's what the reaper endpoint is for.)

Could you please add/update our integration testing to account for the updated/added endpoints?

I would also like to see more comprehensive tests that send two uploads to each endpoint and verify the correct outcome.

Below are two example payloads. Depending on the endpoint used, you should either get one or two sessions:
- `/upload/label` --> one session
- `/upload/uid` --> two session

And, if you modify the second payload to have the same session UID as the first, then with `/upload/reaper`, you should be able to move the session to a different project after the first upload and still only get one session.

Does our current framework allow for such complex tests?

```
{
    "group": {
        "_id": "group"
    },
    "project": {
        "label": "project"
    },
    "session": {
        "subject": {
            "code": "subj1"
        },
        "uid": "sess1",
        "label": "sess"
    },
    "acquisition": {
        "files": [
            {
                "name": "f1"
            }
        ],
        "uid": "acq1",
        "label": "acq"
    }
}

{
    "group": {
        "_id": "group"
    },
    "project": {
        "label": "project"
    },
    "session": {
        "subject": {
            "code": "subj1"
        },
        "uid": "sess2",
        "label": "sess"
    },
    "acquisition": {
        "files": [
            {
                "name": "f1"
            }
        ],
        "uid": "acq2",
        "label": "acq"
    }
}
```

gsfr:
I’d like to get this out this week. @lmperry (for Ansgar) and Steve want to use the dicom folder sniper and they should be using the new endpoint from this PR.

@ryansanford @ambrussimon-invenshure Can we reasonably expect to get testing in by the end of the week or should we proceed without?

ambrussimon-invenshure:
I've looked at the scope on Friday and I'm sure that all tests mentioned
can be added before the weekend.

ambrussimon-invenshure:
@gsfr @nagem I added the new endpoint and the more thorough uid/label tests. Please note that I switched the reaper JSON schema, I assumed it was a typo.

gsfr:
Thanks, @ambrussimon-invenshure. I think that schema change is right, but @nagem should confirm.

Otherwise, LGTM.

nagem:
LGTM too, schema change was the correct call. Thanks!

PR #771, opened 2017-04-28
Add config option to disable proper SSL connection verification to the JWT Auth service

nagem:
WARNING: Should only be used for testing environments.


ryansanford:
LGTM, thanks!

PR #772, opened 2017-04-28
Increase coverage low hanging fruit 3

ambrussimon-invenshure:
Low hanging coverage vol. 3 - 80%

Changes in api/ files:
- whitespace corrections
- note and no cover pragma comments for [api.dao.hierarchy.merge_fileinfos](https://github.com/scitran/core/compare/increase-coverage-low-hanging-fruit-3?expand=1#diff-de688aada2a9bbfba0e67dad93cd5508R575)
- handle request.json only after checking that project exists in [api.jobs.handlers.RuleHandler.put](https://github.com/scitran/core/compare/increase-coverage-low-hanging-fruit-3?expand=1#diff-8ae88fa922271b27b6066aae46fda7abR165)
- eliminate dead code in [api.jobs.handlers.JobHandler.get_logs](https://github.com/scitran/core/compare/increase-coverage-low-hanging-fruit-3?expand=1#diff-8ae88fa922271b27b6066aae46fda7abR336) and add_logs. @kofalt I added this before you got back to me. I didn't start with introducing the custom error classes as that seems too big, and I think the excepts might be cleaner for the time being than leaving the 500s. Please take a look and let me know if you would still prefer the (or any) other fix.

### Review Checklist


kofalt:
LGTM. I like the rule changes :+1: 

Reporting changes not reviewed; I leave that to Megan.

api/dao/hierarchy.py
```diff
@@ -572,7 +572,8 @@ def _update_container_nulls(base_query, update, container_type):
     return config.db[coll_name].find_one(base_query)
 
 
-def merge_fileinfos(parsed_files, infos):
+# NOTE skip coverage since this function is currently not used
+def merge_fileinfos(parsed_files, infos): # pragma: no cover
```
nagem: Thanks for calling that out - I'm doing some work in this file and will just remove it in that PR.

test/integration_tests/python/test_download.py
```diff
@@ -98,3 +99,128 @@ def test_download(data_builder, file_form, as_admin, api_db):
...
+    r = as_admin.get(analysis_files + '/two.zip', params={'ticket': ticket, 'info': 'true'})
+    assert r.ok
+
+    # try to get non-existent zip member
+    r = as_admin.get(analysis_files + '/two.zip', params={'ticket': ticket, 'member': 'nosuch'})
+    assert r.status_code == 400
+
+    # get zip member
+    r = as_admin.get(analysis_files + '/two.zip', params={'ticket': ticket, 'member': 'two.csv'})
+    assert r.ok
```
nagem: This part of the product definitely needed testing - nice work!

test/integration_tests/python/test_uploads.py
```diff
@@ -134,8 +149,22 @@ def test_uid_upload(data_builder, file_form, as_admin, as_user, as_public):
     r = as_user.post('/upload/uid', files=file_form(*uid_files, meta=uid_meta))
     assert r.status_code == 403
 
+    # TODO figure out why api.dao.hierarchy._group_id_fuzzy_match is NOT called below
```
nagem: Is that function not called at all, or just not the fuzzy match part? I can look into this too. I don't think it holds up this merging, though.

PR #791, opened 2017-05-18
Database upgrade double tagging fix

### LINKED ISSUE #785: DB update #26 adds gear name to tag list regardless of tag already being present
Some jobs were already tagged with their gear's name in existing data sets so `upgrade_to_26` ends with some jobs having the gear name listed twice in the job tag list.

Since double tags don't really cause any issue, we're just going to update the existing db update instead of creating a new one that cleans up the problem. 

If there is an efficient way to only process jobs that don't already have the gear tag that would improve the speed at which this update runs as well. Maybe get a list of all gear names from the gears table and then modify the `find` on the jobs table to only return jobs that don't have any tags that match the list of gear names.

hkethi002:
Fixes #785 
Some of the changes in upgrade_to_26() are only for testing
### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -974,7 +974,7 @@ def upgrade_to_26_closure(job):
         if tag == gear_name:
             gear_name_tag_count = gear_name_tag_count + 1
     if gear_name_tag_count > 1:
-        raise Exception("Job has multiple gear name tags")
+        logging.info("job "+ job + " has multiple gear name tags")
 
     # Update doc
     result = config.db.jobs.update_one({'_id': job['_id']}, {'$push': {'tags': gear_name }})
--- a/bin/database.py
+++ b/bin/database.py
@@ -973,7 +973,6 @@ def upgrade_to_26_closure(job):
 
 
     # Update doc
-    result = config.db.jobs.update_one({'_id': job['_id']}, {'$addToSet': {'tags': gear_name }})
 
     # Check for multiples
     for tag in job['tags']:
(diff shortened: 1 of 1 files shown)


bin/database.py
```diff
@@ -965,22 +965,36 @@ def upgrade_to_26_closure(job):
...
         return True
     else:
         return 'Parallel failed: update doc ' + str(job['_id']) + ' resulted modified ' + str(result.modified_count)
 
+def upgrade_to_26_test_closure(job):
+
+    for count, tag in enumerate(job['tags'],1):
+        logging.info(count)
+        if count > 2:
+            raise Exception("Job has multiple gear tags")
```
kofalt: So, you can actually have arbitrary tags. Gear name is (supposed to be) an always-present tag, but there can be others. I think our infrastructure will add `ad-hoc` and `analysis` for you depending on how the job is enqueued, and you can set any number of your own tags with `/jobs/add?tags=xyz`.

bin/database.py
```diff
@@ -965,22 +965,36 @@ def upgrade_to_26_closure(job):
...
+
 def upgrade_to_26():
     """
     scitran/core #734
 
     Add job tags back to the job document, and use a faster cursor-walking update method
     """
-
-    cursor = config.db.jobs.find({})
+    cursor = config.db.jobs.find({'tags[1]': {'$exists': True}})
```
kofalt: Same issue here, you won't be able to check this condition ahead-of-time.
It might have a tag that is unrelated to the gear name.

nagem:
LGTM! Thanks, @hkethi002!

Issue #763, opened 2017-04-19
Usage report failing on malformed file create timestamps


nagem:
Some datasets seem to have malformed created timestamps for their files:

```
{"status_code": 500, "message": "command SON([('aggregate', 'projects'), ('pipeline', [{'$unwind': '$files'}, {'$match': {}}, {'$project': {'mbs': {'$divide': ['$files.size', 1048576.0]}, 'year': {'$year': '$files.created'}, 'month': {'$month': '$files.created'}}}, {'$group': {'_id': {'year': '$year', 'month': '$month'}, 'mb_total': {'$sum': '$mbs'}}}])]) on namespace scitran.$cmd failed: can't convert from BSON type EOO to Date"}
```

nagem:
A little reading leads me to believe they have an absent timestamp, not a malformed one.

nagem:
The data that caused this issue is from an older data set, it is assumed there are no current problem with core not properly setting a file `created` key. A db update to resolve the missing fields should fix the problem. The file can be given the `created` key of it's parent. Also ensure it has a `modified` timestamp.

PR #792, opened 2017-05-18
Create timestamps for Analysis files if it doesn't have one

### LINKED ISSUE #759: Analysis files do not have created timestamps
To patch existing files, give them the created time of their analysis. 

When this is fixed, update the Usage Report to use these timestamps.

hkethi002:
Fixes #759
### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -1054,7 +1054,7 @@ def upgrade_to_29():
     users = config.db.users.find({})
     process_cursor(users, upgrade_to_29_closure)
 
-def upgrade_to_30_closure_analysis(session):
+def upgrade_to_30_closure_session_analysis(session):
     analyses = session.get('analyses', None)
 
     if analyses is not None:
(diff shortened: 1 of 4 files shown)


ryansanford:
@nagem 
On latest master, I'm not seeing `created` field for analysis output files. I assume that needs to be addressed before this DB update?

Also, the `created` field doesn't get added for input files if that field is missing from the file record in the acquisition/analysis.

nagem:
@ryansanford yeah the fix in handler code still needs to be done. 

>Also, the created field doesn't get added for input files if that field is missing from the file record in the acquisition/analysis.

There is another ticket that cleans up other files, that should perhaps go first, but I don't think it should be an analysis's job to "clean up" data sent to it.

nagem:
Transcribing a tl;dr of an offline conversation:

Since there are files in the system not attached to analyses that do not have `created` fields (#763), a fix for that bug will need to go in first. If a project/session/acquisition that has a file attachment without a `created` key gets added as input to an analysis, the analysis will now also have a file without a `created` key, causing more instances of this issue. 

@hkethi002 plans to add the file cleanup for projects/sessions/acquisitions to this upgrade as well.

gsfr:
@hkethi002 Thanks for cleaning up the commits.

bin/database.py
```diff
@@ -1054,6 +1054,75 @@ def upgrade_to_29():
...
 
+def upgrade_to_30_closure_analysis(session):
+    analyses = session.get('analyses', None)
+
+    if analyses is not None:
+        for analysis_ in analyses:
+            files = analysis_.get('files', [])
+            for file_ in files:
+                if 'created' not in file_:
+                    file_['created'] = analysis_['created']
```
ryansanford: Given unexpected missing fields on other objects, I wouldn't assume `analysis_['created']` exists. I don't have a strong opinion on what the fallback should be if that's the case. Using something like epoc would allow easy identification in the future.

Same comment for the other closure methods. Test both present and missing `created` source keys, and ensure the db update completes fully.

bin/database.py
```diff
@@ -1054,6 +1054,75 @@ def upgrade_to_29():
...
+    cursor = config.db.sessions.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis)
+    cursor = config.db.sessions.find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_sessions)
+    cursor = config.db.acquisitions.find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_acquisitions)
+    cursor = config.db.projects.find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_projects)
+
```
ryansanford: There are also sites with analyses within config.db.collections objects. Those would need similar restitution.

bin/database.py
```diff
@@ -58,7 +58,7 @@ def getMonotonicTime():
     # http://stackoverflow.com/a/7424304
     return os.times()[4]
 
-def process_cursor(cursor, closure):
+def process_cursor(cursor, closure, context = None):
```
nagem: 👍

bin/database.py
```diff
@@ -1054,6 +1057,67 @@ def upgrade_to_29():
...
+    cursor = config.db.collections.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis, config.db.collections)
+
+    cursor = config.db.sessions.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis, config.db.sessions)
+
+    cursor = config.db.sessions.find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_coll, config.db.sessions)
```
nagem: I try to be explicit when using python optional arguments, using:
`process_cursor(cursor, upgrade_to_30_closure_coll, context=config.db.sessions)`
rather than
`process_cursor(cursor, upgrade_to_30_closure_coll, config.db.sessions)`

Python will start filling in the optional args in order if they are not named, but it's easier to read and less likely to cause programmer errors when they are referenced explicitly. (For example, if someone else changes the order of the optional args that would break this call).

bin/database.py
```diff
@@ -1054,6 +1057,67 @@ def upgrade_to_29():
...
+    config.db.sessions.insert_one({'created': datetime.datetime.utcnow(), 'files': [{'name': "file1"}]})
+    config.db.acquisitions.insert_one({'created': datetime.datetime.utcnow(), 'files': [{'name': "file1"}]})
+
+    config.db.projects.insert_one({'files': [{'name': "file2"}]})
+    config.db.collections.insert_one({'analyses': [{'files': [{'name': 'file1'}]}]})
+    config.db.collections.insert_one({'analyses': [{'created': datetime.datetime.utcnow(), 'files': [{'name': 'file2'}]}]})
+
+    cursor = config.db.collections.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis, config.db.collections)
```
nagem: Just an existing pattern in the API is to pass the name of the collection, not a reference to the collection itself when doing work like this. For example:

`process_cursor(cursor, upgrade_to_30_closure_analysis, content='collections')`

and then in the closure:
`config.db[context].insert(...)`

But that is just a style choice and doesn't really affect the result of the code.

nagem:
Changes look good! When the testing work is removed, I think this is ready to go in!

@ryansanford want to do any reviews you think are necessary and change your review status so this doesn't go in as red? @hkethi002 make sure Ryan's review status is green before merging.

bin/database.py
```diff
@@ -1054,6 +1057,60 @@ def upgrade_to_29():
...
+
+    cursor = config.db.collections.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis, context = 'collections')
+
+    cursor = config.db.sessions.find({'analyses': {'$exists': True},
+                                                       'analyses.files.created': {'$exists': False}})
+    process_cursor(cursor, upgrade_to_30_closure_analysis, context = 'sessions')
+
+    cursor = config.db.sessions.find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
```
nagem: If you wanted to get a bit cheeky you could do something like:

```
for coll in ['projects', 'sessions', 'acquisitions', 'collections']:
    cursor = config.db[coll].find({'files': {'$exists': True}, 'files.created': {'$exists': False}})
    process_cursor(cursor, upgrade_to_30_closure_coll, context = coll)
    if coll in ['sessions', 'collections']:
        cursor = config.db[coll].find({'analyses': {'$exists': True}, 'analyses.files.created': {'$exists': False}})
        process_cursor(cursor, upgrade_to_30_closure_analysis, context = coll)
```
But don't feel like you need to change it. Just wanted to show a way you can take advantage of  db[<collection_name>].

PR #797, opened 2017-05-24
Enforce https for avatars

### LINKED ISSUE #776: Do not allow non-https avatar URLs
Creates "mix-content" warning on browsers. Also rectify any existing http avatar URLs.

hkethi002:
Fixes #776
The inserted test cases in upgrade_to_29() will be removed before merging
### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -1032,9 +1032,13 @@ def upgrade_to_29_closure(user):
     avatars = user['avatars']
     if avatars.get('custom') and not 'https:' in avatars['custom']:
         if user['avatar'] == user['avatars']['custom']:
-            config.db.users.update_one({'_id': user['_id']},
-                {'$set': {'avatar': user['avatars'].get('provider', None)}}
-            )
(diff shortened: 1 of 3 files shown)


bin/database.py
```diff
@@ -1024,6 +1025,32 @@ def upgrade_to_28():
...
+        )
+    return True
+
+
+def upgrade_to_29():
+    """
+    Enforces HTTPS urls for user avatars
+    """
+    
+    users = config.db.users.find({})
```
nagem: I just thought of one other thing: Since we always try to limit the number of items we have to iterate over in db updates, you could modify the `find` to only return users that have `avatar.custom` set.

nagem:
Change to just null out the avatar if provider does not exist looks good!

hkethi002:
So by null out, do you mean unsetting it?

nagem:
Yep! I see now you are calling an unset, my mistake.

nagem:
LGTM! Feel free to merge.

Issue #798, opened 2017-05-31
Failing test on master 


kofalt:
There's a test failing on master:

```python
def test_usage_report(data_builder, file_form, as_user, as_admin):
        assert r.ok
        usage = r.json()
        >       assert len(usage) == 1
E       assert 2 == 1
E        +  where 2 = len([{'file_mbs': 1.9073486328125e-05, 'gear_execution_count': 1, 'month': '5', 'session_count': 1, ...}, {'file_mbs': 0, 'gear_execution_count': 0, 'month': '6', 'session_count': 0, ...}])
```

I'd look into it myself, but I know very little about usage reporting...
Ambrus, mind having a look while Megan's out?

kofalt:
From Ambrus: this test makes an assumption that tomorrow is in the same month, but as we're at the last day of the month, this assumption doesn't hold. Tomorrow tests will pass again.

Not quite a phase-of-the-moon test, but close enough.

![werewolf](https://cloud.githubusercontent.com/assets/265071/26645287/238cf70e-45fd-11e7-92d0-18c2a4928c7b.jpg)

PR #799, opened 2017-05-31
Quick fix to SSE upload race

kofalt:
If a connection breaks during packfile-end, keep going rather than crashing with a low-level I/O error.
Prevents a race condition where if the connection breaks, the packfile may or may not succeed.


kofalt:
CI failed due to unrelated test, ref #798.

kofalt:
Merging based on offline comments from @ryansanford.

PR #800, opened 2017-05-31
Group permissions

### LINKED ISSUE #662: Rename `group.roles` --> `group.permissions`
1. Inconsistency (this is aliased away in the SDK)
1. Despite the name, roles are actually permissions in that they control R/W access to that group

hkethi002:
Fixes #662 and #814 
Will not be merged until frontend is changed to reflect the changes made to the endpoints 
_Fields Changed_
- group.roles -> group.permissions
- group.name -> group.label

_Endpoint Parameter_
- Non admin GET request return user's groups with all permissions for those groups
- Admin GET request return all groups with all permissions for those groups

### Review Checklist


kofalt:
Note: This will need a corresponding SDK change as well.

nagem:
The changes here LGTM! I think it's ready to merge when the clients have made the key changes as well.

kofalt:
SDK is ready; ref above link.

nagem:
This PR is already suffering from severe scope creep, but @gsfr requested one more change:
https://github.com/scitran/core/issues/814

ryansanford:
@gsfr Is reaper impacted by key change to group name/label key change?

Looks like users of the script `bin/load_users_drone_secret.py` in this repo will be impacted.

gsfr:
@ryansanford Reaper is not affected; it only uses group IDs.

bin/database.py
```diff
@@ -1111,6 +1111,23 @@ def upgrade_to_30():
...
+    else:
+        return 'Parallel failed: update doc ' + str(job['_id']) + ' resulted modified ' + str(result.modified_count)
+
+def upgrade_to_31():
+    """
+    Changes group.roles -> groups.permissions
+
+    scitran/core #662
+    """
+    cursor = config.db.groups.find({'roles': {'$exists': True}})
```
nagem: You don't always have to use a process_cursor, only when the change can't be done entirely in mongo (usually when something needs to be computed or files/analyses/other list object are involved).

Here you could do `config.db.groups.update_many({}, {'$rename': {'roles': 'permissions'}})`

kofalt:
The latest CLI has been ported to use the SDK for all underlying network :tada: 

As such, it should not be affected by the roles --> permissions rename, but I have not looked at whatever scope creep we've added to this change.

dpuccetti:
@hkethi002 The `/projects/groups` seems to be missing the label.  This endpoint is used to get the group label for each project that the user is in to allow the group label to be displayed even if the user is not in the group.

**group-permissions branch**
Request: https://docker.local.flywheel.io:8443/api/projects/groups
Response: ``` [{"_id": "scitran"}] ```

**Dev**
Request: https://dev.flywheel.io/api/projects/groups
Response: ``` [{"_id": "adni", "name": "adni"}] ```

PR #801, opened 2017-06-05
Removal of external site logic

### LINKED ISSUE #638: Remove external site access code/permission logic from core
External site access is in a semi-maintained state. Some permissions logic requires the user's `site_id` to be checked, some doesn't. We currently do not support access from external sites. This should be removed until there is a more concentrated effort to implement this feature.

Todo:
  - remove any reference to `site` on all permissions lists on all containers
  - remove references to `site` from permissions checking logic ( most of which is in `api/auth/`)
  - remove logic from `BaseHandler()` and `centralclient.py` (or maybe all of `centralclient.py`) that references the cross-site logic

hkethi002:
Fixes #638 
### Breaking Changes
- SiteId is no longer an accepted input for permissions
### Review Checklist


api/api.py
```diff
@@ -1,7 +1,7 @@
 import webapp2
 import webapp2_extras.routes
 
-from .centralclient               import CentralClient
+# from .centralclient               import CentralClient
```
nagem: You can remove this line 👍

bin/database.py
```diff
@@ -1054,6 +1054,65 @@ def upgrade_to_29():
...
+    process_cursor(cursor, upgrade_to_30_acquisitions_closure)
+
+    cursor = config.db.groups.find({'roles.site': {'$exists': True}})
+    process_cursor(cursor, upgrade_to_30_groups_closure)
+
+    cursor = config.db.projects.find({'permissions.site' : {'$exists' : True}})
+    process_cursor(cursor, upgrade_to_30_projects_closure)
+
+    cursor = config.db.sessions.find({'permissions.site': {'$exists' : True}})
+    process_cursor(cursor, upgrade_to_30_sessions_closure)
```
nagem: We should definitely rebase this branch off master when the timestamp fix lands and maybe bring in the roles->permissions rename changes as well.

raml/schemas/definitions/permission.json
```diff
@@ -16,7 +16,7 @@
     },
     "permission-output-default-required":{
```
nagem: We'll want to remove site from the `permission` schema definition.

nagem:
Only issue I saw was noted in the comment above (modifying the json schema to prevent clients from sending the site when modifying permissions lists). Unfortunately that makes this a breaking change to the API.

bin/database.py
```diff
@@ -1115,6 +1115,20 @@ def upgrade_to_31():
...
+            del permission_['site']
+    result = config.db[coll].update_one({'_id': coll_item['_id']}, {'$set': {'permissions' : permissions}})
+    if result.modified_count == 0:
+        return "Failed to remove site field"
+    return True
+
+def upgrade_to_32():
+    for coll in ['acquisitions', 'groups', 'projects', 'sessions']:
+        cursor = config.db[coll].find({'permissions.site': {'$exists': True}})
+        process_cursor(cursor, upgrade_to_32_closure, context = coll)
```
nagem: Here we can also remove the "sites" collection, as well as remove it's creation in `config.py`.

nagem: @ryansanford you never use the `sites` collection in Mongo, correct?

nagem: Currently it should only hold the current site's `id`, `name`, and `site_url` which are all available in the config.

kofalt:
> +135 −368

:tada: :tada: :tada:

PR #804, opened 2017-06-07
re-add roles and tags to group projection

### LINKED ISSUE #803: Fix always empty roles array on groups

ambrussimon-invenshure:
Closes #803 

### Background:
- the roles and the tags projection used to be set to an empty list []
- while creating the unit test infrastructure I ran into mongomock not handling the projection value []
- assumed the intent was to *not* include roles and tags, and simply removed those
- frontend bug arose because of the now missing roles key (realized pymongo *includes* when proj=[])
- re-added roles and tags to projection, now with value 1

### Review Checklist

### DIFF
--- a/api/handlers/grouphandler.py
+++ b/api/handlers/grouphandler.py
@@ -37,7 +37,7 @@ class GroupHandler(base.RequestHandler):
         return result
 
     def get_all(self, uid=None):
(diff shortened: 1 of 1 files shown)


nagem:
LGTM!

Issue #660, opened 2017-02-10
Orginary users are not allowed to `GET /users/<x>`


kofalt:
You cannot retrieve any user other than yourself, despite the response not having any sensitive fields.

PR #805, opened 2017-06-07
Actions requiring root tag made by root user don't need root tag

### LINKED ISSUE #657: Actions that require root, made by a user with root, should not require the root flag.
Example: `POST /users` will **never** work without `?root=true`!

ippkj

* [ ] `POST /users`
* [ ] More

hkethi002:
Fixes #657 and #660

Instead of checking if user is superuser, the handlers and auths for jobs, gear, groups, and users check if the user's root is set to true.

Tests were modified to use as_user so that root was not true.

_Changes Made_
- PUT, POST, and DELETE user only require admin and not root param (superuser) as well
- User can get any other user, not just themselves 
   - (Don't know if this belongs here) <https://github.com/scitran/core/issues/660>
- POST and DELETE gears only require admin
- POST, PUT, and GET next jobs only requires admin

### Review Checklist

### DIFF
--- a/api/auth/groupauth.py
+++ b/api/auth/groupauth.py
@@ -11,8 +11,6 @@ def default(handler, group=None):
                 pass
             elif handler.public_request:
                 handler.abort(400, 'public request is not valid')
-            elif config.db.users.find_one({'_id': handler.uid}).get('root'):
-                pass
             elif method in ['DELETE', 'POST']:
                 handler.abort(403, 'not allowed to perform operation')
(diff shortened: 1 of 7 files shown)


hkethi002:
The group root tag change was moved to #800 because it would be easier for Dan to make that change at the same time as the group key changes.

kofalt:
Alright, then we can go ahead with what's here. SDK is ready, but needs manual merge work from myself.

nagem:
Looks good, feel free to merge!

PR #812, opened 2017-06-08
Optional arg to force upgrade from previous version

hkethi002:
And a quick fix for upgrade 30.

### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import argparse
 import bson
 import copy
 import datetime
@@ -31,7 +32,7 @@ def get_db_version():
         return 0
     return version.get('database')
 
-def confirm_schema_match(force_from = None):
+def confirm_schema_match():
     """
     Checks version of database schema
 
@@ -45,8 +46,6 @@ def confirm_schema_match(force_from = None):
     """
 
     db_version = get_db_version()
-    if force_from:
-        db_version = int(force_from)
     if not isinstance(db_version, int) or db_version > CURRENT_DATABASE_VERSION:
         logging.error('The stored db schema version of %s is incompatible with required version %s',
                        str(db_version), CURRENT_DATABASE_VERSION)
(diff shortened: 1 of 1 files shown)


bin/database.py
```diff
@@ -1112,14 +1113,31 @@ def upgrade_to_30():
...
     Upgrades db to the current schema version
 
     Returns (0) if upgrade is successful
     """
 
     db_version = get_db_version()
+
+    if isinstance(db_version,int) and db_version >= force_from:
+        if(force_from):
+            db_version = force_from
```
nagem: I would switch these if statements around (1125 and 1126). `int >= None` functionally works, but the conditional makes more sense when written as "if you are attempting to use a force from, make sure the db_version is greater than or equal to the force_from".

nagem:
Logic looks good, the table I tested the logic with (by reading, not through running the updates) was:

|                                    | ok | ok | error |error | error |
|--------------------- |----|----|----|----|----|
| current db_version    | 29 | 30 | 30 | 30 | 31 |
| database.py version  | 30 | 30 | 30 | 30 | 30 |
| force_from                  | 28 | 29 | 30 | 31 | 30 |

bin/database.py
```diff
@@ -1067,21 +1068,21 @@ def upgrade_to_30_closure_analysis(coll_item, coll):
                 if 'created' not in file_:
                     file_['created'] = analysis_.get('created', datetime.datetime(1970, 1, 1))
         result = config.db[coll].update_one({'_id': coll_item['_id']}, {'$set': {'analyses': analyses}})
-        if result.modified_count == 1:
+        if result.modified_count == 1 or files == []:
```
nagem: I think this should read `or analyses = []`?

nagem: Or maybe this is better written as
```
if analyses:     # analyses is not None nor an empty list, therefore an update should succeed
    for analysis_ in analyses:
    ...
    if result.modified_count == 1:
        return True
    else:
        return "File timestamp ..."
else:  # Analyses is None or an empty list, move on ...
   return True
```

nagem: Specifically, Ryan is seeing this error on dev which is probably caused by an empty list of analyses:
```
  File "/var/scitran/code/api/bin/database.py", line 1146, in upgrade_schema
    globals()[upgrade_script]()
  File "/var/scitran/code/api/bin/database.py", line 1101, in upgrade_to_30
    process_cursor(cursor, upgrade_to_30_closure_analysis, context = 'sessions')
  File "/var/scitran/code/api/bin/database.py", line 117, in process_cursor
    result = closure(document, context)
  File "/var/scitran/code/api/bin/database.py", line 1071, in upgrade_to_30_closure_analysis
    if result.modified_count == 1 or files == []:
UnboundLocalError: local variable 'files' referenced before assignment
```

PR #819, opened 2017-06-15
Subject name hashes removed

### LINKED ISSUE #813: Remove firstname_hash and lastname_hash from subject
These fields are not currently used (just set by the reaper) and cause issues with languages that cannot handle underscores in keynames.

TODO:
  - [ ] DB upgrade to remove all references to these two keys
  - [ ] Update subject schemas to reject these keys
  - [ ] Remove any reference to these keys from projections/other code in the api
  - [ ] Update/write relevant tests

hkethi002:
Fixes #813 
### Review Checklist


nagem:
LGTM!

nagem:
We should make sure the classifier gear does not send this information before merging, although I'm pretty sure it doesn't.

PR #820, opened 2017-06-15
analysis downloads use label istead of id

### LINKED ISSUE #609: Rename Analysis Download
Name not very user-friendly. Should reflect analysis name/label:
screen shot 2016-12-05 at 3 02 51 pm

hkethi002:
Fixes #609 
### Review Checklist

### DIFF
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -226,6 +226,10 @@ class AnalysesStorage(ListStorage):
         else:
             return files
 
+    def get_analysis_label(self, _id, analysis_id):
+        _id = bson.ObjectId(_id)
+        return self._get_el(_id, {'_id': analysis_id}).get('label', 'No Label')
+
     def add_note(self, _id, analysis_id, payload):
         _id = bson.ObjectId(_id)
         query = {
--- a/api/dao/liststorage.py
+++ b/api/dao/liststorage.py
@@ -226,9 +226,9 @@ class AnalysesStorage(ListStorage):
         else:
             return files
 
-    def get_analysis_label(self, _id, analysis_id):
+    def get_analysis(self, _id, analysis_id):
         _id = bson.ObjectId(_id)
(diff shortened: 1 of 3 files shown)


api/dao/liststorage.py
```diff
@@ -226,6 +226,10 @@ def get_fileinfo(self, _id, analysis_id, filename = None):
         else:
             return files
 
+    def get_analysis_label(self, _id, analysis_id):
+        _id = bson.ObjectId(_id)
+        return self._get_el(_id, {'_id': analysis_id}).get('label', 'No Label')
+
```
nagem: As a general rule, we like to have storage classes return the document as close to the original state as possible and let the handler classes, which are much more specialized, handle those null cases. If there was a different handler class that would act differently when there is no label, it would wouldn't be able to tell from what gets returned from AnalysesStorage (someone could name their analysis "No Label", for example).

Sidenote: we do sometimes fill in missing fields with acceptable null values (like None, [] or {}) from the Storage classes, but we assume in this case a missing field and a false-y field are the same db state.

nagem: tl;dr, move this check to the handler

api/handlers/listhandler.py
```diff
@@ -797,7 +797,7 @@ def download(self, cont_name, list_name, **kwargs):
                 ticket = util.download_ticket(self.request.client_addr, 'file', _id, filename, total_size, origin=self.origin)
             else:
                 targets, total_size, file_cnt = self._prepare_batch(fileinfo)
-                filename = 'analysis_' + analysis_id + '.tar'
+                filename = 'analysis_' + storage.get_analysis_label(_id, analysis_id) + '.tar'
```
nagem: We'll want to make sure the file name we create is a reasonable file name that file systems can handle. I think we have a util function that strips them of all special characters already, I'll take a look.

nagem:
LGTM, thanks for making the change. Feel free to merge this one.

Issue #775, opened 2017-05-05
If job creation fails during analysis creation, the analysis persists


nagem:
We currently add the analysis to it's container in the db, create the job doc in the db, and then attach them. This method leaves an unusable analysis when the job creation fails. There are several ways to fix this:
  - clean up the analysis if the job creation fails. This would create a situation where a GET on the analysis's container might return this container during the short interval between anaysis creation and job attempted creation, but is a rare and minimally problematic race condition. (Also, this is possible now).
  - Generate an id for the analysis, add the job (which needs the analysis id for its destination key), and then add the analysis. If the analysis addition fails, we'll be in a similar but opposite situation - except failure is more problematic (engine picks up job before it is remove, etc).
  - Some kind of overengineered batch update to ensure there is minimal chance for race condition.

If anyone has another idea, feel free to suggest. It's a pretty classic circular reference problem.

kofalt:
I'm not super worried by a race condition because we accept those as part of our choice of database. I'm a fan of your first approach - attempt to clean up dependent documents if sibling creation fails.

We could also add a new queue state `hold` which will not be consumed, are fully mutable, can be deleted (!), and can only transition to `pending` or `cancelled`. That would be a clean approach to your second option: the job sits forever until & unless it is complete and ready to be consumed.

Also: heh, I just realized I misspelled that state when I wrote the job queue. Nice :sunglasses: :+1:

nagem:
I like the addition of the `hold` state - I think there are other applications for it than just this one, too.

I think to resolve this current issue I'll do step #1 and reference this ticket (but not close it) and we can revisit if there is a larger jobs enhancement is the works (or when we break analyses into their own mongo connection).

kofalt:
SGTM.

PR #829, opened 2017-06-27
Analyses collection

### LINKED ISSUE #808: Analyses should be their own doc in mongodb
Stepping stone to ease analyses interaction with jobs and search.
@nagem add other benefits here as well.

Phase 1:
- [ ] No functional changes to API
  - jobs & batch changes would be significant.
- [x] Migrate existing analyses out containers into their own collection using `bin/database.py`.
  - sessions
  - collections
  - @nagem Any other containers supported that need to be accounted for?
- [x] Add appropriate index for efficient lookup/join
- [ ] Add tests for missing coverage
- [ ] Update search replication code in dependent 

Followup work would be proper support in jobs, batch and resolver for analyses within any supporting container.

ambrussimon-invenshure:
Closes #808 

Important notes:
- added db upgrade and indexes
- AnalysisStorage is now subclassing ContainerStorage
- AnalysisHandler is a custom handler (not a ListHandler anymore)
- analysis notes are handled via NotesListHandler
- analysis files are still handled via the custom method (partly redundant w/ FileListHandler - separating the logic would take some time)
- changed DELETE analysis output schema! {modified: 1} -> {deleted: 1}
- analyses are not persisted if job=true and job couldn't be created #775 
- abao revealed that analyses used to be on "GET /{Container}" - it is my understanding they shouldn't be, and now there are not even analysis _id's (no analyses attr whatsoever) upon container get_all
- moving AnalysisStorage into containerstorage.py introduced an import cycle (gear validation), which is currently "resolved" w/ delayed imports in AnalysisStorage methods
- huge diff, sorry for any merge inconvenience later...

### Review Checklist


api/dao/containerstorage.py
```diff
@@ -429,3 +435,162 @@ def get_all_for_targets(self, target_type, target_ids,
...
+
+            for i in getattr(job, 'inputs',{}):
+                fileref = job.inputs[i]
+                contref = containerutil.create_containerreference_from_filereference(job.inputs[i])
+                file_ = contref.find_file(fileref.name)
+                if file_:
+                    file_['input'] = True
+                    files.append(file_)
+
+            q = {'analyses._id': analysis['_id']}
```
ambrussimon-invenshure: This must be wrong, I will look into adding a test to cover this tomorrow.

raml/schemas/mongo/analysis.json
```diff
@@ -4,6 +4,13 @@
     "type": "object",
     "properties": {
     	"_id":          {"type": "string"},
+        "parent":       {
+                            "type": "object",
+                            "properties": {
+                                "type": "string",
+                                "id":   "string"
+                            }
+                        },
```
nagem: It would also need `permissions`, `public` and `archived`, correct? Are we even using the mongo schemas for analysis? These have kind of fallen out of maintenance as new development didn't use them.

api/api.py
```diff
@@ -250,20 +248,27 @@ def prefix(path, routes):
...
-                route('/<list_name:notes>/<_id:{nid}>',                          NotesListHandler, name='notes',   m=['GET', 'PUT', 'DELETE']),
+                route('/<list_name:notes>',             NotesListHandler,               m=['POST']),
+                route('/<list_name:notes>/<_id:{nid}>', NotesListHandler, name='notes', m=['GET', 'PUT', 'DELETE']),
             ])
         ]),
 
 
+        # Analysis notes
+
+        prefix('/<:{cname}>/<:{cid}>/<cont_name:analyses>/<cid:{cid}>', [
```
nagem: 👍

api/dao/containerstorage.py
```diff
@@ -429,3 +435,162 @@ def get_all_for_targets(self, target_type, target_ids,
...
+        parent_type = containerutil.singularize(parent_type)
+        parent_id = bson.objectid.ObjectId(parent_id)
+        analyses = self.get_all_el({'parent.type': parent_type, 'parent.id': parent_id}, None, None)
+        if inflate_job_info:
+            for analysis in analyses:
+                self.inflate_job_info(analysis)
+        return analyses
+
+
+    def get_fileinfo(self, _id, filename=None):
```
nagem: I wonder why there was a specific method created for this rather than just getting the analysis and reading it's file list. It's only used in one location so I would be in favor of removing it and moving the parts of it we need to where it is called a la `AnalysisStorage.get_container(<>).get('files', [])` or if that's too wordy writing a generic `get_files_for_container()` or something on `ContainerStorage`.

api/dao/containerstorage.py
```diff
@@ -429,3 +435,162 @@ def get_all_for_targets(self, target_type, target_ids,
...
+                'id': bson.objectid.ObjectId(cid)
+            },
+            '_id': bson.objectid.ObjectId(),
+            'created': datetime.datetime.utcnow(),
+            'modified': datetime.datetime.utcnow(),
+            'user': origin.get('id'),
+            'permissions': parent['permissions'],
+            'public': parent.get('public', False),
+        }
+        for key in defaults:
```
nagem: I'm curious about why you chose to use `setdefault` rather than creating a default and then calling `return default.update(analysis)`? Just a style choice? I just very rarely use `setdefault()`.

api/dao/containerstorage.py
```diff
@@ -429,3 +435,162 @@ def get_all_for_targets(self, target_type, target_ids,
...
+
+
+    def create_job_and_analysis(self, cont_name, cid, analysis, job, origin):
+        """
+        Create and insert job and analysis.
+        """
+
+        # TODO fix import cycle - separate analysisstorage module?
+        from ..jobs.gears import validate_gear_config, get_gear
+        from ..jobs.jobs import Job
```
nagem: An idea would be to move the job creation and gear validation to the handler. You would need to update the created job object here before calling save but it would solve the problem ¯\\\_(ツ)\_/¯

nagem: You'd also have to move the `inflate_job_info` stuff below too. I'll think on this a little longer.

api/dao/containerutil.py
```diff
@@ -192,3 +193,18 @@ def create_containerreference_from_dictionary(d):
 
 def create_containerreference_from_filereference(fr):
     return ContainerReference.from_filereference(fr)
+
+
+def pluralize(cont_name):
```
nagem: Like I've mentioned before, these are great. A much better implementation than the makeshift functions we have scattered around the products.

api/handlers/containerhandler.py
```diff
@@ -240,7 +232,7 @@ def get_jobs(self, cid):
 
         permchecker(noop)('GET', cid)
 
-        analyses = cont.get('analyses', [])
+        analyses = list(config.db.analyses.find({'parent.type': 'session', 'parent.id': cont['_id']}).sort('created', -1))
```
nagem: Can we use `AnalysisStorage().get_analyses(<>)` here like above? It doesn't always happen, but I try to not directly make db calls from container handlers because then we can put calls that should always happen in the `storage` layer and feel relatively confident they will always be called. Things like compliance calculation, default field filling, etc.

api/handlers/refererhandler.py
```diff
@@ -0,0 +1,353 @@
...
+            return {'deleted': result.deleted_count}
+        else:
+            self.abort(404, 'Analysis {} not removed from container {} {}'.format(_id, cont_name, cid))
+
+
+    def download(self, cont_name, cid, _id, filename=None):
+        """
+        .. http:get:: /api/(cont_name)/(cid)/analyses/(analysis_id)/files/(file_name)
+
+            Download a file from an analysis or download a tar of all files
```
nagem: Feel free to trim this down. We no longer use Sphynx doc generation that this was written for.

nagem: Although, it is very thorough for an overly complex function.

nagem:
@ambrussimon-invenshure looks good! I gave it once-over, but I'll probably be back tomorrow to look over it again. Still thinking about that circular reference problem ..

nagem:
Also I'd like @hkethi002 to take a look at this when he gets a chance as well.

ambrussimon-invenshure:
@nagem thanks for catching those, updated PR accordingly! (The only one left is the ugly late import.)

api/dao/containerstorage.py
ambrussimon-invenshure: dict.update() returns None - one would have to `return default` after that, which is (subjectively, and only slightly) odd. Happy to switch if this doesn't convince you, I find both to be fairly pythonic :)

nagem:
All the changes look good 👍.

We'll have to hold off on merging until our search infra moves over to the new analysis location in mongo.

ambrussimon-invenshure:
@nagem merged master into feature, resolved conflicts and made sure everything clicks. In the process I found the flaky time-dependent test reared it's head again, PR'd that separately (that's why there's a CI failure on this branch.)

PR #836, opened 2017-07-05
Analysis downloads can now use the general container endpoints

### LINKED ISSUE #831: All steps of Analysis download should go through download.py
As of now, the creation of the ticket for the download uses the endpoint "sessions/<session_id>/analyses/<analysis_id>/files" and the actual downloading of the files uses "download?ticket=<ticket_id>". The ticket creation should also go through download.py as well.

hkethi002:
Fixes #831 
### Breaking Changes
- Frontend can switch the analysis downloads to use the same endpoints the other containers use (not really breaking)
- ```/download``` for bulk analysis download
- ```/analyses/{AnalysisId}/files/{Filename}``` for single file download
### Review Checklist

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -248,7 +248,7 @@ endpoints = [
                 route( '/analyses',                                AnalysesHandler,                  m=['POST']),
                 prefix('/analyses', [
                     route('/<_id:{cid}>',                          AnalysesHandler,                  m=['GET', 'DELETE']),
-                    route('/<_id:{cid}>/files',                    AnalysesHandler, h='download',    m=['GET']),
+                    route('/<_id:{cid}>/files',                    Download,        h='download',    m=['GET']),
                     route('/<_id:{cid}>/files/<filename:{fname}>', AnalysesHandler, h='download',    m=['GET']),
                 ]),
 
--- a/api/api.py
+++ b/api/api.py
@@ -248,7 +248,7 @@ endpoints = [
                 route( '/analyses',                                AnalysesHandler,                  m=['POST']),
                 prefix('/analyses', [
                     route('/<_id:{cid}>',                          AnalysesHandler,                  m=['GET', 'DELETE']),
-                    route('/<_id:{cid}>/files',                    Download,        h='download',    m=['GET']),
+                    route('/<_id:{cid}>/files',                    AnalysesHandler, h='download',    m=['GET']),
                     route('/<_id:{cid}>/files/<filename:{fname}>', AnalysesHandler, h='download',    m=['GET']),
                 ]),
 
--- a/api/api.py
+++ b/api/api.py
@@ -35,7 +35,7 @@ routing_regexes = {
     'uid': '[0-9a-zA-Z.@_-]*',
 
     # Container name
-    'cname': 'projects|sessions|acquisitions|collections',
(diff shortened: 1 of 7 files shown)


api/dao/containerutil.py
```diff
@@ -200,11 +201,11 @@ def pluralize(cont_name):
         return SINGULAR_TO_PLURAL[cont_name]
     elif cont_name in PLURAL_TO_SINGULAR:
         return cont_name
-    raise Exception('Could not pluralize unknown container name {}'.format(cont_name))
+    raise InputValidationException('Could not pluralize unknown container name {}'.format(cont_name))
```
nagem: I would rather we keep this as an exception and try/catch this in the handler layer. If code gets to containerutil.py (or anything in `api.dao`) we expect it has already been validated*. The handler can then throw InputValidationException. 

*This might not be 100% true ATM but is a goal I'd like to achieve

api/download.py
```diff
@@ -84,12 +86,12 @@ def _bulk_preflight_archivestream(self, file_refs):
         targets = []
 
         for fref in file_refs:
-            cont_name   = fref.get('container_name','')+'s'
+            cont_name   = pluralize(fref.get('container_name',''))
```
nagem: This would be the try/catch location (or modify the if statement below to be singular and check first)

api/download.py
```diff
@@ -84,12 +86,12 @@ def _bulk_preflight_archivestream(self, file_refs):
...
             cont_id     = fref.get('container_id', '')
             filename    = fref.get('filename', '')
 
-            if cont_name not in ['projects', 'sessions', 'acquisitions']:
-                self.abort(400, 'Bulk download only supports files in projects, sessions and acquisitions')
+            if cont_name not in ['projects', 'sessions', 'acquisitions', 'analyses']:
+                self.abort(400, 'Bulk download only supports files in projects, sessions, analyses and acquisitions')
```
nagem: Although this might be unreachable because of the json schema check...

nagem:
LGTM with some minor changes.

api/api.py
```diff
@@ -35,7 +35,7 @@
     'uid': '[0-9a-zA-Z.@_-]*',
 
     # Container name
-    'cname': 'projects|sessions|acquisitions|collections',
+    'cname': 'projects|sessions|acquisitions|collections|analyses',
```
ambrussimon-invenshure: I think this now also allows for the removal of [250-251](https://github.com/scitran/core/pull/836/files#diff-e0978ea022e5e946d30dfa3911881b5eR250) and [260-267](https://github.com/scitran/core/pull/836/files#diff-e0978ea022e5e946d30dfa3911881b5eR260), cool! (Also refererhandler.py#116-)

nagem:
@hkethi002 Removing the code @ambrussimon-invenshure mentioned and returning the failure to an exception with a try/catch in the handler are the final changes. After those are done, this PR can start the merge party!

api/api.py
ambrussimon-invenshure: It just occurred to me now, that removal would make this a breaking change, forcing UI to switch immediately. Deferring judgement.

hkethi002: I've added this change as issue https://github.com/scitran/core/issues/856  for later because it did not seem to be non-trivial

PR #839, opened 2017-07-06
New endpoint for string field type ahead

### LINKED ISSUE #838: Add endpoint to support type-ahead on custom search strings
Add the endpoint 

```
POST api/dataexplorer/search/fields/aggregate
{
    "field_name": "file.info.something",
    "value": "something a user is typing"
}
```

That creates an elasticsearch aggregation like so (with a filter added):
```
{
    "size": 0,
    "query": {
                "match" : { "file.info.something" : "something the user typed in" }
            },
    "aggs" : {
        "results" : {
               "terms" : {
                        "field" : "file.info.something.raw",
                        "size" : 15
                    }
        }
	}
}
```

And modify the result set to return a list of options so a client can implement typeahead match on values stored in the field referenced in `field_name` that match `value`.

hkethi002:
Closes #838 
### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -252,90 +252,85 @@ class DataExplorerHandler(base.RequestHandler):
         return return_type, modified_filters, search_string
 
     @require_login
-    def get_type_ahead_values(self):
+    def custom_field_values(self):
         """
         Return list of type ahead values for a key given a value 
         that the user has already started to type in for the value of 
-        a custom string field
+        a custom string field or a set of statistics if the field type is
+        a number.
         """
-        user_value = self.request.json_body['value']
+        
         custom_field = self.request.json_body['field_name']
-        modified_filters = []
-        # Add permissions filter to list if user is not requesting all data
-        if not self.request.json_body.get('all_data', False):
-            modified_filters.append({'term': {'permissions._id': self.uid}})
+        filters = [{'term': {'permissions._id': self.uid}}]
+        field = config.es.indices.get_field_mapping(custom_field,
+                                                    index='data_explorer',
+                                                    doc_type='flywheel')['data_explorer']['mappings']['flywheel'][custom_field]
+        field_type = self._get_field_type(field['mapping'][custom_field.split('.')[-1]]['type'])
 
-        body = {
-            "size": 0, 
-            "query": {
-                "bool": {
-                    "must" : {
-                        "match" : { custom_field : user_value}
-                    },
-                    "filter" : {
-                        "bool" : {
-                            "must" : modified_filters
-                        }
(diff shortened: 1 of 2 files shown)


api/handlers/dataexplorerhandler.py
```diff
@@ -251,6 +251,91 @@ def _parse_request(self, request_type='search'):
...
+        Return list of type ahead values for a key given a value 
+        that the user has already started to type in for the value of 
+        a custom string field
+        """
+        user_value = self.request.json_body['value']
+        custom_field = self.request.json_body['field_name']
+        modified_filters = []
+        # Add permissions filter to list if user is not requesting all data
+        if not self.request.json_body.get('all_data', False):
+            modified_filters.append({'term': {'permissions._id': self.uid}})
```
nagem: I think we would want this filter applied all the time - a user might put something they'd rather no one else see in a custom field. Retrieving all results does't allow a user to see more than a few descriptive fields of results.

api/handlers/dataexplorerhandler.py
```diff
@@ -251,6 +251,91 @@ def _parse_request(self, request_type='search'):
...
+        aggs = config.es.search(
+            index='data_explorer',
+            doc_type='flywheel',
+            body=body
+        )['aggregations']['results']['buckets']
+        aggs = [bucket['key'] for bucket in aggs]
+        return {'type_aheads': aggs}
+
+    def get_type_ahead_numbers(self):
+        # user_value = self.request.json_body['value']
```
nagem: If we add a lookup to the data_explorer_fields index to discover the type, we could just have one function and one endpoint, `aggregate`. It's also not really a type-ahead for numbers - it's more of a stats endpoint for a specific field, the user would not be providing a search string. 

I think a good interface for this endpoint would be:
```
{
    'search_string': '<the search string provided by the user, used for type ahead on string values>', #Not required
    'field_name': '<full name of field>"
}
```

And then a lookup to see what type the field name is, using a stats agg if `integer`, `float`, or `date`, and a terms agg if `string`. If `string`, and no `search_string`, remove the filter on the `search_string` from the query.

api/handlers/dataexplorerhandler.py
```diff
@@ -251,6 +251,91 @@ def _parse_request(self, request_type='search'):
...
+        body = {
+            "size": 0, 
+            "query": {
+                "bool": {
+                    "must" : {
+                        "match" : { custom_field : user_value}
+                    },
+                    "filter" : {
+                        "bool" : {
+                            "must" : modified_filters
```
nagem: If the user permission is not added to the filter, we wouldn't want an empty `must` clause.

nagem:
Changes merged into https://github.com/scitran/core/pull/846

kofalt:
Remove branch? Y/N

PR #840, opened 2017-07-06
Update api console

### LINKED ISSUE #828: Update api-console for anchor support
Upcoming version 4 of API console will support anchoring. This makes it easier to point collaborators at relevant API documentation.
https://github.com/mulesoft/api-console/issues/60

Github served docs for this repo are at https://github.com/scitran/core/blob/master/docs/api/v1/index.html

and hosted via: https://scitran.github.io/core/api/v1/index.html?branch=master

ambrussimon-invenshure:
Closes #828 

### Notes

* The generated v4 console looks/feels very different!
* There used to be a top level index.html redirecting to /v1 - this kind of versioning is not implemented now
* Didn't implement loading animation either
* Added readme including instructions for viewing api console using local html/js/etc files (raml always loaded from github)

### Review Checklist


docs/src/assets/examples/array.raml
```diff
@@ -0,0 +1,58 @@
+#%RAML 1.0
+#baseUri: http://www.tshirt.com/api
```
nagem: Is it possible to slim down some of this? We shouldn't need the examples committed to the repo, and we might be able to cut a few other things out as well.

nagem:
@ryansanford  and I talked about the possibility of delayed generation of these docs so they don't have to live statically in the repo. He's going to look into it and share his thoughts.

docs/src/assets/examples/array.raml
ambrussimon-invenshure: Didn't find any [build options](https://www.npmjs.com/package/api-console-builder#options) for keeping it more minimal. Can go through generated items manually and write a list of files/folders that can be deleted, but that seems painful, error-prone (maybe something gets deleted that leads to an error only way later) and wouldn't scale well with api-console version bumps. Though I expect upgrades on our side to be very rare.

ryansanford:
I don't believe `src` or `bower_components` are required for hosting. Am I missing something obvious @ambrussimon-invenshure ?

https://ryansanford.github.io/project1/?branch=classification#/docs/summary
https://github.com/ryansanford/project1/tree/master/docs

ambrussimon-invenshure:
Just wow... I don't think you are missing anything, apparently the _standalone build_ artifact is the fat `index.html` and nothing more. I'm embarrassed I didn't find that.

ambrussimon-invenshure:
@ryansanford @nagem Rebased and removed clutter.

PR #842, opened 2017-07-07
Batch states now include running, complete, and failed

### LINKED ISSUE #710: Set of batch states != set of job states
Two different enums :pensive: 

Renaming batch `launched` to `running` would solve it.

Arguably, batches could also get `complete` and `failed`, but... eh.

hkethi002:
Fixes #710 

Batch is complete when all jobs are in a final (stable) state and is failed if all jobs are in the failed state.
### Review Checklist

### DIFF
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -16,8 +16,8 @@ log = config.log
 
 BATCH_JOB_TRANSITIONS = {
     # To  <-------  #From
-    'launched':     'pending',
-    'cancelled':    'launched'
+    'running':     'pending',
+    'cancelled':    'running'
 }
 
 
@@ -184,7 +184,7 @@ def run(batch_job):
         jobs.append(job)
         job_ids.append(job_id)
 
-    update(batch_job['_id'], {'state': 'launched', 'jobs': job_ids})
+    update(batch_job['_id'], {'state': 'running', 'jobs': job_ids})
     return jobs
 
 def cancel(batch_job):
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -218,9 +218,9 @@ def check_state(batch_id):
     batch = get(str(batch_id))
 
     if batch.get('state') == 'cancelled':
-        return False
-    batch_jobs = config.db.jobs.find({'_id':{'$in': batch.get('jobs')}, 'state': {'$nin': ['complete', 'failed']}})
-    non_failed_batch_jobs = config.db.jobs.find({'_id':{'$in': batch.get('jobs')}, 'state': {'$ne': 'failed'}})
+        return None
+    batch_jobs = config.db.jobs.find({'_id':{'$in': batch.get('jobs', [])}, 'state': {'$nin': ['complete', 'failed', 'cancelled']}})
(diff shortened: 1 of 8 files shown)


api/jobs/batch.py
```diff
@@ -161,7 +163,8 @@ def run(batch_job):
                 'config':   config_,
                 'gear_id':  gear_id,
                 'inputs':   inputs,
-                'tags':     tags
+                'tags':     tags,
+                'batch':    "batch is" + str(batch_job.get('_id'))
```
nagem: Looks like some testing code made it's way into the PR.

test/integration_tests/python/test_batch.py
```diff
@@ -126,3 +128,59 @@ def test_batch(data_builder, as_user, as_admin, as_root):
...
+        r = as_root.put('/jobs/' + job, json={'state': 'running'})
+        r = as_root.put('/jobs/' + job, json={'state': 'complete'})
+        assert r.ok
+
+    # test batch is complete
+    r = as_admin.get('/batch/' + batch_id)
+    assert r.json()['state'] == 'complete'
+
+    # Test batch failed
+    # create a batch w/ acquisition target and target_context
```
nagem: Could we also add a test for an analysis target?

api/jobs/batch.py
```diff
@@ -206,6 +209,25 @@ def cancel(batch_job):
     update(batch_job['_id'], {'state': 'cancelled'})
     return cancelled_jobs
 
+def check_state(batch_id):
+    """
+    Return True if all jobs are complete or failed
+    """    
+
+    batch = config.db.batch.find_one({'_id': batch_id})
```
nagem: Mind using the local get method instead of accessing the db directly?

api/jobs/batch.py
```diff
@@ -206,6 +209,25 @@ def cancel(batch_job):
     update(batch_job['_id'], {'state': 'cancelled'})
     return cancelled_jobs
 
+def check_state(batch_id):
+    """
+    Return True if all jobs are complete or failed
```
nagem: The description mentions it's a True/False response but it looks like it returns a state below.

api/jobs/batch.py
```diff
@@ -206,6 +209,25 @@ def cancel(batch_job):
...
+    """    
+
+    batch = config.db.batch.find_one({'_id': batch_id})
+
+    if batch.get('state') == 'cancelled':
+        return False
```
nagem: Maybe return "None" here, it makes more sense with the string responses.

kofalt:
Presumably this will need a DB upgrade, yeah?

nagem:
Oh yeah, nice catch @kofalt.

@hkethi002 we could import the function you use to determine status and just iterate over the batch items.

nagem:
Changes look good.

api/jobs/queue.py
```diff
@@ -72,7 +72,6 @@ def mutate(job, mutation):
         if result.modified_count != 1:
             raise Exception('Job modification not saved')
 
-        # If the job did not succeed, check to see if job should be retried.
```
kofalt: Why was this comment removed?

hkethi002: That was probably removed by mistake, I'll that back in, thanks for catching it!

kofalt:
Requires SDK signoff

PR #845, opened 2017-07-10
Empty group put request 500s

### LINKED ISSUE #843: Empty group / other-noun PUTs respond 200
If you send off `{}` to `PUT /groups`, for example, it will respond 200.

I would argue that this only makes sense for jobs, where the engine is explicitly heartbeating the noun.

hkethi002:
Fixes #843 
### Review Checklist

### DIFF
--- a/api/handlers/collectionshandler.py
+++ b/api/handlers/collectionshandler.py
@@ -57,6 +57,8 @@ class CollectionsHandler(ContainerHandler):
         mongo_validator, payload_validator = self._get_validators()
 
         payload = self.request.json_body or {}
+        if not payload:
+            self.abort(400, 'Bad Request')
         contents = payload.pop('contents', None)
         payload_validator(payload, 'PUT')
         permchecker = self._get_permchecker(container=container)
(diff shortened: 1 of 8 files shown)


nagem:
An empty payload is an improper request, so the more accurate response code would be `400: Bad Request`. I would add a check to group/project/session/acquisition/collection/user PUT endpoints that checks to make sure the payload is not empty as soon as possible in the handler logic.

nagem:
It should really be on all PUT endpoints, feel free to add if you'd like, otherwise that list above is a good start.

api/handlers/collectionshandler.py
```diff
@@ -57,6 +57,8 @@ def put(self, **kwargs):
         mongo_validator, payload_validator = self._get_validators()
 
         payload = self.request.json_body or {}
+        if not payload:
+            self.abort(400, 'Bad Request')
```
nagem: "Bad Request" is the title of the 400 error, you can put something more descriptive here. "PUT request body cannot be empty" or something.

nagem:
Looks good after response message is changed to be a bit more descriptive!

nagem:
LGTM!

kofalt:
Thanks Harsha.

PR #851, opened 2017-07-12
Remove redundant key from ContainerReference

kofalt:
It looks like this key slipped in as part of 3f6cc65457309b4e736a960e24999e69c216bb77, and it's making its way to the database.
This isn't great because it's the same data as the `type` key, isn't shipped by the SDK, etc.

I did a quick check and the key isn't used anywhere but its own class, so it seems safe to remove.
Could be accompanied by a database migration if you reeeeally want one.

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -116,7 +116,6 @@ class ContainerReference(object):
             raise Exception('Container id must be of type str')
 
         self.type = type
-        self.collection = pluralize(type)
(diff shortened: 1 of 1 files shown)


nagem:
LGTM

ambrussimon-invenshure:
Thanks for fixing! Didn't realize at the time that they get stored, and was hasty with getting rid of the builtin override. LGTM

PR #854, opened 2017-07-12
Descriptive error when search is down

### LINKED ISSUE #852: Better error message when elastic connection is unsuccessful
If elastic is down:

```
POST api/dataexplorer/search return_type="file"
```

```json
{
    "message": "ConnectionError(<urllib3.connection.HTTPConnection object at 0x7fdc00428810>: Failed to establish a new connection: [Errno -2] Name or service not known) caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x7fdc00428810>: Failed to establish a new connection: [Errno -2] Name or service not known)", 
    "status_code": 500
}
```

Ideally, it would be something more like this:

```js
{
    "message": "Search is currently down. Try again later.",
    "status_code": 503 // <-- note
}
```

hkethi002:
Fixes #852 
### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -334,12 +334,14 @@ class DataExplorerHandler(base.RequestHandler):
         else:
             self.abort(400, 'Aggregations are only allowed on string, integer, float, data and boolean fields.')
 
-        aggs = config.es.search(
-            index='data_explorer',
-            doc_type='flywheel',
-            body=body
-        )['aggregations']['results']
-
+        try:
+            aggs = config.es.search(
+                index='data_explorer',
+                doc_type='flywheel',
+                body=body
+            )['aggregations']['results']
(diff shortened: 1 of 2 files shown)


nagem:
It would most likely be a `elasticsearch.TransportError`, but a catch-all `elasticsearch.ElasticsearchError` is fine. LGTM.

kofalt:
Question: What would you think of instead adding `ElasticsearchException` (or something more specific) to the global exception handler instead? This method looks brittle in that it depends on a duplicated try/catch with each ES interaction.

nagem:
👍 changes look good

kofalt:
LGTM

kofalt:
Mild retcon - I guess this swallows any ElasticSearch root-cause stack trace, right?

Maybe we want to log those, so that if we get weird ES problems we at least have debug info.

PR #866, opened 2017-07-18
Subject name added to  projection for resolver

hkethi002:
### Review Checklist


api/resolver.py
```diff
@@ -18,6 +18,8 @@ class Node(object):
         'subject.race':      0,
         'subject.ethnicity': 0,
         'subject.info':      0,
+        'subject.firstname': 0,
+        'subject.lastname':  0,
```
nagem: We also added the global `info` and `tags` keys in the lost commit, mind adding those?

nagem:
LGTM!

PR #867, opened 2017-07-18
Change rule logic to use before/after container changes

### LINKED ISSUE #863: Add container-has-measurement rule
Will match when the container has a file that has the specified measurement.

nagem:
Closes #863

Adds a new rule type `container.has-measurement` that, like the existing type `container.has-type`, will "match" when the container has a file (including the one being evaluated) with the given measurement string. Direct match only, wildcards are not currently supported. 

Example of rule using new rule type:
```
{
    "alg": "qa-report-fmri",
    "name": "Run qa-report-fmri on nifti",
    "all": [
        { "type": "file.type", "value": "nifti" }
    ],
    "any": [
        { "type": "container.has-measurement", "value": "functional" },
        { "type": "container.has-measurement", "value": "anatomical" }
    ]
}
``` 

This PR also changes the behavior of rule evaluation after a file is added or modified via one of the placers. 
  - Before, new jobs spawned via rules would only be triggered for a file when it was added. Using that logic, any rule that referenced the container would only be triggered if the container was in the proper state when the file was added. For example, if a nifti is added the dicom must already be classified and have a measurement, or this rule would not trigger. Later, when the dicom's measurement is set, the nifti has already had its one and only rule evaluation during upload so the job would never trigger. 

  - Now, when files are added or edited via one of the placers, the container is evaluated as a whole before and after the change. The list of jobs that would be spawned before and after is compared, and only those in the after set are spawned, which solved the problem noted above. 

### Breaking Changes
New rule type: `container.has-measurement`

### Review Checklist

### DIFF
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -117,6 +117,12 @@ class ContainerReference(object):
         self.type = type
         self.id = id
 
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
+        return not self.__dict__ == other.__dict__
+
     @classmethod
     def from_dictionary(cls, d):
         return cls(
@@ -175,6 +181,12 @@ class FileReference(ContainerReference):
     def __init__(self, type, id, name):
         super(FileReference, self).__init__(type, id)
         self.name = name
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    def __ne__(self, other):
+        return not self.__dict__ == other.__dict__
 
     @classmethod
     def from_dictionary(cls, d):
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -261,38 +261,27 @@ def is_session_compliant(session, template):
     return True
 
 def upsert_fileinfo(cont_name, _id, fileinfo):
-    # TODO: make all functions take singular noun
-    cont_name += 's'
 
-    # TODO: make all functions consume strings
+    cont_name = containerutil.pluralize(cont_name)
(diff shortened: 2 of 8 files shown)


kofalt:
Tested 4a44063 with several clean runs, found:

1. Switching to branch changed nothing (good)
1. Changing the specified rule removed most failures (good)

Raw data:
image

Applied the following rule change for the third trial:

```diff
                        ]
                },
                {
-                       "rule": {
-                               "alg": "qa-report-fmri",
-                               "name": "Run qa-report-fmri on nifti",
-                               "all": [
-                                       { "type": "file.type", "value": "nifti" }
-                               ],
-                               "any": []
-                       },
-                       "projects": [
-                               "Neuroscience", "Psychology", "Testdata"
-                       ]
+                   "rule": {
+                       "alg": "qa-report-fmri",
+                       "name": "Run qa-report-fmri on nifti",
+                       "all": [
+                           { "type": "file.type", "value": "nifti" }
+                       ],
+                       "any": [
+                           { "type": "container.has-measurement", "value": "functional" },
+                           { "type": "container.has-measurement", "value": "anatomical" }
+                       ]
+                   },
+                   "projects": [
+                       "Neuroscience", "Psychology", "Testdata"
+                   ]
                }
        ],
```

ryansanford:
Tested out great for me. 👍

ryansanford:
@nagem Do we need to adjust any db indexes to avoid costly db scans with jobs or acquisition mongo collections?

nagem:
@ryansanford  no db scans as part of this logic. The only db interaction is getting the before and after of the container where we are adding the file(s), which is by `_id`.

PR #868, opened 2017-07-19
Analysis PUT method for labels

### LINKED ISSUE #331: Allow editing of analyses labels
Currently `PUTs` on analyses are not allowed.

hkethi002:
Fixes #331 

### Review Checklist

### DIFF
--- a/api/handlers/refererhandler.py
+++ b/api/handlers/refererhandler.py
@@ -8,6 +8,7 @@ container (eg. ListHandler)
 
 import os
 import zipfile
+import datetime
 from abc import ABCMeta, abstractproperty
 
 from .. import config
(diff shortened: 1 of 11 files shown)


api/handlers/refererhandler.py
```diff
@@ -88,6 +96,31 @@ def post(self, cont_name, cid):
...
+    def put(self, cont_name, **kwargs):
+        cid = kwargs.pop('cid')
+        _id = kwargs.pop('_id')
+
+        parent = self.storage.get_parent(cont_name, cid)
+        permchecker = self.get_permchecker(parent)
+        permchecker(noop)('PUT')
+
+
+        payload = self.request.json_body
```
nagem: This would technically 500 if the `PUT` contained no request body, not just `{}`.

nagem: We do this everywhere, it would be nice to have a decorator such as `@verify_payload_exists` that just does this for us. Feel free to add one if you'd like :)

api/handlers/refererhandler.py
```diff
@@ -88,6 +96,31 @@ def post(self, cont_name, cid):
...
+        if not payload:
+            self.abort(400, 'PUT request body cannot be empty')
+        self.update_validator(payload, 'PUT')
+
+        payload['modified'] = datetime.datetime.utcnow()
+        
+        try:
+            result = self.storage.update_el(_id, payload)
+        except APIStorageException as e:
```
nagem: This would only ever get thrown when we can't transform the `_id` to an ObjectId. We have route filters that require an ObjectId structure. I think allowing the APIStorageException to bubble up on it's own and cause a 500 is a better way to handle it, it would show us where there are holes in the filter, if any.

nagem:
Overall looks good, I did make some comments on code that is following existing examples in our codebase, but those existing examples have issues I'd like to correct with time. Mind correcting them here and we can go back and cleanup problems as we have time?

nagem:
I simplified the decorator, with that change LGTM!

PR #871, opened 2017-07-21
Fix permissions so rw users can POST sessions and acquisitions

nagem:
Tests were added for use case. Group `admin` rights are still required to create new project. 

### Review Checklist


nagem:
FYI @ryansanford

PR #877, opened 2017-07-26
PHI permission Enhancements

### LINKED ISSUE #869: PHI enhancements
Requested PHI enhancements:
  - [ ] add new permission type:`ro-no-phi aka Read Only (No PHI)`* which will always filter out PHI fields on all endpoints, both `get` and `get_all`. Otherwise, it will have the same permissions as the existing `ro` user. 
  - [ ] add new query param to `get_all` endpoints called `phi=true`*, defaults to false when not present. Using this flag will prevent the filtering of PHI and will log access to each item in the list. Logging an action for viewing the session and the subject seems most appropriate while subjects are subdocuments of sessions. 
  - [ ] Investigate using an aggregation pipeline with a `$project` on `find`s via Mongo to replace PHI fields with a default value to indicate the PHI value has been withheld. This would allow the user to know this field is populated, but they do not have access to it.** 

\* These names can be changed in actual implementation
** Still up for discussion, sometimes even having a PHI field present is a breech of PHI depending on the field.

hkethi002:
Fixes #869 
### Changes Made
- New `phi-access` permission property blacklists phi fields (This is a frontend change)
  - phi-access is a required property for every permission `POST` except at group level, defaults to true
- `phi` flag can be used to return PHI fields on `get_all` calls

|           |         | phi=false/missing                          | phi=true/superuser invoked      | 
|-----------|---------|-------------------------------------|------------------|
| phi=true  | get_all | 403                                 | phi returned     |
| phi=false/missing | get_all | phi not returned                    | phi not returned |
| phi=true  | get     | phi not returned                              | phi returned     |
| phi=false/missing | get     | phi not returned | phi returned*     |

\* This is to support existing behavior

**This PR has not applied PHI permissions to POST and PUT methods on containers**

### Review Checklist

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -42,7 +42,6 @@ class ContainerHandler(base.RequestHandler):
         'sessions': True,
         'acquisitions': True
     }
-    default_list_projection = ['files', 'notes', 'timestamp', 'timezone', 'public']
 
     # Hard-coded PHI fields, will be changed to user set PHI fields
     PHI_FIELDS = {'info': 0, 'subject.firstname':0, 'subject.lastname': 0, 'subject.sex': 0,
@@ -61,7 +60,6 @@ class ContainerHandler(base.RequestHandler):
             'parent_storage': containerstorage.GroupStorage(),
             'storage_schema_file': 'project.json',
             'payload_schema_file': 'project.json',
-            'list_projection': {'info': 0},
             'propagated_properties': ['archived', 'public'],
             'children_cont': 'sessions'
         },
@@ -72,10 +70,6 @@ class ContainerHandler(base.RequestHandler):
             'storage_schema_file': 'session.json',
             'payload_schema_file': 'session.json',
             # Remove subject first/last from list view to better log access to this information
-            'list_projection': {'info': 0, 'analyses': 0, 'subject.firstname': 0,
-                                'subject.lastname': 0, 'subject.sex': 0, 'subject.age': 0,
-                                'subject.race': 0, 'subject.ethnicity': 0, 'subject.info': 0,
-                                'files.info': 0, 'tags': 0},
             'propagated_properties': ['archived'],
             'children_cont': 'acquisitions'
         },
@@ -84,8 +78,7 @@ class ContainerHandler(base.RequestHandler):
             'permchecker': containerauth.default_container,
             'parent_storage': containerstorage.SessionStorage(),
             'storage_schema_file': 'acquisition.json',
-            'payload_schema_file': 'acquisition.json',
-            'list_projection': {'info': 0, 'collections': 0, 'files.info': 0, 'tags': 0}
+            'payload_schema_file': 'acquisition.json'
         }
     }
 
@@ -322,8 +315,6 @@ class ContainerHandler(base.RequestHandler):
         self.config = self.container_handler_configurations[cont_name]
         self.storage = self.config['storage']
         projection = None
-        # if self.is_true('info'): Seems redundant with new phi functionality
-        #     projection.pop('info')
         if self.is_true('permissions'):
             if not projection:
                 projection = None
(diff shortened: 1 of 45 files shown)


ryansanford:
Confirmed today that the scope of this PR does not include ability to configure which fields are impacted.

kofalt:
I'll need to hop on this to make sure the SDK will work with this change.

kofalt:
Questions:

1. How can I tell the difference between a field that was censored, and a field set to `***`?
1. Any objection to changing `no-phi-ro` to `ro-no-phi`? Seems more logical.
1. You mention that PHI is returned on a `get_all` call when you have > ro-no-phi access. Can you give an example?

hkethi002:
1. We are removing `***`
2. Nope
3. If your access is ro, and the phi param is set to true on a get_all, the PHI will be returned and logged.

api/handlers/containerhandler.py
```diff
@@ -346,6 +349,8 @@ def get_all(self, cont_name, par_cont_name=None, par_id=None):
                 result = containerutil.get_stats(result, cont_name)
             result = self.handle_origin(result)
             modified_results.append(result)
+            if self.is_true('phi'):
+                self.log_user_access(AccessType.view_container, cont_name, result.get('_id'))
```
nagem: We'll probably want to create a more efficient way to call this with a list. Right now it will recreate the hierarchy from this container up. If this level is acquisitions, it means grabbing the same session-project-group from the db for each acquisition, even when it's the same session-project-group.

test/integration_tests/python/test_access_log.py
```diff
@@ -245,8 +245,8 @@ def test_access_log_fails(data_builder, as_admin, log_db):
     r = as_admin.delete('/projects/' + project + '/files/' + file_name)
     assert r.status_code == 500
 
+    log_db.command('collMod', 'access_log', validator={}, validationLevel='strict')
+
     r = as_admin.get('/projects/' + project)
     assert r.ok
     assert r.json()['files']
```
nagem: Why did this change order?

hkethi002: If I remember correctly, it had to do with some persistence of the validator when the test failed before the moved line that made it difficult to understand the underlying error.

nagem:
I see some tests were added, mind adding a few more to get diff coverage closer to 100%? Looks like some of the missed lines involved PHI permissions, covering those lines are most important.

api/handlers/containerhandler.py
```diff
@@ -296,14 +306,20 @@ def get_jobs(self, cid):
     def get_all(self, cont_name, par_cont_name=None, par_id=None):
         self.config = self.container_handler_configurations[cont_name]
         self.storage = self.config['storage']
-
-        projection = self.config['list_projection'].copy()
-        if self.is_true('info'):
-            projection.pop('info')
+        projection = None
+        # if self.is_true('info'): Seems redundant with new phi functionality
+        #     projection.pop('info')
```
nagem: This can be removed

ryansanford:
I need to create an instance tracking this branch for integration testing.

ryansanford:
Was it intentional for phi-access to be a required parameter on group POST? Could we get a summary in PR description where this is to be required, and what the default is where it is not required?

```
2017-11-14 14:41:22      scitran.api                  base.py    54:INFO Initialized request request_id=d505f2d1-1510670482
2017-11-14 14:41:22      scitran.api                  base.py   340:WARN u'phi-access' is a required property

Failed validating u'required' in schema:
    {u'$schema': u'http://json-schema.org/draft-04/schema#',
     u'allOf': [{u'$ref': u'../definitions/permission.json#/definitions/permission'}],
     u'key_fields': [u'_id'],
     u'required': [u'_id', u'access', u'phi-access'],
     u'type': u'object'}

On instance:
    {u'_id': u'ci@flywheel.io', u'access': u'admin'} request_id=d505f2d1-1510670482
[pid: 23|app: 0|req: 28/58] 172.20.0.11 () {40 vars in 576 bytes} [Tue Nov 14 14:41:22 2017] POST /api/groups/scitran/permissions => generated 447 bytes in 14 msecs (HTTP/1.1 400) 3 headers in 121 bytes 
(2 switches on core 1)
```

hkethi002:
Having the phi-access property at the group level would simplify the propagation logic and allow projects and groups to keep the same permission logic. However I am not sure if that would complicate the frontend more than its worth, so I can change it if it would make more sense to only require it at the project and collection level permissions.

hkethi002:
Also, it may be better to track #991 as it is branched off of this PR.

nagem:
Closing PR for now as requirements for this enhancement are reworked within the team.

PR #883, opened 2017-07-31
Groups join=projects Parameter

### LINKED ISSUE #853: Allow project label join to '/groups' endpoint
This would allow each project associated with a group to be displayed without calling `group/<groupId>/projects` for each group.

hkethi002:
Fixes #853 
For `api/groups?join=projects`, each group result is returned with a a list of that group's project that the user has access to, permissions are filtered for the projects as well.
Specifically returned for each project are its
- id
- label
- description
- permissions

### Review Checklist

### DIFF
--- a/api/dao/base.py
+++ b/api/dao/base.py
@@ -80,12 +80,17 @@ class ContainerStorage(object):
             cont[CHILD_MAP[self.cont_name]] = children
         return cont
 
-    def get_children(self, _id, projection=None):
+    def get_children(self, _id, projection=None, uid=None):
         try:
             child_name = CHILD_MAP[self.cont_name]
         except KeyError:
             raise APINotFoundException('Children cannot be listed from the {0} level'.format(self.cont_name))
-        query = {self.cont_name[:-1]: bson.ObjectId(_id)}
+        if self.cont_name == 'groups':
+            query = {self.cont_name[:-1]: _id}
+            if uid:
+                query['permissions'] = {'$elemMatch': {'_id': uid}}
+        else:
+            query = {self.cont_name[:-1]: bson.ObjectId(_id)}
         if not projection:
             projection = {'info': 0, 'files.info': 0, 'subject': 0, 'tags': 0}
         return ContainerStorage.factory(child_name).get_all_el(query, None, projection)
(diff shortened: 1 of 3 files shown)


api/dao/base.py
```diff
@@ -80,12 +80,17 @@ def get_container(self, _id, projection=None, get_children=False):
...
         return cont
 
-    def get_children(self, _id, projection=None):
+    def get_children(self, _id, projection=None, uid=None):
         try:
             child_name = CHILD_MAP[self.cont_name]
         except KeyError:
             raise APINotFoundException('Children cannot be listed from the {0} level'.format(self.cont_name))
-        query = {self.cont_name[:-1]: bson.ObjectId(_id)}
+        if self.cont_name == 'groups':
```
nagem: We should use `self.use_object_id` here instead of keying directly on groups, right now there isn't a difference because the other container that don't use an object id don't ever call get_children

api/dao/base.py
```diff
@@ -80,12 +80,17 @@ def get_container(self, _id, projection=None, get_children=False):
...
+            query = {self.cont_name[:-1]: _id}
```
nagem: We have a utility method in `contianerutil.py` that singularizes container names, mind changing this existing code to use that instead? Analysis -> Analyses ruined the "remove the s" hack (also wouldn't affect us here since analyses don't have children but good to clean up that old hack when we can).

nagem:
Looks good 👍 

Just a bit of cleanup. Also FYI @dpuccetti

api/dao/base.py
```diff
@@ -80,12 +80,17 @@ def get_container(self, _id, projection=None, get_children=False):
...
             child_name = CHILD_MAP[self.cont_name]
         except KeyError:
             raise APINotFoundException('Children cannot be listed from the {0} level'.format(self.cont_name))
-        query = {self.cont_name[:-1]: bson.ObjectId(_id)}
+        if not self.use_object_id:
+            query = {containerutil.singularize(self.cont_name): _id}
+            if uid:
+                query['permissions'] = {'$elemMatch': {'_id': uid}}
+        else:
+            query = {self.cont_name[:-1]: bson.ObjectId(_id)}
```
nagem: Just one more right here.

nagem: And wouldn't you want to respect the uid match no matter the container? I must have missed that in the first review.

nagem:
Changes look good, thanks!
Feel free to merge.

PR #884, opened 2017-08-01
Tests added to show download filters working

hkethi002:
Tests for the download filters
### Review Checklist


nagem:
Thanks! It's interesting they were written to support filtering by tags on files, as that's not a very well supported use case (tagging files, usually it's containers). It will need to get supplemented/modified to support the requested use case of filtering by a broader list of fields but it's nice to know what is there works.

ryansanford:
@hkethi002 
General question, not a blocker for this PR: I was expecting a bit for the coverage to go up, but it remained the same. Guessing that's because the same file type/tag filtering logic gets covered by other tests?

hkethi002:
Yeah, I think the filter is used already for tags but not types however, asserts are never made as to whether the filter actually filters the files

PR #888, opened 2017-08-01
Two small cleanups

kofalt:
1. Remove some debug logs that are no longer needed 
1. Set the gears library to a released tag!


nagem:
Changes LGTM.

PR #891, opened 2017-08-07
Add ci coverage check

### LINKED ISSUE #873: Add coverage check to CI
- [x] exclude remaining uncovered lines to get to 100%
   - at least with a custom comment for grepping
   - preferably with easy way of still running coverage disregarding these "artificial exclusions"
- [x] add final check for 100% coverage after tests

ambrussimon-invenshure:
Closes #873 

### Review Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+# pragma: no cover - the standard coverage exclusion pattern
+# To be used for code that's impossible to reach and/or not worth the effort.
+
+# cover 100 - custom coverage exclusion pattern
+# Used for one-time jump to 100% coverage to enable checking whether test were added for new code.
+# Not meant for further use - existing exclusions are considered TODOs.
+
+# Adding code to excluded blocks should be considered carefully and avoided if possible.
+
+[report]
+exclude_lines =
+    pragma: no cover
+    cover 100
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 /runtime
 bootstrap.json
 .cache
-/.coverage*
+/.coverage
 coverage.xml
 /htmlcov
 node_modules/
--- a/api/auth/__init__.py
+++ b/api/auth/__init__.py
@@ -87,7 +87,7 @@ def require_superuser(handler_method):
         return handler_method(self, *args, **kwargs)
     return check_superuser
 
-def require_drone(handler_method):
+def require_drone(handler_method): # cover 100
     """
     A decorator to ensure the request is made as a drone.
 
--- a/api/auth/authproviders.py
+++ b/api/auth/authproviders.py
@@ -37,7 +37,7 @@ class AuthProvider(object):
         else:
             raise NotImplementedError('Auth type {} is not supported'.format(auth_type))
 
-    def validate_code(self, code, **kwargs):
+    def validate_code(self, code, **kwargs): # cover 100
         raise NotImplementedError
 
     def ensure_user_exists(self, uid):
@@ -71,7 +71,7 @@ class AuthProvider(object):
             if not token:
                 # user does not have refresh token, alert the client
                 raise APIRefreshTokenException('invalid_refresh_token')
-            else:
+            else: # cover 100
                 # user does have a previously saved refresh token, move on
                 return
 
@@ -331,7 +331,7 @@ class APIKeyAuthProvider(AuthProvider):
         user = config.db.users.find_one_and_update({'api_key.key': key}, {'$set': {'api_key.last_used': timestamp}}, ['_id'])
         if user:
             return user['_id']
-        else:
+        else: # cover 100
             raise APIAuthProviderException('Invalid API key')
 
 
--- a/api/auth/containerauth.py
+++ b/api/auth/containerauth.py
@@ -34,7 +34,7 @@ def default_container(handler, container=None, target_parent_container=None):
                 if target_parent_container:
                     has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
                 else:
-                    has_access = _get_access(handler.uid, container) >= INTEGER_PERMISSIONS[required_perm]
+                    has_access = _get_access(handler.uid, container) >= INTEGER_PERMISSIONS[required_perm] # cover 100
             elif method == 'PUT' and target_parent_container is not None:
                 has_access = (
                     _get_access(handler.uid, container) >= INTEGER_PERMISSIONS['admin'] and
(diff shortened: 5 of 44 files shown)


kofalt:
Ambrus, I know this PR is experimental, so I want to thank you for trying this out before offering my feedback. My comments are on the project, not your work :)



---

I don't support line-level semantic comments, especially to this extent. It's a clutter on the codebase and an abuse of the purpose of having test coverage in the first place. In my opinion attempting to artificially jump to 100, then measuring the drop, is worse in any meaningful way than simply measuring the drop from an honest percentage.

For what it's worth, I would be a bit more amiable to class- or file-level exclusion comments, that isolate entire swathes of functionality. That's at bit more honest and clearly stated, as opposed to "it's inconvenient that [this error handling line](https://github.com/scitran/core/blob/73d95059037e016669d09e7ff99aa7edea00a816/api/auth/listauth.py#L35) is not covered".

It's very tempting to use technical solutions to a social problem. In the software engineering industry this is almost always a mistake. I assert that "we have more priorities than code coverage alone" is a social problem, and "sprinkle comments on hundreds of lines to bump a metric" is a technical solution.

I also don't think the stated purpose of `cover 100` is pragmatic or realistic. Right now that's at 357 occurrences - a whopping 4% of the code base - which makes me think that "Not meant for further use - existing exclusions are considered TODOs" is unlikely to remain true.

Overall this gets a strong :-1: from me.

ryansanford:
@ambrussimon-invenshure Thanks for the PR. 👍 

@kofalt 

I want a culture where tested code is the default. I want to lean on tools to predictably and consistently promote that culture. I want opting code out of coverage to be visible and explicit.

I view this PR as an effective means to achieve the above.

> I don't support line-level semantic comments, especially to this extent. It's a clutter on the codebase and an abuse of the purpose of having test coverage in the first place. 

Could you help me understand the extend of your objection here? Is it specifically semantic comment on the same line as the code? Is there a more appealing approach to mark individual lines?

> In my opinion attempting to artificially jump to 100, then measuring the drop, is worse in any meaningful way than simply measuring the drop from an honest percentage.

Measuring a drop in percentage means coverage rates for new development needs to be just as good/bad as the old.

> For what it's worth, I would be a bit more amiable to class- or file-level exclusion comments, that isolate entire swathes of functionality. That's at bit more honest and clearly stated, as opposed to "it's inconvenient that this error handling line is not covered".

Skipping existing files/modules does nothing to encourage future development within that file/module to be covered.

> It's very tempting to use technical solutions to a social problem. In the software engineering industry this is almost always a mistake. I assert that "we have more priorities than code coverage alone" is a social problem, and "sprinkle comments on hundreds of lines to bump a metric" is a technical solution.

The goal is not to achieve 100% coverage for the sake of stating we have 100% coverage.  The goal is to move forward with higher standards for new development.

> I also don't think the stated purpose of cover 100 is pragmatic or realistic. Right now that's at 357 occurrences - a whopping 4% of the code base - which makes me think that "Not meant for further use - existing exclusions are considered TODOs" is unlikely to remain true.

I agree with the sentiment. We should be honest about what skipped lines we consider important enough for covering in the next 1-3 months, and use a phrase more specific than `cover 100` for those, and use `no cover` for the chaff.

kofalt:
> I want a culture where tested code is the default. I want to lean on tools to predictably and consistently promote that culture. 

I think this speaks to the heart of the "solve a social problem with a technical solution" anti-pattern.

My guess is that a lot of projects have those goals, yet very few have adopted something like this strategy.
Offhand, I've never heard of someone doing this. It strikes me as very strange.

> Could you help me understand the extend of your objection here?

My objection is with line-level exclusions. No coverage exclusion is great to my mind, but ranked in descending order of palatability: file, class, branch, line. [My link above](https://github.com/scitran/core/blob/73d95059037e016669d09e7ff99aa7edea00a816/api/auth/listauth.py#L35) shows one example of my reasoning. 

This information is [already exposed](https://coveralls.io/builds/12731855/source?filename=api%2Fjobs%2Fqueue.py#L51) by better tools that do not require re-implementing a highlighter by hand.

> Measuring a drop in percentage means coverage rates for new development needs to be just as good/bad as the old.

Agreed. This is true for both strategies, if I'm understanding you correctly.

> Skipping existing files/modules does nothing to encourage future development within that file/module to be covered.

Agreed! This is why I don't like coverage exclusions. Exclusions should be a "throw up our hands, this will never be tested because of effort level / state of the universe" last resort. Not a #TODO.

I think it's noteworthy that a few test frameworks I've ran across don't support exclusions at all, by design. To quote the internet: "Code coverage tells you what you definitely haven't tested, not what you have."

> The goal is to move forward with higher standards for new development.

I don't see how this change helps us do that.

I submit that there is no useful advantage to be gained moving from having X% coverage and hoping to obtain X+Y, versus having fake-100% coverage and hoping to obtain < 357 exclusion comments.

ambrussimon-invenshure:
I completely agree that this solution is far from elegant, but this is a grey area in my opinion. The question I think is whether

a) there is a nicer (but still feasible!) solution at all or
b) the approach is bad enough to offset the _intended benefits_*

>> The goal is to move forward with higher standards for new development.

> I don't see how this change helps us do that.

I think the goal here (the intended benefit) is enabling enforcement of adding _some_ tests to any new code. Note that file level exclusions leave way more freedom for avoiding this requirement.

Pertaining to a) above, I was also considering the incremental comparison approach, as I agree that is a "more correct" solution. However, there are some drawbacks there, too:

1) any refactoring (removing code that's better covered than the avg) may lead to decreased coverage
2) needs more effort* to implement properly (getting previous cov%, local vs CI)

(Regarding 2., this discussion gave me the idea of always comparing to the latest master cov% _from coveralls_ effectively leaving only 1., but I'm sure there could be edge cases/problems with that too, that I just didn't think about yet.)

Again, I'm neither for or against, just trying help with sharing my 50 cents.

kofalt:
> I think the goal here (the intended benefit) is enabling enforcement of adding some tests to any new code.

I agree that this PR might help accomplish that goal, but I disagree with the approach. I think we understand each other there :)

> However, there are some drawbacks there, too

I agree with your drawbacks - these are some of the reasons our brief usage of coverall's "mark the PR red if % change is > X" was annoying and useless. This, to me, clearly marks a need for the human reviewer to make a human decision using the [information at hand](https://coveralls.io/builds/12731855/source?filename=api%2Fjobs%2Fqueue.py#L51).

This is precisely the same as if I had added 500 lines of untested code with 500 `# pragma: no cover` comments: a human would subjectively review the PR and respond.

nagem:
We discussed my opinion in a meeting, but I'll summarize here for transparency.

Ambrus I appreciate your work, thank you for taking the time to give us a real, tangible solution to discuss. I'm glad we looked into this option, but I don't believe the reality of the solution is something I would like to move forward with. In an effort to promote higher testing standards for our work we are sacrificing code cleanliness, transparency of actual coverage, and maintainability, but I believe the biggest problem with this solution is it doesn't directly address the problem of gaining more insight into _how_ a PR changes the coverage.

In the meeting, we discussed using a different tool that will give us that insight. It looks like [Codecov](https://codecov.io/) does a much better job relaying how a PR would change the coverage that exists, most importantly the coverage of the diff itself. I suggest we investigate using this tool and retire the artificial 100% coverage approach.

PR #892, opened 2017-08-07
Node search

### LINKED ISSUE #887: Search Query to Nodes endpoint
A new endpoint that takes a search query and returns the nodes template used by batch download and add to collection endpoints.
 - Elastic Search should return only return level and _id

hkethi002:
Fixes #887 
Returns list of nodes of all containers that are returned from a given search query
`/dataexplorer/search/nodes`
### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -470,55 +470,82 @@ class DataExplorerHandler(base.RequestHandler):
         aggs['by_session']['subject.age'] = age_node['subject.age']
         return {'facets': aggs}
 
+    def search_size(self, return_type):
+        body = {
+            "size": 0,
+            "aggs" : {
+                "count" : {
+                    "cardinality" : {
+                        "field" : return_type + "._id",
+                        "precision_threshold": 100000
+                    }
+                }              
+            }
+        }
+
+        size = config.es.search(
+            index='data_explorer',
+            doc_type='flywheel',
+            body=body)['aggregations']['count']['value']
+        size = int(size*1.02)
+        return size
+
     def get_nodes(self):
 
         return_type, filters, search_string = self._parse_request()
         if return_type == 'file':
-            return get_file_nodes(return_type, filters, search_string)
+            return self.get_file_nodes(return_type, filters, search_string)
 
-        query = {
-            "bool": {
-                "must": {
-                    "match": {
-                        "_all": search_string
+        size = self.search_size(return_type)
+        body = {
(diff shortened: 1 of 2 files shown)


api/handlers/dataexplorerhandler.py
```diff
@@ -450,6 +450,112 @@ def get_facets(self):
...
+    def get_nodes(self):
+
+        return_type, filters, search_string = self._parse_request()
+        if return_type == 'file':
+            return self.get_file_nodes(return_type, filters, search_string)
+
+        size = self.search_size(return_type)
+        body = {
+            "size": 0,
+            "_source": ["session._id"],
```
nagem: we'll want this to be based on return_type, I'm sure this is just a hard code you missed taking out.

api/handlers/dataexplorerhandler.py
```diff
@@ -450,6 +450,112 @@ def get_facets(self):
...
+        return size
+
+    def get_nodes(self):
+
+        return_type, filters, search_string = self._parse_request()
+        if return_type == 'file':
+            return self.get_file_nodes(return_type, filters, search_string)
+
+        size = self.search_size(return_type)
+        body = {
```
nagem: I think we could just call "_construct_query" and change the "_source" key?

api/handlers/dataexplorerhandler.py
```diff
@@ -450,6 +450,112 @@ def get_facets(self):
...
+                "filter": {
+                    "bool" : {
+                        "must" : [
+                            { "term" : {"container_type" : return_type}}
+                        ]
+                    }
+                }
+            }
+        }
+
```
nagem: and this would call "_construct_file_query" ?

api/handlers/dataexplorerhandler.py
```diff
@@ -450,6 +450,112 @@ def get_facets(self):
...
+        }
+        
+
+        # Add filters list to filter key on query if exists
+        if filters:
+            query['bool']['filter']['bool']['must'].extend(filters)
+        nodes = []
+        results = helpers.scan(client=config.es, query={'query': query}, scroll='5m', size=1000, index='data_explorer', doc_type='flywheel', _source=[return_type+'._id'])
+        log.debug(results)
+        for result in results:
```
nagem: and we'd want to set the source whitelist

nagem:
Changes look good! I was originally confused about removing part of the `by_container` aggregation, but ordering the results is unnecessary for this type of endpoint.

PR #894, opened 2017-08-07
Add context information to gear input files

kofalt:
1. Scalar gear configuration options (so at the moment, all of them) are exposed as environment variables to the gear. For example:

```js
"FW_CONFIG_PHONE" : "555-5555",
"FW_CONFIG_STRING2" : "ggggg",
"FW_CONFIG_BOOLEAN" : "true",
"FW_CONFIG_INTEGER" : "7",
"FW_CONFIG_STRING" : "ger4536y45",
"FW_CONFIG_NUMBER" : "3.5",
"FW_CONFIG_MULTIPLE" : "20",
```

This allows you to do something like `python call.py --number $FW_CONFIG_PHONE` if you're not using a scripting language.

2. The provided contextual `config.json` file will now carry useful information about the files, their location on disk, their location in the hierarchy (if you want to use the API from a gear), and relevant scientific information recorded in the system. For example:

```js
{
    "config" : {
        "multiple" : 20,
        "string" : "ger4536y45",
        "number" : 3.5,
        "phone" : "555-5555",
        "boolean" : true,
        "integer" : 7,
        "string2" : "ggggg"
    },

    "inputs" : {
        "dicom" : {
            "base" : "file",
            "hierarchy" : {
                "type" : "acquisition",
                "id" : "5988d38b3b48ee001bde0853"
            },
            "location" : {
                "path" : "/flywheel/v0/input/dicom/8892_1_1_localizer.dicom_3Plane_Loc_SSFSE_20150215134437_1b.nii.gz",
                "name" : "8892_1_1_localizer.dicom_3Plane_Loc_SSFSE_20150215134437_1b.nii.gz"
            },
            "object" : {
                "info" : {},
                "mimetype" : "application/octet-stream",
                "tags" : [],
                "measurements" : [],
                "type" : "nifti",
                "modality" : null,
                "size" : 2913379
            }
        },
        "file" : {
            // ..
        },
        "text" : {
            // ..
        }
    }
}
```

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -14,6 +14,7 @@ from ..dao import APIPermissionException, APINotFoundException
 from ..dao.containerstorage import ProjectStorage, AcquisitionStorage
 from ..dao.containerutil import create_filereference_from_dictionary, create_containerreference_from_dictionary, create_containerreference_from_filereference, ContainerReference
 from ..web import base
+from ..web.encoder import pseudo_consistent_json_encode
 from ..validators import InputValidationException
 from .. import config
 from . import batch
@@ -317,7 +318,7 @@ class JobsHandler(base.RequestHandler):
         # B) a user editing the file object
         #
         # You can count on neither occurring before a job starts, because the queue is not globally FIFO.
-        # So option #2 is potentially more convenient, but unintuitive and prone to customer confusion.
+        # So option #2 is potentially more convenient, but unintuitive and prone to user confusion.
 
 
         for x in inputs:
@@ -412,11 +413,11 @@ class JobHandler(base.RequestHandler):
 
         if c.get('config') is not None and c.get('inputs') is not None:
             # New behavior
-            encoded = json.dumps(c, sort_keys=True, indent=4, separators=(',', ': ')) + '\n'
+            encoded = pseudo_consistent_json_encode(c)
             self.response.app_iter = StringIO.StringIO(encoded)
         else:
             # Legacy behavior
-            encoded = json.dumps({"config": c}, sort_keys=True, indent=4, separators=(',', ': ')) + '\n'
+            encoded = pseudo_consistent_json_encode({"config": c})
             self.response.app_iter = StringIO.StringIO(encoded)
 
 
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -300,6 +300,48 @@ class JobsHandler(base.RequestHandler):
             self.abort(400, 'Gear marked as invalid, will not run!')
         validate_gear_config(gear, config_)
 
+        # Config options are stored on the job object under the "config" key
+        config_ = {
+            'config': config_,
+            'inputs': { }
+        }
+
+        # Implementation notes: with regard to sending the gear file information, we have two options:
+        #
+        # 1) Send the file object as it existed when you enqueued the job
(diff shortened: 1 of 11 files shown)


kofalt:
@ambrussimon-invenshure I was unable to get the job tests to work with this change.
I'm not 100% sure what's going on, but I think there might be a bug in the test code.

It looks like the [default gear](https://github.com/scitran/core/blob/93443c4fd70792df91b664f4b32b99cf7d2a9f8c/test/integration_tests/python/conftest.py#L110-L132) from the `DataBuilder` does not have gear inputs that match the [payload used](https://github.com/scitran/core/blob/93443c4fd70792df91b664f4b32b99cf7d2a9f8c/test/integration_tests/python/test_jobs.py#L30-L45) in `test_jobs`: specifically, the input names don't match up. If I'm reading this correctly, this wasn't a valid payload and the main difference is that we crash when that happens now :)

I tried modifying line 24 of `test_jobs` to set an explicit `gear=` like the line after it, but that doesn't seem to help. The crash can be found [here](https://travis-ci.org/scitran/core/builds/261999515#L1848-L1853), I'm pretty sure. It also seems (not sure) that the referenced file is missing, which would be another problem.

Could you help me get to the point where these tests pass?
If we can get there, I should be able to add more asserts to test the new functionality.

api/jobs/handlers.py
```diff
@@ -275,6 +275,48 @@ def add(self):
...
+        # 1) Send the file object as it existed when you enqueued the job
+        # 2) Send the file object as it existed when the job was started
+        #
+        # Option #2 is possibly more convenient - it's more up to date - but the only file modifications after a job is enqueued would be from
+        #
+        # A) a gear finishing, and updating the file object
+        # B) a user editing the file object
+        #
+        # You can count on neither occurring before a job starts, because the queue is not globally FIFO.
+        # So option #2 is potentially more convenient, but unintuitive and prone to customer confusion.
```
nagem: s/customer/user 🙃

api/jobs/handlers.py
```diff
@@ -336,12 +378,22 @@ def get_config(self, _id):
...
-        # Serve config as formatted json file
-        encoded = json.dumps({"config": c}, sort_keys=True, indent=4, separators=(',', ': ')) + '\n'
-        self.response.app_iter = StringIO.StringIO(encoded)
+        # Detect if config is old- or new-style.
+        # TODO: remove this logic with a DB upgrade, ref database.py's reserved upgrade section.
+
+        if c.get('config') is not None and c.get('inputs') is not None:
+            # New behavior
+            encoded = json.dumps(c, sort_keys=True, indent=4, separators=(',', ': ')) + '\n'
+            self.response.app_iter = StringIO.StringIO(encoded)
```
nagem: If this is used multiple places, could we add it as a named encoder in web/encoder.py

api/jobs/handlers.py
kofalt: Done

api/jobs/handlers.py
kofalt: Done

nagem:
Broken tests have been resolved.

api/jobs/handlers.py
```diff
@@ -2,7 +2,6 @@
 API request handlers for the jobs module
 """
 import bson
-import json
```
kofalt: lol

PR #899, opened 2017-08-17
Fix loading of analyses in gear suggest endpoint

nagem:
Fixes #899 

Analyses used to be attached to containers, now they need to be loaded separately. 

### Review Checklist


nagem:
Small change, will merge when CI is green

ryansanford:
@nagem Can we get a test that will catch this type of regression in the future?

nagem:
@ryansanford Writing a test would have taken 10x the time it took to fix it, so I wanted to get that part of the change merged first. I'll add a test in a followup PR.

lmperry:
@ryansanford can you ping me and @tcbtcb when there is a release with this patch? Columbia and VA Palo Alto need this.

PR #905, opened 2017-08-23
Fix UID upload to new project with same UID bug 

hkethi002:
`/upload/uid` respects the project label when uploading and so will not upload to other projects
### Review Checklist

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -370,6 +370,7 @@ def _create_query(cont, cont_type, parent_type, parent_id, upload_type):
         return q
     elif upload_type == 'uid':
         return {
+            parent_type : bson.ObjectId(parent_id),
             'uid': cont['uid']
         }
     else:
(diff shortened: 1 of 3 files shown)


ryansanford:
Good catch @hkethi002 

I presume step 2 below will behave the same if step 1 was an upload to `/upload/uid` at different project?
1. Reaper uploads via `/upload/reaper` with session uid `abc`
2. Client uploads via `/upload/uid` to same group, different project with session uid `abc`

I assume that if steps 1 and 2 are reversed, we would end up with 1 session and not 2. It is still the responsibility of clients that use `/upload/reaper` to ensure their uid values are globally unique.

hkethi002:
Correct, the `upload/uid` endpoint only considers the sessions/acquisitions in the project that was provided by the client and so anything that was uploaded to another project is ignored.

nagem:
Thanks for writing a test as well, and for finding the bug. LGTM!

PR #906, opened 2017-08-24
Add info_exists key to file results, fix count bug when switching between files and containers

nagem:
### Review Checklist


nagem:
@ryansanford

nagem:
The changes to make the unit tests work were pretty ugly. When files become first class objects in the API and the info key is being actively used/supported at other levels of the hierarchy, we should move the calculation of `info_exists` to before the document is indexed.

PR #919, opened 2017-08-30
Download hierarchy uses correct paths

### LINKED ISSUE #917: Download endpoint creating new subject folder for each acquisition
From @lmperry:

Selecting a single session with many acquisitions:
image

Leads to many subject folders, each with one of the acquisitions:
image

It is likely the path logic that finds an existing folder for the file before creating a new one is broken.

hkethi002:
Fixes #917 
### TODO
- add more tests specifically with regards to #917 
### Changes 
- `_path_from_container()` uses container id instead of parent id, allows us to use new dictionary, `ids_of_paths` which checks if the path is used for a different container, if it is, then it will create a new path.
### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -111,6 +111,7 @@ class Download(base.RequestHandler):
         filename = None
 
         used_subpaths = {}
+        ids_of_paths = {}
         base_query = {}
         if not self.superuser_request:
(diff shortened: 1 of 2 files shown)


nagem:
Changes look good, thanks! I'll take a look again when you add the additional test you mentioned.

PR #925, opened 2017-09-08
Add all and simple options to search

nagem:
Adds ability to pass `simple=true` query param to get results returned in a simple format (a list of the results without the elastic wrappers).

Adds ability to pass `size=all` query param which will return all results rather than a specific size (or default size of 100)

### Breaking Changes
None

### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -487,7 +487,7 @@ class DataExplorerHandler(base.RequestHandler):
                         "field" : return_type + "._id",
                         "precision_threshold": 100000
                     }
-                }              
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


api/handlers/dataexplorerhandler.py
```diff
@@ -562,12 +562,28 @@ def search_fields(self):
...
+
         results = self._run_query(self._construct_query(return_type, search_string, filters, size), return_type)
-        response = {'results': results}
-        if self.is_true('facets'):
-            response['facets'] = self.get_facets()
-        return response
+
+        if self.is_true('simple'):
+            #return a list of the results' `_source` key only
+            return [x['_source'] for x in results]
```
kofalt: If `simple` and `facets` are set to true, the endpoint should reject.

kofalt:
Code LGTM other than a small change.

PR #926, opened 2017-09-11
Code is sent with subject and test doesn't filter out acquisition files

hkethi002:
### Changes
- Subject code is sent with subject to `_path_from_container`
- Download filter was filtering out the acquisition files (expected) so part of the filter was removed
### Review Checklist


test/integration_tests/python/test_download.py
```diff
@@ -60,6 +56,7 @@ def test_download(data_builder, file_form, as_admin, api_db):
     # Verify a single file in tar with correct file name
     found_second_session = False 
     for tarinfo in tar:
+        print tarinfo.name
```
nagem: Might want to take this out, but looks good!

nagem:
LGTM!

PR #927, opened 2017-09-11
Add CAS Auth Type

nagem:
Adds CAS Auth type support

### Breaking Changes
None

### Review Checklist


api/auth/authproviders.py
```diff
@@ -278,6 +281,56 @@ def set_user_avatar(self, uid, identity): # pragma: no cover
...
+        uid = self.validate_user(code)
+        return {
+            'access_token': code,
+            'uid': uid,
+            'auth_type': self.auth_type,
+            'expires': datetime.datetime.utcnow() + datetime.timedelta(days=14)
+        }
+
+    def validate_user(self, token):
+        config.log.warning('the config is {}\n\n'.format(self.config))
```
kofalt: Remove when you're ready to merge, or add more text to specify

api/auth/authproviders.py
```diff
@@ -278,6 +281,56 @@ def set_user_avatar(self, uid, identity): # pragma: no cover
...
+
+        return uid
+
+    def _parse_xml_response(self, response):
+
+        # parse xml
+        tree = ElementTree.fromstring(response)
+
+        # check to see if xml response labeled request as success
+        if tree[0].tag.endswith('authenticationSuccess'):
```
kofalt: Add a link to where this logic is sourced from - just a github link to cas library source is fine.

api/auth/authproviders.py
```diff
@@ -278,6 +281,56 @@ def set_user_avatar(self, uid, identity): # pragma: no cover
...
+
+            # get username from response
+            namespace = tree.tag[0:tree.tag.index('}')+1]
+            username = tree[0].find('.//' + namespace + 'user').text
+
+        else:
+            raise APIAuthProviderException('Auth provider ticket verification unsuccessful.')
+
+        if not username:
+            raise APIAuthProviderException('Auth provider did not provide username')
```
kofalt: Will these two exceptions be client-visible? Maybe remove "auth provider" from verbiage.

nagem: We use that generically throughout AuthProviders but it is not shown to the client.

nagem: Also this line has been removed.

ryansanford:
@nagem Environmental var `SCITRAN_SITE_INACTIVITY_TIMEOUT` is now working. Could we change the name to make it obvious the value represents seconds, or at a minimum document that in `sample.config`? Thanks.

api/web/base.py
```diff
@@ -127,6 +127,27 @@ def authenticate_user_token(self, session_token):
...
+                # If now - last_seen is greater than inactivity timeout, clear out session
+                if last_seen and (timestamp - last_seen).total_seconds() > inactivity_timeout:
+
+                    # Token expired and no refresh token, remove and deny request
+                    config.db.authtokens.delete_one({'_id': cached_token['_id']})
+                    config.db.refreshtokens.delete({'uid': cached_token['uid'], 'auth_type': cached_token['auth_type']})
+                    self.abort(401, 'Inactivity timeout')
+
+                # set last_seen to now
+                config.db.authtokens.update_one({'_id': cached_token['_id']}, {'$set': {'last_seen': timestamp}})
```
kofalt: I would have assumed that the user's `last_seen` is always getting updated on every request.
Are you sure it needs to be set here? If so, was it not being updated elsewhere?

PR #933, opened 2017-09-20
Adds user permissions when updating group permissions to projects without the user

hkethi002:
### Review Checklist


nagem:
It's too bad it'll take two calls, but changes look good. Maybe we can redesign how we propagate permissions at some point to make it a little smoother.

PR #934, opened 2017-09-20
New endpoint that returns count, mb size, and nodes for each file type

hkethi002:
### Notes
- New endpoint `POST /download/summary`
- request body 
```
[{
    "level": {"projects"|"sessions"|"acquisitions"|"analyses"},
     "_id": {ContainerId}
}]
```
- response
```
{
    "dicom": {
        "count" : 1,
        "mb_total": 23124,
        "_id": "dicom"
    }
    .
    .
    .
}
```
### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -318,42 +318,62 @@ class Download(base.RequestHandler):
 
     def summary(self):
         """Return a summary of what has been/will be downloaded based on a given query"""
-        req = self.request.json_body
-        req['_id'] = bson.ObjectId(req['_id'])
-        level = req['level']
-
-        containers = ['projects', 'sessions', 'acquisitions']
-        cont_query = {}
-        if level == 'projects':
-            # Grab sessions and their ids
-            sessions = config.db.sessions.find({'project': req['_id']}, {'_id': 1})
-            session_ids = [s['_id'] for s in sessions]
-
-            # for each type of container below it will have a slightly modified match query
-            cont_query = {
-                'projects': {'_id': req['_id']},
-                'sessions': {'project': req['_id']},
-                'acquisitions': {'session': {'$in': session_ids}}
-            }
-        elif level == 'sessions':
-
-            # for each type of container below it will have a slightly modified match query
-            cont_query = {
-                'sessions': {'_id': req['_id']},
-                'acquisitions': {'session': req['_id']}
-            }
-            containers = containers[1:]
-        elif level == 'acquisitions':
-
-            cont_query['acquisitions'] = {'_id': req['_id']}
-            containers = containers[-1:]
-        elif level == 'analyses':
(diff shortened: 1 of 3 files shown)


nagem:
Looks good, but we don't actually need the node list returned with the summary results. Are you providing the node list to make the eventual download request more efficient?

api/download.py
```diff
@@ -315,3 +315,92 @@ def download(self):
...
+                    'mb_total': {'$sum':'$mbs'}, 
+                    'nodes' : {
+                        '$addToSet': {'level': {'$literal':cont_name}, '_id': '$_id'}
+                    }
+                }}
+            ]
+
+            try:
+                result = config.db.command('aggregate', cont_name, pipeline=pipeline)
+            except Exception as e: # pylint: disable=broad-except
```
nagem: The string version of the failure would get returned to the API user. In situations like this it's better to internally log the error (`log.warning(e)`) and then `self.abort(500, 'Failure to load summary')` or something like that so the user doesn't get sent a mongo error.

api/download.py
```diff
@@ -315,3 +315,92 @@ def download(self):
...
+                'analyses': {'parent.id': {'$in': parent_ids}}
+            }
+            containers = containers[1:]
+        elif level == 'acquisitions':
+
+            cont_query['acquisitions'] = {'_id': req['_id']}
+            containers = ['acquisitions']
+        elif level == 'analyses':
+            cont_query['analyses'] = {'_id': req['_id']}
+            containers = containers[-1:]
```
nagem: We should make sure we have tests if we're going to support file summaries for analysis downloads.

api/download.py
```diff
@@ -315,3 +315,92 @@ def download(self):
...
+                'analyses': {'parent.id': {'$in': parent_ids}}
+            }
+        elif level == 'sessions':
+
+            # Grab acquisitions and their ids
+            acquisitions = config.db.acquisitions.find({'session': req['_id']}, {'_id': 1})
+            acquisition_ids = [a['_id'] for a in acquisitions]
+            parent_ids = [req['_id']] + acquisition_ids
+
+            # # Grab analyses and their ids
```
nagem: We won't want want analyses included in project/session/acquisition download summary (because they aren't included in the actual download), so it's okay to remove this.

hkethi002:
That was the idea, it seemed like an easy place to generate it because we are already aggregating through the files anyways.

nagem:
Unfortunately that list might be 1000s of acquisitions long for large projects, and the user might choose multiple filters, meaning a frontend join of multiple lists of acquisitions - which would lose the gain in efficiency.

nagem:
Thanks for making the requested changes, LGTM!

PR #936, opened 2017-09-20
Access log query project filter added

### LINKED ISSUE #924: Access log query needs to support project filter
Use case is that subject codes are expected to not be unique across projects, so the usefulness of that filter without accompanying project is limited.

The API allowing multiple projects with the same name within the same group complicates user workflow for selecting the correct project. Adding some restriction should also be considered, and turned into its own issue.

hkethi002:
Fixes #924 
### Notes
Another param, similar in function to the subject filter. Uses the project id over label, although if it is going to implemented as a text box instead of drop down, label is probably a better choice.
### Review Checklist


nagem:
LGTM!

PR #937, opened 2017-09-21
Help wanted: attempt to implement an abstract container route

kofalt:
A big problem with exposing our functionality is that everything has a 5x multiplier: `AddAcquisitionNote`. This blocks doing some useful stuff in the SDK because the functions would be irritating and confusing.

We should provide `/container/x` routes that figure out which type of container we mean, and just serve the equivalent route. This PR is an incomplete version of doing just that!

**Solved problem:** we need an efficient way to resolve the container noun. Might be solved for free if we ever can truly homogenize containers. While the func snippet + db command approach here is a bit ugly on the code, I can confirm it is quite efficient and safe to use.

**Unsolved problem:** make webapp2 just serve a different handler, dangit. The commented-out `self.redirect` approach totally works: you can `GET api/container/x/sessions` where x is a project ID. But that won't work for non-gets, and we shouldn't redirect anyway. We should just serve.

I'm 90% sure we can take these code scraps, and either:

1. Modify the (~5) fields of the request object to reflect the "proxying"
1. Construct a new context with the same body reader?

And call a handler method. Help?

### DIFF
--- a/api/handlers/abstractcontainerhandler.py
+++ b/api/handlers/abstractcontainerhandler.py
@@ -15,7 +15,9 @@ CONTAINER_SEARCH_JS = r"""
         "groups": db.getCollection("groups").findOne({"_id" : _id}, {"_id": 1}),
         "projects": db.getCollection("projects").findOne({"_id" : _id}, {"_id": 1}),
         "sessions": db.getCollection("sessions").findOne({"_id" : _id}, {"_id": 1}),
-        "acquisitions": db.getCollection("acquisitions").findOne({"_id" : _id}, {"_id": 1})
+        "acquisitions": db.getCollection("acquisitions").findOne({"_id" : _id}, {"_id": 1}),
+        "analyses": db.getCollection("analyses").findOne({"_id" : _id}, {"_id": 1}),
+        "collections": db.getCollection("collections").findOne({"_id" : _id}, {"_id": 1})
     }
 })("%s");
 """
--- a/api/handlers/abstractcontainerhandler.py
+++ b/api/handlers/abstractcontainerhandler.py
@@ -1,95 +1,64 @@
-from webob import exc
+from webapp2 import Request
 
-from ..web import base
 from .. import config
+from ..web import base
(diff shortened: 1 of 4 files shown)


kofalt:
A bunch of doc links:

https://webapp2.readthedocs.io/en/latest/guide/request.html
https://webapp2.readthedocs.io/en/latest/guide/routing.html

https://webapp2.readthedocs.io/en/latest/api/webapp2.html#webapp2.Request
https://webapp2.readthedocs.io/en/latest/api/webapp2.html#webapp2.Router
https://webapp2.readthedocs.io/en/latest/api/webapp2.html#webapp2.Route
https://webapp2.readthedocs.io/en/latest/_modules/webapp2.html#Router.default_matcher

I've mostly solved the puzzle just from hopping through various function sigs & bits of source code.

api/handlers/abstractcontainerhandler.py
```diff
@@ -1,95 +1,64 @@
...
+CONTAINER_SEARCH_JS = r"""
+(function searchContainer(_id) {
+    if (/^[a-f\d]{24}$/i.test(_id)) {
+        _id = ObjectId(_id);
+    }
+    return {
+        "groups": db.getCollection("groups").findOne({"_id" : _id}, {"_id": 1}),
+        "projects": db.getCollection("projects").findOne({"_id" : _id}, {"_id": 1}),
+        "sessions": db.getCollection("sessions").findOne({"_id" : _id}, {"_id": 1}),
+        "acquisitions": db.getCollection("acquisitions").findOne({"_id" : _id}, {"_id": 1})
```
nagem: We'll probably want to add analyses to this list as well

kofalt: Collections, Y/N ?

kofalt:
So far, I tried swapping out SDK tests for acquisitions, and it works :tada: 

More to follow.

kofalt:
This was later completed in flywheel-io/core#937

PR #938, opened 2017-09-25
Gear improvements: custom environment variables, command templating, non-scalar config values, job-bounded API keys

nagem:
Tests incoming

### JobAPIKey Usage
```
from api.auth.apikeys import JobAPIKey

# Generates API key for user within job context
# Returns API key (nonce)
api_key = JobAPIKey.generate(uid, job_id)

# Removes API key via job_id
JobAPIKey.remove(job_id)
```

### Breaking Changes
None

### Review Checklist

### DIFF
--- a/api/auth/apikeys.py
+++ b/api/auth/apikeys.py
@@ -94,6 +94,8 @@ class JobApiKey(APIKey):
         Returns an API key for user for use by a specific job.
         Re-uses such a key if it already exists.
         """
+        
+        job_id = str(job_id)
 
         existing_key = config.db.apikeys.find_one({
             'uid': uid,
@@ -106,7 +108,7 @@ class JobApiKey(APIKey):
         else:
             api_key = cls.generate_api_key(cls.key_type)
             api_key['uid'] = uid
-            api_key['job'] = str(job_id)
+            api_key['job'] = job_id
 
             config.db.apikeys.insert_one(api_key)
             return api_key['_id']
--- a/api/auth/apikeys.py
+++ b/api/auth/apikeys.py
@@ -91,14 +91,25 @@ class JobApiKey(APIKey):
     @classmethod
     def generate(cls, uid, job_id):
         """
-        Generates an API key for user for use by a specific job
+        Returns an API key for user for use by a specific job.
+        Re-uses such a key if it already exists.
         """
-        api_key = cls.generate_api_key(cls.key_type)
-        api_key['uid'] = uid
-        api_key['job'] = str(job_id)
 
-        config.db.apikeys.insert_one(api_key)
-        return api_key['_id']
+        existing_key = config.db.apikeys.find_one({
+            'uid': uid,
+            'job': job_id,
+        })
+
+        if existing_key is not None:
+            return existing_key['_id']
+
+        else:
+            api_key = cls.generate_api_key(cls.key_type)
+            api_key['uid'] = uid
+            api_key['job'] = str(job_id)
+
+            config.db.apikeys.insert_one(api_key)
+            return api_key['_id']
 
     @classmethod
     def remove(cls, job_id):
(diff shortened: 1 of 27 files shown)


api/jobs/handlers.py
```diff
@@ -280,6 +287,9 @@ def add(self):
         if submit.get('destination', None) is not None:
             destination = create_containerreference_from_dictionary(submit['destination'])
         else:
+            if len(inputs.keys()) < 1:
+                raise Exception('No destination specified and no inputs to derive from')
```
nagem: We shouldn't 500 because user data wasn't formatted as expected. 400 or something like an `InputValidationException` would be more helpful.

raml/schemas/definitions/file.json
```diff
@@ -24,7 +24,16 @@
...
                 "type":{"type":"string"},
-                "name":{"type":"string"}
+                "name":{"type":"string"},
+                "via": {
+                  "type":"object",
+                  "properties":{
+                      "id":{"type":"string"},
+                      "method":{"type":"string"},
+                      "type":{"type":"string"},
+                      "name":{"type":"string"}
```
kofalt: Could this be recursive instead? :thinking:

api/auth/apikeys.py
```diff
@@ -91,14 +91,25 @@ class JobApiKey(APIKey):
...
         """
-        api_key = cls.generate_api_key(cls.key_type)
-        api_key['uid'] = uid
-        api_key['job'] = str(job_id)
 
-        config.db.apikeys.insert_one(api_key)
-        return api_key['_id']
+        existing_key = config.db.apikeys.find_one({
+            'uid': uid,
+            'job': job_id,
```
nagem: I'd `str()` the job_id until we don't have to worry about getting passed an ObjectId

api/jobs/handlers.py
```diff
@@ -425,6 +427,35 @@ def get_config(self, _id):
...
+                    if input['base'] == 'api-key':
+                        if j.origin['type'] != 'user':
+                            raise Exception('Cannot provide an API key to a job not launched by a user')
+
+                        uid = j.origin['id']
+                        api_key = JobApiKey.generate(uid, j.id_)
+                        parsed_url = urlparse(config.get_item('site', 'api_url'))
+
+                        if parsed_url.port != 443:
+                            api_key = parsed_url.hostname + ':' + str(parsed_url.port) + ':' + api_key
```
nagem: Is there a reason to not just pass 443?

api/jobs/handlers.py
```diff
@@ -425,6 +427,35 @@ def get_config(self, _id):
...
+                gear = get_gear(j.gear_id)
+
+                for key in gear['gear']['inputs']:
+                    input = gear['gear']['inputs'][key]
+
+                    if input['base'] == 'api-key':
+                        if j.origin['type'] != 'user':
+                            raise Exception('Cannot provide an API key to a job not launched by a user')
+
+                        uid = j.origin['id']
```
nagem: A lot of times a job isn't triggered by a user, we'll want to make sure the job was triggered by a real user (`origin['type'] = 'job`) before assigning it an API key. We can re-evaluate if devices should be able to use the SDK via an API in the future but currently it is expected to be a user.

raml/schemas/definitions/file.json
nagem: Yeah I was being lazy.

api/jobs/handlers.py
kofalt: Yes, ref flywheel-io/frontend#968

api/jobs/handlers.py
kofalt: I agree with what you're saying. Maybe you missed the user check above this line?

api/auth/apikeys.py
kofalt: Good call.

api/jobs/handlers.py
nagem: I did.

PR #947, opened 2017-10-04
Intify search limit if set

kofalt:
The search `size` param would never succeed if set, because URL params are strings.


nagem:
LGTM

PR #950, opened 2017-10-05
Have term search use .raw field, alert user when query filter syntax is malformed

nagem:
### Review Checklist


kofalt:
:tada:

PR #951, opened 2017-10-09
Fix analysis download 

nagem:
The endpoint for a ticketed download at `api/analyses/files` caused issues when it did not provide enough information about the analysis to support access logging. 

A request for the future is that all container file downloads move to using the targeted downloads at `api/download`. The above endpoint should be considered legacy. 

### Review Checklist


nagem:
Test for member inclusion added.

PR #952, opened 2017-10-09
/users/uid/groups returns permission for User(uid)

### LINKED ISSUE #948: /users/user_id/groups endpoint not returning permissions
When using the `/api/users/user_id/groups` endpoint, the permissions array is empty unless `?root=true` is specified.  When the current user is a site manager I would expect this array to be populated so that the selected user's current permission level can be determined.

hkethi002:
Fixes #948 
Only returns permission for the user with the id in the url
### Review Checklist


nagem:
LGTM

PR #961, opened 2017-10-20
Custom rule config (+use gear_id)

### LINKED ISSUE #922: Add config to gear rules
Add config object to project and site gear rules that will be used as the config for jobs spawned by these rules. Validate config against gear manifest when modified/created.

### LINKED ISSUE #959: Ensure rules are properly formatted when submitted
Currently, rules are not validated when submitted to the API, but are expected to be valid and complete when used by the system. For example, a user can add a rule with a match type but not a match value, which fails a key access in the rule evaluation code. 

Proposal - add a proper json schema if one does not exist or is incomplete. Validate POST and PUT payloads on project rule endpoints. Respond with a `400` or allow the validator to raise an `InputValidationException`. 

Stack trace:
```
  File "./api/upload.py", line 188, in upload
    return process_upload(self.request, strategy, origin=self.origin, context=context)
  File "./api/upload.py", line 137, in process_upload
    placer.process_file_field(field, file_attrs)
  File "./api/placer.py", line 186, in process_file_field
    self.save_file(field, file_attrs)
  File "./api/placer.py", line 98, in save_file
    rules.create_jobs(config.db, container_before, self.container, self.container_type)
  File "./api/jobs/rules.py", line 220, in create_jobs
    jobs_before.extend(create_potential_jobs(db, container_before, container_type, f))
  File "./api/jobs/rules.py", line 181, in create_potential_jobs
    if eval_rule(rule, file_, container):
  File "./api/jobs/rules.py", line 132, in eval_rule
    if not eval_match(match['type'], match['value'], file_, container):
KeyError: 'value'
```

ambrussimon:
Resolves #922 
Resolves #959 

Could someone with more JSONSchema experience reassure me that this looks OK?
https://github.com/scitran/core/compare/rule-config-and-validation?expand=1#diff-31c3baaff2a2d926bb13668c2bbd23bdR5

### Review Checklist

### DIFF
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -8,7 +8,6 @@ import bson.objectid
 import datetime
 from jsonschema import Draft4Validator, ValidationError
 import gears as gear_tools
-import pymongo
 
 from .. import config
 from .jobs import Job
@@ -39,17 +38,10 @@ def get_gears():
     return map(lambda x: x['original'], cursor)
 
 def get_gear(_id):
-    return config.db.gears.find_one({'_id': bson.ObjectId(_id)})
-
-def get_gear_by_name(name):
-
-    # Find a gear from the list by name
-    gear_doc = list(config.db.gears.find({'gear.name': name}).sort('created', pymongo.DESCENDING))
-
-    if len(gear_doc) == 0 :
-        raise APINotFoundException('Unknown gear ' + name)
-
-    return gear_doc[0]
+    gear = config.db.gears.find_one({'_id': bson.ObjectId(_id)})
+    if gear is None:
+        raise APINotFoundException('Cannot find gear {}'.format(_id))
+    return gear
 
 def get_invocation_schema(gear):
     return gear_tools.derive_invocation_schema(gear['gear'])
--- a/api/jobs/gears.py
+++ b/api/jobs/gears.py
@@ -5,6 +5,7 @@ Gears
 from __future__ import absolute_import
 
 import bson.objectid
+import copy
 import datetime
 from jsonschema import Draft4Validator, ValidationError
 import gears as gear_tools
@@ -96,13 +97,13 @@ def suggest_for_files(gear, files):
     return suggested_files
 
 def validate_gear_config(gear, config_):
-    if len(gear.get('manifest', {}).get('config', {})) > 0:
-        invocation = gear_tools.derive_invocation_schema(gear['manifest'])
+    if len(gear.get('gear', {}).get('config', {})) > 0:
+        invocation = gear_tools.derive_invocation_schema(gear['gear'])
         ci = gear_tools.isolate_config_invocation(invocation)
         validator = Draft4Validator(ci)
 
         try:
-            validator.validate(config_)
+            validator.validate(fill_gear_default_values(gear, config_))
         except ValidationError as err:
             key = None
             if len(err.relative_path) > 0:
@@ -120,8 +121,7 @@ def fill_gear_default_values(gear, config_):
     Given a gear and a config map, fill any missing keys using defaults from the gear's config
     """
 
-    if config_ is None:
-        config_ = {}
+    config_ = copy.deepcopy(config_) or {}
 
     for k,v in gear['gear'].get('config', {}).iteritems():
         if 'default' in v:
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -24,7 +24,7 @@ from ..web.encoder import pseudo_consistent_json_encode
 from ..web.errors import APIPermissionException, APINotFoundException, InputValidationException
 from ..web.request import AccessType
 
-from .gears import validate_gear_config, get_gears, get_gear, get_invocation_schema, remove_gear, upsert_gear, suggest_container, get_gear_by_name, check_for_gear_insertion
+from .gears import validate_gear_config, get_gears, get_gear, get_invocation_schema, remove_gear, upsert_gear, suggest_container, check_for_gear_insertion
 from .jobs import Job, JobTicket, Logs
 from .batch import check_state, update
 from .queue import Queue
@@ -147,10 +147,7 @@ class RulesHandler(base.RequestHandler):
 
         validate_data(payload, 'rule-new.json', 'input', 'POST', optional=True)
         validate_regexes(payload)
-        try:
-            get_gear_by_name(payload['alg'])
-        except APINotFoundException:
-            self.abort(400, 'Cannot find gear for alg {}, alg not valid'.format(payload['alg']))
+        get_gear(payload['gear_id'])
 
         payload['project_id'] = cid
 
@@ -200,11 +197,8 @@ class RuleHandler(base.RequestHandler):
         updates = self.request.json
         validate_data(updates, 'rule-update.json', 'input', 'POST', optional=True)
         validate_regexes(updates)
-        if updates.get('alg'):
-            try:
-                get_gear_by_name(updates['alg'])
-            except APINotFoundException:
-                self.abort(400, 'Cannot find gear for alg {}, alg not valid'.format(updates['alg']))
+        if updates.get('gear_id'):
+            get_gear(updates['gear_id'])
 
         doc.update(updates)
         config.db.project_rules.replace_one({'_id': bson.ObjectId(rid)}, doc)
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -147,7 +147,7 @@ class RulesHandler(base.RequestHandler):
 
         validate_data(payload, 'rule-new.json', 'input', 'POST', optional=True)
         validate_regexes(payload)
-        get_gear(payload['gear_id'])
+        validate_gear_config(get_gear(payload['gear_id']), payload.get('config'))
 
         payload['project_id'] = cid
 
@@ -197,8 +197,9 @@ class RuleHandler(base.RequestHandler):
         updates = self.request.json
         validate_data(updates, 'rule-update.json', 'input', 'POST', optional=True)
         validate_regexes(updates)
-        if updates.get('gear_id'):
-            get_gear(updates['gear_id'])
+        gear_id = updates.get('gear_id', doc['gear_id'])
+        config_ = updates.get('config', doc.get('config'))
+        validate_gear_config(get_gear(gear_id), config_)
 
         doc.update(updates)
         config.db.project_rules.replace_one({'_id': bson.ObjectId(rid)}, doc)
--- a/api/jobs/rules.py
+++ b/api/jobs/rules.py
@@ -138,16 +138,16 @@ def eval_rule(rule, file_, container):
 
     return True
 
-def queue_job_legacy(algorithm_id, input_):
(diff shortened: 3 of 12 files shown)


kofalt:
Go ahead & do a rebase; a large amount of gear-related activity just landed on master.

ryansanford:
@kofalt 
When this lands, can we make guarantees to gear authors that config.json will always be present and contain values for all non-optional config options? I think that's the big win for gear authors with this PR.

kofalt (approved):
Brush-pass LGTM

kofalt:
@ryansanford That should generally be the case, yes.

kofalt (changes requested):

api/jobs/handlers.py
```diff
@@ -175,13 +175,14 @@ def post(self, cid):
...
-        validate_data(doc, 'rule-add.json', 'input', 'POST', optional=True)
+        validate_data(doc, 'rule-new.json', 'input', 'POST', optional=True)
         try:
-            get_gear_by_name(doc['alg'])
+            gear = get_gear_by_name(doc['alg'])
         except APINotFoundException:
             self.abort(400, 'Cannot find gear for alg {}, alg not valid'.format(doc['alg']))
 
         doc['project_id'] = cid
+        doc.setdefault('config', gear['gear']['config'])
```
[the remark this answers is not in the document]
nagem: Kofalt is referencing that the config doc on the gear stores the default value (if available) under a `default` key. I added [this function](https://github.com/scitran/core/pull/962/files#diff-bc08ea7e81709ca29a114e3e12108683R127) in `jobs.gears` that, given a gear doc and a config map, will set missing keys if a default value exists.

nagem:
LGTM, with change to use new function to set defaults for submitted config. 

The `PUT` method for rules properly does a key-wise replace. We would want to make sure any config object sent in an update does a full replace of the existing config doc, so they can properly remove keys.

ambrussimon-invenshure:
According to further discussions since, it seems that triggered jobs using the latest available gear for a given name is the present _and_ desired future behavior, so I opted for not storing a gear default config on new rules. Ties in nicely with @nagem's addition, where defaults are applied as late and as unified as possible.

@kofalt Please sanity check [manifest -> gear intended fix](https://github.com/scitran/core/pull/961/files#diff-bc08ea7e81709ca29a114e3e12108683R108).

@nagem @kofalt Should I add a check upon gear version addition, which makes sure that any rules that have a custom config still validate against the new gear manifest?

api/jobs/gears.py
```diff
@@ -105,8 +105,8 @@ def suggest_for_files(gear, files):
     return suggested_files
 
 def validate_gear_config(gear, config_):
-    if len(gear.get('manifest', {}).get('config', {})) > 0:
-        invocation = gear_tools.derive_invocation_schema(gear['manifest'])
+    if len(gear.get('gear', {}).get('config', {})) > 0:
+        invocation = gear_tools.derive_invocation_schema(gear['gear'])
```
nagem: Oh nice catch

api/jobs/handlers.py
```diff
@@ -531,7 +539,7 @@ def post(self):
         gear = get_gear(gear_id)
         if gear.get('gear', {}).get('custom', {}).get('flywheel', {}).get('invalid', False):
             self.abort(400, 'Gear marked as invalid, will not run!')
-        validate_gear_config(gear, config_)
+        validate_gear_config(gear, fill_gear_default_values(gear, config_))
```
nagem: The config defaults will be filled in `Queue.enqueue_job`. Setting it at the beginning is maybe slightly more efficient, but it will still iterate over all of the gear config keys for each batch job that is enqueued. You might have already been aware of this.

api/jobs/handlers.py
```diff
@@ -175,14 +175,17 @@ def post(self, cid):
...
         try:
-            get_gear_by_name(doc['alg'])
+            gear = get_gear_by_name(doc['alg'])
         except APINotFoundException:
             self.abort(400, 'Cannot find gear for alg {}, alg not valid'.format(doc['alg']))
 
         doc['project_id'] = cid
 
+        if 'config' in doc:
+            validate_gear_config(gear, fill_gear_default_values(gear, doc['config']))
```
kofalt: This should be called unconditionally, setting `doc['config']` to {} and calling `fill_gear_default_values` on it.

kofalt: This should be called unconditionally, setting `doc['config']` to {} and calling `fill_gear_default_values` on it.

api/jobs/handlers.py
```diff
@@ -234,6 +237,11 @@ def put(self, cid, rid):
             except APINotFoundException:
                 self.abort(400, 'Cannot find gear for alg {}, alg not valid'.format(updates['alg']))
 
+        if 'alg' in updates or 'config' in updates:
+            gear = get_gear_by_name(updates.get('alg', doc['alg']))
+            config_ = fill_gear_default_values(gear, updates.get('config', doc.get('config', {})))
+            validate_gear_config(gear, config_)
```
kofalt: I think `validate_gear_config` will throw an exception; maybe catch that and throw a @nagem -friendly exception for status codes & such?

api/jobs/handlers.py
kofalt: You're right Megan, but I think he's doing the right thing here. We probably need `fill_gear_default_values` whenever `validate_gear_config` is called.

api/jobs/rules.py
```diff
@@ -199,6 +199,9 @@ def create_potential_jobs(db, container, container_type, file_):
                 gear = gears.get_gear_by_name(alg_name)
                 job = Job(str(gear['_id']), inputs, tags=['auto', alg_name])
 
+            if 'config' in rule:
+                job.config = rule['config']
```
kofalt: Will the rule config have the defaults filled out, or will it need to be pushed through `fill_gear_default_values`?

raml/schemas/definitions/rule.json
```diff
@@ -1,22 +1,59 @@
...
-          "type":"array",
-          "items":{
-            "type":"string"
-          }
+    "definitions": {
+        "_id":         { "type": "string" },
+        "project_id":  { "type": "string" },
+        "alg":         { "type": "string" },
+        "name":        { "type": "string" },
+        "config":      { "type": "object" },
```
kofalt: I would remove these five definitions, and instead just use their `{ "type": "string" }` / etc directly.

We probably do this a lot in other parts of the schema, but it's a bit overboard.

raml/schemas/input/rule-update.json
```diff
@@ -1,62 +1,6 @@
...
-                    },
-                    "value": {
-                        "type": "string"
-                    }
-                }
-            }
-        }
-    },
-    "additionalProperties": false
+    "allOf": [{"$ref": "../definitions/rule.json#/definitions/rule-input"}]
```
kofalt: @nagem Confirm for me that this is correct?

raml/schemas/output/rule-list.json
```diff
@@ -0,0 +1,5 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "array",
+    "items": { "$ref":"../definitions/rule.json#/definitions/rule-output" }
```
kofalt: @nagem Same here?

kofalt (approved):
Diff is a little rough, but LGTM to the best of my ability

api/jobs/gears.py
```diff
@@ -111,7 +111,7 @@ def validate_gear_config(gear, config_):
         validator = Draft4Validator(ci)
 
         try:
-            validator.validate(config_)
+            validator.validate(fill_gear_default_values(gear, config_))
```
nagem: Defaults will be filled when the job is spawned, so I would think we wouldn't fill them here. What is the expected behavior if a gear is updated and the update has a new default that the rule didn't explicitly set? If we fill defaults here, it won't use the updated default. It is possible they wouldn't want to use an updated default without any user action. 

If I missed a discussion on this, let me know. @kofalt @ambrussimon

api/jobs/rules.py
nagem: The defaults will get filled in `Queue.enqueue_job()`. Let's not use `fill_defaults` when creating a job that will get enqueued through that codepath.

raml/schemas/input/rule-update.json
nagem: 👍

api/jobs/gears.py
kofalt: Given that it's a job rule, I would expect valid defaults to be used, rather than invalid ones from a different gear document. Really the core issue here is that `fill_gear_default_values` modifies the map, it's reasonable to call this in places that you won't want to change.

kofalt: So I would either not call `fill_gear_default_values` in places where you don't want to modify the document (quick fix), or (better fix) change `fill_gear_default_values` to use `copy.deepcopy`, then check the various callees.

nagem:
To be clear: we still haven't decided and acted on a solution to "what happens when someone uploads a new version of a gear whose config conflicts with an existing gear rule config", so this PR cannot be merged.

api/jobs/gears.py
nagem: That's a good point, but I was asking if we wanted defaults persisted to the database for rule config if the user didn't explicitly send them.

api/jobs/handlers.py
ambrussimon: @kofalt The current (already friendly) exc thrown by it is `APIValidationException` (422). Wondering if that's appropriate or maybe `InputValidationException` is more fitting?

api/jobs/handlers.py
ambrussimon: @kofalt Do you think unconditional validation would be preferred here, too? https://github.com/scitran/core/pull/961/files#diff-8ae88fa922271b27b6066aae46fda7abR239

nagem:
I think we've at least come to the conclusion that, when attempting to upload a gear, the user should be warned of project rules that would now be in conflict with with the gear's new config. I believe most clients use `/gears/check` before uploading a new gear, @kofalt would you say it's safe to place this warning logic as output of that endpoint? Do you have any opinions on the format of the warning? `200` but clients are expected to review the body of the response? Use a different status code?

kofalt:
`/gears/check` should definitely return 200 unless a gear upload with the submitted manifest would not succeed. Let's add a small body return:

```json
{
    "disables_project_rules": [
        {
            "path": "scitran/testdata",
            "id": "aex"
        },
        {
            "path": "your-group/whatever",
            "id": "aex2"
        },
    ]
}
```

If the array is non-empty, frontends can print that as a warning before/after an upload.

ryansanford:
This is on hold until gear rules referencing gear id (version) is implemented

ambrussimon:
@nagem Bump. Freshly rebased on top of
- support for disabled rules
- rules using gear id instead of name
(Can PR separately if desired.)

I feel like I'm missing something... So is there any auto-disabling/auto-version-bump feature required on the API side, or is that now completely left for the UI/clients to sort out? (Now that rules use gear_id)

PR #962, opened 2017-10-23
Use central job enqueue logic for jobs spawned by rules

### LINKED ISSUE #923: Jobs triggered from rules should use their default config
Right now, jobs that are launched via a rule end up with an empty config. Before the job is stored in the database, we should ensure it has the proper config - in this case, copying over the default config.

Probably a dependency of #922.

Additional details added by @nagem:

Gear documents stored in mongo will have a `config` field (possibly empty). An example:
```
"config" : {
            "multiple" : {
                "exclusiveMaximum" : true,
                "description" : "Any two-digit multiple of ten.",
                "default" : 20,
                "multipleOf" : 10,
                "maximum" : 100,
                "type" : "number"
            },
            "string" : {
                "default" : "Example",
                "type" : "string",
                "description" : "Any string."
            },
            "number" : {
                "default" : 3.5,
                "type" : "number",
                "description" : "Any number."
            },
            "phone" : {
                "default" : "555-5555",
                "pattern" : "^[0-9]{3}-[0-9]{4}$",
                "type" : "string",
                "description" : "Any local phone number, no country or area code."
            },
            "boolean" : {
                "default" : true,
                "type" : "boolean",
                "description" : "Any boolean."
            },
            "integer" : {
                "default" : 7,
                "type" : "integer",
                "description" : "Any integer."
            },
            "string2" : {
                "default" : "Example 2",
                "minLength" : 2,
                "type" : "string",
                "description" : "Any string from 2 to 15 characters long.",
                "maxLength" : 15
            }
        }
```

Before a job is persisted to the database (in `Job.save()`), if config is `None` (or an empty dictionary*), create a config map using the `default` key listed for each key in the gear's config. If there is no default, do not add the config key.

nagem:
Resolves #923 

Changes in this PR
- Jobs launched by rules will get input file context and meta
- Jobs launched by rules will get the defaults of all configs from the gear manifest

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -335,7 +335,7 @@ class JobHandler(base.RequestHandler):
         # Detect if config is old- or new-style.
         # TODO: remove this logic with a DB upgrade, ref database.py's reserved upgrade section.
 
-        if c.get('config') is not None and c.get('inputs') is not None:
+        if 'config' in c and c.get('inputs') is not None:
             # New behavior
 
             # API keys are only returned in-flight, when the job is running, and not persisted to the job object.
--- a/api/jobs/rules.py
+++ b/api/jobs/rules.py
@@ -1,6 +1,7 @@
 import fnmatch
 
 from .. import config
+from ..types import Origin
 from ..dao.containerutil import FileReference
 
 from . import gears
@@ -236,10 +237,14 @@ def create_jobs(db, container_before, container_after, container_type):
 
 
     spawned_jobs = []
+    origin ={
+        'type': str(Origin.system),
+        'id': None
+    }
 
     for pj in potential_jobs:
         job_map = pj['job'].map()
-        Queue.enqueue_job(job_map, None) # passing no origin results in system origin
+        Queue.enqueue_job(job_map, origin) # passing no origin results in system origin
 
         spawned_jobs.append(pj['rule']['alg'])
 
--- a/api/jobs/rules.py
+++ b/api/jobs/rules.py
@@ -239,10 +239,8 @@ def create_jobs(db, container_before, container_after, container_type):
 
     for pj in potential_jobs:
         job_map = pj['job'].map()
-        try:
-            Queue.enqueue_job(job_map, None) # passing no origin results in system origin
-        except Exception as e: # pylint: disable=broad-except
-            config.log.exception('Unable to enqueue job from rules: {}'.format(e))
+        Queue.enqueue_job(job_map, None) # passing no origin results in system origin
+
         spawned_jobs.append(pj['rule']['alg'])
 
     return spawned_jobs
--- a/api/jobs/rules.py
+++ b/api/jobs/rules.py
@@ -5,6 +5,7 @@ from ..dao.containerutil import FileReference
 
 from . import gears
 from .jobs import Job
+from .queue import Queue
 
 log = config.log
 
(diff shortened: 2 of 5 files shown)


ryansanford:
As-is this change means when job is spawned from gear rule, gears should expect config.json files that contain meta on the inputs, but not contain config. 

There are at least 2 gears that don't operate well in this situation (scitran-apps/dcm2niix and flywheel-apps/mriqc) and guessing several more due to the pattern being used which assumes config is included in config.json whenever that file exists.

This situation would no longer be expected once gear rules pass along config defaults from the gear manifest. https://github.com/scitran/core/pull/961

I double checked the gear spec and defaults for config options are not required. @kofalt can you confirm I'm reading the schema correct?
  - https://github.com/flywheel-io/gears/tree/master/spec

I initially thought it to be good to update those and other gears to support this PR as-is, as those changes would align with the gear spec, and increase their robustness. After recognizing the pattern for how config.json is used, I no longer think merging this before #961 would be prudent.

kofalt:
No, all default config values are required. It (has generally been) expected that the submitter to jobs-add fill in any default values, for example.

api/jobs/handlers.py
```diff
@@ -335,7 +335,7 @@ def get_config(self, _id):
         # Detect if config is old- or new-style.
         # TODO: remove this logic with a DB upgrade, ref database.py's reserved upgrade section.
 
-        if c.get('config') is not None and c.get('inputs') is not None:
+        if 'config' in c and c.get('inputs') is not None:
```
kofalt: Can you explain the rationale behind this change? This would fail if `c['config']` is None.

api/jobs/rules.py
```diff
@@ -237,9 +238,10 @@ def create_jobs(db, container_before, container_after, container_type):
     spawned_jobs = []
 
     for pj in potential_jobs:
-        pj['job'].insert()
-        spawned_jobs.append(pj['rule']['alg'])
+        job_map = pj['job'].map()
+        Queue.enqueue_job(job_map, None) # passing no origin results in system origin
```
kofalt: Given this usage, what would your opinion be of making `origin` a named parameter, and always naming it when calling?

nagem: It is more important we set and origin when we can, this is the only situation I am aware of that does not set an origin. I worry if the origin is optional, it will not be set when it should.

api/jobs/handlers.py
nagem: It was failing when config existed but was not set. AKA, it was the "new" style, but there was no config for the job.

api/jobs/rules.py
kofalt: Out of band: the `None` here is really the problem, this will be replaced with a system default origin to make the call clearer.

api/jobs/rules.py
```diff
@@ -236,10 +237,14 @@ def create_jobs(db, container_before, container_after, container_type):
...
     spawned_jobs = []
+    origin ={
+        'type': str(Origin.system),
+        'id': None
+    }
 
     for pj in potential_jobs:
         job_map = pj['job'].map()
-        Queue.enqueue_job(job_map, None) # passing no origin results in system origin
+        Queue.enqueue_job(job_map, origin) # passing no origin results in system origin
```
kofalt: Should remove that comment now

ryansanford:
LGTM. Nice work!

Confirmed following
- Above documented gear rule still works as intended.
- Existing dcm2niix and mriqc gears work as-is.
- Batch works as expected for this same group of gears.
- File context and meta still passed successfully to Analysis jobs.

nagem:
Removed unneeded comment after change, will merge when CI is green.

Issue #844, opened 2017-07-07
Is it possible to remove config dupe key race?


kofalt:
I'm seeing [transient CI failures](https://circleci.com/gh/flywheel-io/sdk/157) with error:

```
(500) E11000 duplicate key error collection: scitran.singletons index: _id_ dup key: { : "config" }
```

Here's [what I'm using](https://github.com/flywheel-io/sdk/blob/bd85a1b7ead2bca6be3523dfe54081fb8437b54b/.circleci/config.yml#L22-L28) to boot scitran, from an utterly clean slate:

```
cd /var/scitran

python /var/scitran/code/api/bin/database.py upgrade_schema

/var/scitran/uwsgi-entrypoint.sh uwsgi --ini /var/scitran/config/uwsgi-config.ini --socket "[::]:9000" --processes 4 --threads 2
```

Do you think there's a way to fix this transient failure?
Possible relation to #212?

gsfr:
Looks like we should be doing certain init tasks before uWSGI forks. Colton seemed to think that was possible. That's all I know.

Apologies for the utterly useless comment.

kofalt:
Cool, that's somewhere to look then!

kofalt:
Just a note that I saw this [again](https://circleci.com/gh/flywheel-io/sdk/168) today, with a new key:

```
'(500) E11000 duplicate key error collection: scitran.groups index: _id_ dup key: { : "unknown" }'

...

'{"status_code": 500, "message": "E11000 duplicate key error collection: scitran.singletons index: _id_ dup key: { : \"config\" }"}'
```

I thought `_id_ dup key: { : "unknown" }` was a little interesting...

kofalt:
This ain't fixed

PR #966, opened 2017-10-25
Mongo filetypes

### LINKED ISSUE #955: Move file types to mongo container, use regex for matching
For all files that are added to the system without the user specifying a type, the API attempts to assign the file a type using the extension. Currently, this mapping of extension -> type is hardcoded in the source code here: https://github.com/scitran/core/blob/master/api/filetypes.json. This is no longer sustainable for a few reasons: 
 
  - Changes/additions are tied to releases of core
  - Extensions alone are not enough to properly type files (ex. efile and pfile have the same extension but can be differentiated by beginning with "E" or "P")
  - All sites must use the same mapping

**Proposal:** 
  - [x] Add new mongo collection `filetypes` that stores the file type as the `_id` and a regular expression that matches on files that should be 
  - [x] Add API endpoints that would support clients adding/replacing and removing file types
     - Since these are simple key-value, modifying functionality seems overkill, a singular insert or replace endpoint probably fine. 
  - [x] Update util method to use new file type matching 
  - [x] Remove `filetypes.json`, convert existing mappings to objects that can be inserted into mongo
as defaults

**Concerns:** 
  - If a filename matches multiple types, we might want to think about ordering the types by precedence

davidfarkas:
Resolves #955
Resolves part of #844 specifically the 'unknown' duplicated key issue in the initialize_db method

### Review Checklist


kofalt:
Partial review. It looks like the regexes are largely all of the form "file ends with dot-patttern", so it might be nicer to have a function generate those to make it a little more readable, but up to you.

I'm lukewarm on doing a JS query scan on every file upload - maybe we should just pull the table in and loop ourselves? Or use a singleton doc, but then modifying the doc is hard. I defer to you and @nagem.

api/dao/dbutil.py
```diff
@@ -4,6 +4,7 @@
 from pymongo.errors import DuplicateKeyError
 from . import APIStorageException
 
+
```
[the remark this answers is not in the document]
davidfarkas: I just followed pep8. It says: Surround top-level function and class definitions with two blank lines.

api/handlers/filetypehandler.py
```diff
@@ -0,0 +1,38 @@
...
+    def get(self):
+        """Get file types"""
+        return config.db.filetypes.find()
+
+    def post(self):
+        """
+        Insert or replace a file type. Required fields: '_id' and 'regex' where the '_id' is the unique name of
+        the file type and 'regex' is a regular expression which is used to figure out the file type from the file name.
+        """
+        permchecker = userauth.default(self)
```
nagem: Some of the container types have permissions checking decorators written for them, the one you used is for the User container. For these endpoints, I would expect any site admin (a user whose user doc has `root=true`) to be able to add/change/remove and any logged-in user to be able to get the list of filetypes. For these kinds of checks, there is are decorators [here](https://github.com/scitran/core/blob/master/api/auth/__init__.py#L53) you can use to `require_admin` and `require_login`. [Here](https://github.com/scitran/core/blob/master/api/handlers/devicehandler.py#L54) is an example of one being used.

Note: the difference between `superuser_request` and `user_is_admin` is currently different but should be combined. The first is old behavior that called out you were making a superuser request via a query param. The second is notes the privileges you inherently get on all requests because your user doc distinguishes you as a site admin.

api/handlers/filetypehandler.py
```diff
@@ -0,0 +1,38 @@
...
+        payload = self.request.json_body
+        mongo_schema_uri = validators.schema_uri('mongo', 'filetype.json')
+        mongo_validator = validators.decorator_from_schema_path(mongo_schema_uri)
+        mongo_validator(permchecker(noop))('PUT', payload=payload)
```
nagem: We can also use a mongo schema validator (although we've kind of stopped that practice as we didn't see many real world situations where it prevented dev error), but this is better labeled as an input validator as it's using a json schema to validate request bodies from clients. Rather than creating a decorator*, you can call the validator directly ([here](https://github.com/scitran/core/blob/master/api/jobs/handlers.py#L230) is an example). 

Related, when you call the existing validator with `PUT`, it drops the `required` piece of the json schema to allow for patch updates. `POST` does not. 

\* This decorator practice was created in the old rewrite of containerhandler to facilitate calling the permissions check, validators and storage function in one line. It works _alright_ in that handler for now, but new functionality does not need to follow that convention.

bootstrap.sample.json
```diff
@@ -25,5 +25,34 @@
       "_id": "local",
       "type": "engine"
     }
+  ],
+  "filetypes": [
+    { "_id": "bval", "regex": ".*\\.(bval$|bvals$)" },
```
ambrussimon: This implementation might identify `x.dcm.zip` as an `archive` (instead of as a `dicom`) depending on the order of iteration over filetypes/regexes. This was working in the previously because we were using an exact match for *after the first dot* [like so](https://github.com/scitran/core/blob/master/api/files.py#L163).

Let's remove `.*` from the beginning of regexes (they get even simpler) and switch to `re.search()` instead of `re.match()`. If multiple matches are found, one could still infer that the best match is the longest one. That way we can still get the added flexibility of regexes (eg. for matching EFiles) but don't lose existing functionality.

ambrussimon (changes requested):

bootstrap.sample.json
kofalt: Nice catch!

I wonder if there's an algorithm that can tell us if a set of regexes are mutually exclusive :thinking:

ambrussimon (approved):

nagem:
Thanks for making the requested changes, @davidfarkas! Looks good to me. I'll let @ryansanford confirm the file type loading works with what he was expecting and then it's good to merge.

ryansanford:
@gsfr Looking for the stylized file type names

gsfr:
I've added the stylized types from the classification branch and also added a few new types.

Does this implementation still prefer longer matches over shorter ones (i.e., .dicom.zip over .zip)?

I used square brackets in some of the regular expressions. Hope that works.

This branch needs to be rebased onto master to resolve the Django related CI error.

bootstrap.sample.json
```diff
@@ -25,5 +25,44 @@
...
+    { "_id": "PDF",             "regex": "\\.pdf$" },
+    { "_id": "Plain Text",      "regex": "\\.txt$" },
+    { "_id": "Presentation",    "regex": "\\.(ppt|pptx)$" },
+    { "_id": "Source code",     "regex": "\\.(c|py|cpp|js|m|json|java|php|css|toml|yaml|yml)$" },
+    { "_id": "Spreadsheet",     "regex": "\\.(xls|xlsx)$" },
+    { "_id": "TOML",            "regex": "\\.toml$" },
+    { "_id": "Tabular data",    "regex": "\\.([ct]sv\\.gz|[ct]sv)$" },
+    { "_id": "Video",           "regex": "\\.(mpeg|mpg|mov|mp4|m4v|mts)$" },
+    { "_id": "XML",             "regex": "\\.xml$" },
+    { "_id": "YAML",            "regex": "\\.(yaml|yml)$" }
```
kofalt: Megan alerted me to this change: I don't think we can do this since it would break the gear surface area. Suggest using the original names in the `filetypes.json` file

gsfr: @kofalt Thanks for the heads up. Could you please elaborate? What are all the aspects you see impacted by this change?

Anything beyond this?
- massive DB migration
- manifest updates
- all gears updated at all sites

kofalt: So, my understanding because these filetypes are presented to gears, and they might rely on that field, a check for `== "MATLAB data"` would fail if the string is now `"MATLAB Data"`.

This is a good example of how we don't quite have the full gear "surface area" mapped out and committed to: with our quickly-built deep hierarchy integration, there's more that could make a gear fail through no fault of its own.

So our options (I think) are:

1. Not change existing strings
1. Change the strings, but have the old versions for when they're provided externally (?)
1. Triage the full set of gears & gear authors, communicate a breaking change

ryansanford:
Final review pending DB change by @nagem

nagem:
Adding breaking change label as this PR contains a DB update that retypes files that exist in the system.

kofalt:
@gsfr Did you ever make a decision regarding my (collapsed) review comment above?

I'm fine with us either resolving that, or shipping this to customers after we use our TBD new customer-notification strategy.

nagem:
@ryansanford @gsfr @davidfarkas @kofalt 

I added the database update that will retype files (and insert the starting state of the filetypes collection for existing customers). Currently the only thing preventing this from landing is resolving any conflicts with existing gears and properly educating existing users.

gsfr:
Thanks for the ping, @kofalt.

Properly defining the _gear surface area_ seems like a fairly urgent matter. Gear authors must know what they can rely on and what not. We need to make that explicit, rather than retroactively holding ourselves to standards we never knowingly committed to. So, you certainly have good point!

I would like to go ahead with the current change, which can be rolled out to most sites without much communication, as determined by ops. Any site in question, needs to be properly contacted and advised, again, as determined by ops. Cc @tcbtcb.

@lmperry @jenreiter @kjamison Please advise on FW-internal gear impact.

kofalt:
I'd like to communicate with every site, just to get in the habit. But this plan SGTM.

kofalt (approved):

gsfr (approved):

kjamison:
I'm pinged on this but have no idea what this change actually means, can someone TL;DR?  As far as gears, is this just a change in the meta-data for input files?

nagem:
@kjamison the stylization of a file's type available on the inputs map will change. For example, 

```
"config" : {
        "inputs" : {
            "nifti" : {
                "hierarchy" : {...},
                "base" : "file",
                "location" : {...},
                "object" : {
                    "info" : {...},
                    "mimetype" : "application/octet-stream",
                    "tags" : [],
                    "measurements" : [],
                    "type" : "nifti",  ## This key will have a different stylization, after this update it will be "NIfTI"
                    "modality" : null,
                    "size" : 389164
                }
            }
        },
        "destination" : {...},
        "config" : {...}
    }
``` 

Any gears that reference this key in a way that would be broken by a different string stylization (or in some situations, a different type will be applied (Gunnar knows more if you thinks this applies to you)) should be updated. Anyone else can chime in if they have information that better applies to gear authors, I'm responding with API developer context.

gsfr:
@kjamison If you are not relying on specific file type or classification metadata, you should be unaffected. Thanks!

kofalt:
Breaking change: the `type` key on files will have a new format. Some types will have changed, some have been broken up, and some new types have been added. In addition, it will not be possible to add custom types on a per-installation basis.

If you are using one of the types below that has unambiguously changed, you can prepare for the change by adding a function such as this one to your code:

```python
def updated_type(old_type):
	"""
	Updates an old file type to the new format, if it matches.
	"""
	
	return (
		old_type 
		.replace('archive',       'Archive')
		.replace('bval',          'BVAL')
		.replace('bvec',          'BVEC')
		.replace('dicom',         'DICOM')
		.replace('document',      'Document')
		.replace('eeg',           'EEG')
		.replace('gephysio',      'GE Physio')
		.replace('image',         'Image')
		.replace('ismrmrd',       'HDF5')
		.replace('log',           'Log')
		.replace('markdown',      'Markdown')
		.replace('MATLAB data',   'MATLAB Data')
		.replace('MGH data',      'MGH Data')
		.replace('nifti',         'NIfTI')
		.replace('parrec',        'PAR/REC')
		.replace('pdf',           'PDF')
		.replace('pfile',         'PFile')
		.replace('presentation',  'Presentation')
		.replace('PsychoPy data', 'PsychoPy Data')
		.replace('qa',            'QC')
		.replace('spreadsheet',   'Spreadsheet')
		.replace('tabular data',  'Tabular Data')
		.replace('text',          'Plain Text')
		.replace('video',         'Video')
	)
```

Then, change your logic as shown:

```python

# Old version
if filetype == 'nifti':
	run_script()

# New version
if updated_type(filetype) == 'NIfTI':
	run_script()
```

This will make your code ready for the breaking change.
The `updated_type` function can be removed later. 

A full list of the changes are as follows:


**Unambiguously changed:**

Old          | New
------------ | ----------
archive      | Archive
bval         | BVAL
bvec         | BVEC
dicom        | DICOM
document     | Document
eeg          | EEG
gephysio     | GE Physio
image        | Image
ismrmrd      | HDF5
log          | Log
markdown     | Markdown
MATLAB data  | MATLAB Data
MGH data     | MGH Data
nifti        | NIfTI
parrec       | PAR/REC
pdf          | PDF
pfile        | PFile
presentation | Presentation
PsychoPy data| PsychoPy Data
qa           | QC
spreadsheet  | Spreadsheet
tabular data | Tabular Data
text         | Plain Text
video        | Video

**Ambiguously changed:**

Old          | New
------------ | ----------
markup       | HTML or XML
source code  | C/C++ or CSS or Java or JSON or MATLAB or Python or PHP or JavaScript or TOMLor YAML

**New types:**

Old          | New
------------ | ----------
_n/a_        | Audio 
_n/a_        | EFile 
_n/a_        | Jupyter 
_n/a_        | PFile Header

ehlertjd:
One potential concern with using replace in the python method - if, for example, I added and used a custom type called `text/csv` and they run that function, won't it incorrectly map into `Plain Text/csv`?

kofalt:
Totally - I was considering using regexes but I figured this version would be easier to read. The `type` key is not an enum, but in practice the range of values are pretty small. I am also only suggesting this in-place change for comparison only, not to send back to the DB.

I would accept an alternative that added `^$` to everything to make it more correct.

kofalt:
A quick note: some gear rules may be broken by this change, if any of the types were referenced that were not merely case-changed. We should probably address that in this branch.

gsfr:
Thanks, @kofalt. Not sure we should include code, especially, if it's not bullet proof. The change users need to make is likely trivial compared to the update function.

kofalt:
Well, the trick is that they need to have their changes ready before this break ships, right? So there needs to be a way for a script to work with the old & new versions simultaneously. Unless I'm missing something...

gsfr:
Theoretically, you are right, of course. Practically, I continue to think it won't be a problem.

We may, however, want to use a forward compatible change like that in our own gears. But I still haven't heard that any of our own gears are actually affected.

kofalt:
@gsfr At minimum one gear that Jen is working on is affected. I've been keeping her up to date on this change to help with that. I'm going to assume that any gear break _could_ affect anyone going forward.

ryansanford:
Additional upgrade needs:
- Update existing gear rules
- Update existing session templates
- Reconcile how pending jobs locked into gear IDs at time of upgrade will result in properly typed files.

PR #973, opened 2017-10-30
Analysis downloaded files separated into input and output folders

### LINKED ISSUE #971: Place input and output analysis files into separate folders on download
When downloading an analysis, place files marked with `input: true` in a folder labeled `input` and files marked with `output: true` in a folder labeled `output`.

hkethi002:
Fixes #971 

### Breaking Changes
Analysis downloads will have another folder level in the filepath: `input` and `output`.

### Review Checklist

### DIFF
--- a/api/handlers/refererhandler.py
+++ b/api/handlers/refererhandler.py
@@ -378,7 +378,7 @@ class AnalysesHandler(RefererHandler):
             filepath = os.path.join(data_path, util.path_from_hash(f['hash']))
             if os.path.exists(filepath): # silently skip missing files
                 targets.append((filepath,
-                                util.sanitize_string_to_filename(analysis['label']) + '/' + 'input' if f.get('input') else 'output' + '/'+ f['name'],
+                                util.sanitize_string_to_filename(analysis['label']) + '/' + ('input' if f.get('input') else 'output') + '/'+ f['name'],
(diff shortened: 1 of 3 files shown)


api/handlers/refererhandler.py
```diff
@@ -378,7 +383,7 @@ def _prepare_batch(self, fileinfo, analysis):
             filepath = os.path.join(data_path, util.path_from_hash(f['hash']))
             if os.path.exists(filepath): # silently skip missing files
                 targets.append((filepath,
-                                util.sanitize_string_to_filename(analysis['label']) + '/' + f['name'],
+                                util.sanitize_string_to_filename(analysis['label']) + '/' + self._input_or_output(f) + '/'+ f['name'],
```
nagem: It's not a huge deal, but this can be written inline as 
`util.sanitize_string_to_filename(analysis['label']) + '/' + 'input' if f.get('input') else 'output' + '/'+ f['name'],`

nagem:
I changed it to use a ternary inline, but changes LGTM. I'll merge when CI is green.

nagem:
@kofalt: This is a breaking change to the output of an analysis download. Let me know if the CLI or SDK use the result of an analysis download and would be affected by the change in the file path to the input and output files.

PR #975, opened 2017-10-30
Correctly mark update 38

nagem:
@ryansanford :
If a site was on db version 37 when an update containing the second `upgrade_to_37()` was present, no updates would take place. 

If a site was on 36 or before, the original `upgrade_to_37()` would not be run, only the second (API key upgrade). A proper call to "upgrade from" should be used in these instances. 

### Review Checklist

### DIFF
--- a/bin/database.py
+++ b/bin/database.py
@@ -22,7 +22,7 @@ from api.jobs import gears
 from api.types import Origin
 from api.jobs import batch
 
-CURRENT_DATABASE_VERSION = 37 # An int that is bumped when a new schema change is made
+CURRENT_DATABASE_VERSION = 38 # An int that is bumped when a new schema change is made
 
 def get_db_version():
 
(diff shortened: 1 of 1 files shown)


bin/database.py
```diff
@@ -22,7 +22,7 @@
 from api.types import Origin
 from api.jobs import batch
 
-CURRENT_DATABASE_VERSION = 37 # An int that is bumped when a new schema change is made
+CURRENT_DATABASE_VERSION = 38 # An int that is bumped when a new schema change is made
```
kofalt: Just one question: Do we need to make this 39, or are the ops team just going to backrev this integer?

ryansanford:
@nagem 
- Is there a programatic way to know with certainty which situation a site is in? Which `upgrade_to_37()` was run?
- For either situation, is there a risk after this update that API keys would be lost? Someone generated their api key so it exists in new collection, then this update runs, and replaces it in new location with the old or non-existant value in the old location?

nagem:
1) There is no harm in running both 37 and 38, on a site that receive none, one, or both of those upgrades. But if you wanted to know for other reasons:

```
# Will return items if 37 was not run
db.groups.find({'permissions.site': {'$exists': True}})
```

```
# Will return items if 38 was not run 
# Will return nothing if there are no API keys on the site (false positive)
db.users.find({'api_key': {'$exists': True }})
```

2) I rewrote the update to only migrate the user's API key if they do not already have one in the new location. So if a user regenerated their API key during the mishandled time period, they would key their new key and their old key would be removed. If they did not regenerate an API key, their old key will be migrated.

bin/database.py
nagem: In the PR description I mentioned you can use the `--from` flag to upgrade from a specific version. Rerunning 37 and 38* if either had been run before will not cause issue. Upgrades are intended to be written in a way that they can be rerun, usually in a way that they return an empty cursor and move on.

ryansanford:
Sounds great @nagem . I'm a LGTM based upon your responses. I haven't reviewed the implementation.

bin/database.py
kofalt: Thanks Megan.

kofalt: I don't see that on this PR ticket, I'm guessing you meant the prior PR. SGM though.

PR #976, opened 2017-10-30
Enable individual tests

ambrussimon:
Enable build/lint/unit/integ skip and py.test args for individual test running.

### Review Checklist


test/bin/run-tests-osx.sh
```diff
@@ -1,65 +0,0 @@
-#!/usr/bin/env bash
```
nagem: 🕯

nagem:
Everything works great, just one additional request: can we add a command line arg to turn off aboa tests? For example, if I want to run one test and only one test, I would make this call:

```
./run-tests-docker.sh -B -- -U -L -- -k test_batch
  -B (no rebuild of the container
  -U (skip unit tests)
  -L (skip linting)
  -k test_batch (run only tests that match "test_batch")
```

It goes immediately into running that one individual test, but when it completes it runs all 100+ aboa tests.

test/bin/run-tests-docker.sh
```diff
@@ -1,85 +1,81 @@
...
+        -v $(pwd):/var/scitran/code/api \
+        --entrypoint bash \
+        scitran-core:run-tests \
+        /var/scitran/code/api/test/bin/run-tests-ubuntu.sh \
+        $TEST_ARGS
+}
+
+
+function clean_up() {
+    export TEST_RESULT_CODE=$?
```
ryansanford: Is `export` necessary here?

test/bin/run-tests-ubuntu.sh
```diff
@@ -1,76 +1,165 @@
...
+        # relative $ref's in the jsonschema's
+        pushd raml/schemas/definitions
+        NODE_PATH="$integration_test_node_modules" abao ../../api.raml "--server=$SCITRAN_SITE_API_URL" "--hookfiles=../../../test/integration_tests/abao/abao_test_hooks.js"
+        popd
+    fi
+}
+
+
+function clean_up () {
+    export TEST_RESULT_CODE=$?
```
ryansanford: Is `export` necessary here?

test/bin/run-tests-ubuntu.sh
```diff
@@ -1,76 +1,165 @@
...
+    set +e
+
+    echo
+    echo "Test return code = $TEST_RESULT_CODE"
+
+    if [[ -n "${API_PID:-}" ]]; then
+        # Killing uwsgi
+        kill $API_PID || true
```
ryansanford: What's the purpose of `|| true`? It would be obvious to me if exit on error was still set.

test/bin/run-tests-ubuntu.sh
```diff
@@ -1,76 +1,165 @@
...
+        wait 2> /dev/null
+    fi
+
+    if [[ "${TEST_RESULT_CODE}" == "0" ]]; then
```
ryansanford: This block should not get called if only `-h` flag is passed to this script.

ryansanford:
@ambrussimon I'm done reviewing. Nice work.

kofalt:
Heh, given @nagem's request it might make sense to have flags turn things on, rather than off...

ambrussimon:
Thanks all for the help, I think it's coming together really nicely with the fixes @ryansanford, the abao separation @nagem and switching from opt-out to opt-in @kofalt.

If we need more speed than this, I still have a couple ideas (jumping into a shell with direct access to py.test or similar), but I think this should be sufficient for quite a while.

nagem:
I like the default of running all tests if no options are specified 👍

PR #978, opened 2017-10-31
Removed a few dozen unhelpful log statements

kofalt:
This is driving me a little batty and making the logs hard to read, so I fixed it.

Every `log.debug` that doesn't have some information attached to it has been removed, excepting some in `config.py` because... fine. Oh, and I removed the every-request line that benchmarks our database!

If one of these is important to you, please add it back with more information - and make it conditional.

When I decided to improve this, there were about ~40 calls to `log.debug`, many of them in code paths that tend to be called on many / most / all requests. Now there are about 14. I also got rid of one `log.info`, the rest seemed fine.


kofalt:
You're a real piece of work, codecov.

nagem:
LGTM, these should definitely be removed.

PR #979, opened 2017-10-31
Correct match_ids in containerhandler.get_jobs

### LINKED ISSUE #972: non-analysis jobs are not returned from /api/sessions/{session_id}/jobs?join=gears
Seems to impact all non-analyses jobs, both old and new.

Suspect this was caused by PR https://github.com/scitran/core/pull/967 ?

If behavior of this endpoint is dependent on analysis vs non-analysis, both cases should be checked via python integration tests.

kofalt:
Resolves #972

Yeah, really dumb error here. Late night bug fixes :couch_and_lamp: :tumbler_glass:

### DIFF
--- a/api/handlers/containerhandler.py
+++ b/api/handlers/containerhandler.py
@@ -271,9 +271,9 @@ class ContainerHandler(base.RequestHandler):
         # Stateful closure to remove search duplicates
         # Eventually, this code should not call Queue.search and should instead do its own work.
         def match_ids(x):
-            should_add = str(x['_id']) in match_ids.unique_job_ids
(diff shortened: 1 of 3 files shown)


nagem:
Changes LGTM.

Please add a test to ensure utility and analysis jobs are properly returned via this endpoint. The code is covered, probably just need to modify an existing test to ensure utility and analysis jobs are returned.

ryansanford:
I confirmed this is resolving #972 on my local dev. Thanks @kofalt . I added that issue to the PR description. Assigning back to you for the requested test improvement.

ryansanford:
LGTM

PR #985, opened 2017-11-08
Rule regex match

### LINKED ISSUE #982: Allow regex matching in rules
Extend matches (individual entries in a rule's `all` and/or `any` array) with an optional `regex` boolean field. When this field is present and set to true, use a regular expression match (implicit ^...$) instead of an exact match (or instead of glob match in case of file.name).

ambrussimon:
Closes #982 

Force-updated, stole schema refactor from rule-config branch that's under further discussion.
(-config+regex property)

### Review Checklist


nagem:
Changes LGTM! Thanks for rebasing this on master.

ryansanford:
Thanks for posting this @ambrussimon . Can you summarize for @dpuccetti how to take advantage of this change?

ambrussimon:
@dpuccetti Having added `regex` as a [new optional boolean field](https://github.com/scitran/core/pull/985/files#diff-75791526e7019bbcf1bd5159958cb48bR20), it can be used in rule POST/PUT requests. I just extended the RAML docs a bit to include that, though it's [not too helpful](https://scitran.github.io/core/?branch=rule-regex-match#/docs/resources-19-resources-1-resources-9-methods-1). I believe a label saying regex and a checkbox (or smg similar) for both any/all matches would work nicely with this feature.

`regex = true` can be applied for all available match types and changes the matching behavior *from exact* (or file globs in case of file.name) *to regular expression matching* (case insensitive, full match - eg. uses implicit ^...$)

nagem:
Tested with the following rule:
```
{
    "_id" : ObjectId("5a00ac22b3bcb1001afc3a19"),
    "all" : [ 
        {
            "type" : "file.measurements",
            "value" : "(localizer|diffusion)",
            "regex" : true
        }
    ],
    "name" : "Run dcm2niix on dicom",
    "alg" : "dcm2niix",
    "project_id" : "5a00ac22b3bcb1001afc3a18",
    "any" : []
}
```

Uploaded two new dicoms to the system using the web uploader: one that would get classified as a `Diffusion`, and one that would get classified as `Anatomy_inplane`. Classifier ran on both, but only the one classified as a `Diffusion` had a `dcm2niix` job scheduled.

LGTM, feel free to merge. @ryansanford

PR #988, opened 2017-11-09
Add endpoint for modifying container info fields

nagem:
Similar to #893, add endpoint to modify container info using set/delete/replace keywords.

### `POST /api/<container>/<container_id>/info`
  
Format of request:
```
# set or delete required, can have both
{
	"set": {}, 
	"delete": []
}
```
  - "set" is a non-empty map of `key` : `value` (values may be scalars or complex objects). All keys are added to the `info` key if they do not exist, or are replaced by the `value` if they do exist
  - "delete" is a non-empty list of top level keys to be deleted. If key does not exist, request is ignored.

OR

```
{
	"replace": {}
}
```
  - the value of the "replace" key will be set as the value of the `info` key for the file. It can be an empty map. 

**Example Requests**
```
POST /api/acquisitions/5984d10adf01e2001433dde8/info HTTP/1.1

{
	"set": {
		"top_level_key": 123,
		"complex_object_key": {
			"a": 1,
			"b": 2
		}
	},
	"delete": ["keys", "to", "be", "removed"]
}
```

```
POST /api/acquisitions/5984d10adf01e2001433dde8/info HTTP/1.1

{
	"replace": {
		"notice": "this is a full replace of the info field"
	}
}
```

### Breaking Changes
None

There will be a transition period where setting `info` through `PUT /container_name/container_id`will still be allowed. It is expected that eventually all changes in `info` should go through these new endpoints. 

@kofalt SDK related additive changes

### Review Checklist

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -256,6 +256,9 @@ endpoints = [
         prefix('/<cont_name:{cname}>', [
             route( '/<cid:{cid}>',     ContainerHandler,                                m=['GET','PUT','DELETE']),
             prefix('/<cid:{cid}>', [
+
+                route( '/info',                                ContainerHandler, h='modify_info',    m=['POST']),
+
                 route('/<list_name:tags>',               TagsListHandler, m=['POST']),
                 route('/<list_name:tags>/<value:{tag}>', TagsListHandler, m=['GET', 'PUT', 'DELETE']),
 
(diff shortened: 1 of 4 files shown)


nagem:
Tests are added. Feel free to start SDK development at any time @kofalt

nagem:
Rebased, merging when CI is green.

PR #992, opened 2017-11-14
NGINX Unit

gsfr:
This is work in progress. Help wanted.

Scope proposal:
- Ability to run core's own test suite on NGINX Unit and Alpine Linux.
- Ability to trivially bring up a fully functional core, including mongo, to be used for testing of other components, such as reaper or cli.

Comments and/or requirement definitions welcome.

TO DO
- [ ] Revert `Have Travis push to Docker Hub for this branch`


gsfr:
@ambrussimon I would like to use the setup.py/requirements.txt pattern that you have established in reaper, but I don't know how to handle the necessary `dependency_links`. Any ideas?

requirements.txt
```diff
@@ -1,17 +0,0 @@
...
-python-dateutil==2.4.2
-pytz==2015.7
-requests==2.9.1
-rfc3987==1.3.4
-strict-rfc3339==0.7
-unicodecsv==0.9.0
-uwsgi==2.0.13.1
-webapp2==2.5.2
-WebOb==1.5.1
-git+https://github.com/flywheel-io/gears.git@v0.1.4#egg=gears
```
kofalt: I use this file for [CI](https://github.com/flywheel-io/sdk/blob/3e5dd6ff31b6edc652215753f7865151d2f44b59/.circleci/config.yml#L31), so I would need an equivalent command. I'll want to try things out with this branch if it gets closer to merging.

Dockerfile
```diff
@@ -1,104 +1,42 @@
...
-CMD ["uwsgi", "--ini", "/var/scitran/config/uwsgi-config.ini", "--http", "0.0.0.0:8080", "--http-keepalive", "--so-keepalive", "--add-header", "Connection: Keep-Alive" ]
+FROM python:2.7-alpine3.6 as build
+
+RUN apk add --no-cache build-base curl
+
+WORKDIR /src/nginx-unit
+
+RUN curl -L https://github.com/nginx/unit/archive/master.tar.gz | tar xz --strip-components 1
+RUN ./configure --prefix=/usr/local --modules=lib --state=/var/local/unit --pid=/var/unit.pid --log=/var/log/unit.log \
+ && ./configure python \
```
kofalt: We should check the configure options, nginx-web (or whatever) has quite a few good ones.

setup.py
```diff
@@ -0,0 +1,36 @@
...
+    'strict-rfc3339==0.7',
+    'unicodecsv==0.9.0',
+    'webapp2==2.5.2',
+    'WebOb==1.5.1',
+    'gears',
+]
+
+dependency_links = [
+    'git+https://github.com/flywheel-io/gears.git@v0.1.4#egg=gears',
+]
```
kofalt: I take it this entry can't be in the above array?

kofalt (commented):
Looks interesting!

Do you have any benchmarks or observed behavior?

Dockerfile
gsfr: On it.

setup.py
gsfr: Not that I could tell, but @ambrussimon will look further into it.

requirements.txt
gsfr: We will make sure that you don't need any of that anymore. You should able to simply stand up a core and use it. No internal knowledge required.

gsfr:
@kofalt No benchmarks yet. Still at the 'getting it running' stage. We'll let you know when this is ready to be tested against.

requirements.txt
kofalt: Sounds good, as long as the container checks itself on boot. Frequently I want to CI against a branch whose container (incl updated libraries) has not built yet.

gsfr: That's a good point. We'll need to have images available for arbitrary branches of core to facilitate testing of interacting repos, such as reaper and cli. We typically build images for master and all tags. I think that's enough (rather than building for all branches) and keeps clutter at bay. Would that work for you, @kofalt?

kofalt: Discussed this out of band - basically, as long as any CI container can update itself on boot to the tip of a branch (incl updated libraries) we're fine.

ryansanford:
@gsfr Isn't the definition of this PR being ready that it's not a "Breaking Change" ?

gsfr:
@ryansanford Won't be a breaking chance from the user point of view, but will be for parts of our infra. Removing tag.

nagem:
>Remove abao

🎉🎉🎉🎉🎉🎉🎉🎉🎉

requirements.txt
```diff
@@ -1,16 +1,16 @@
 django>=1.11.0,<1.12.0
 elasticsearch==5.3.0
 enum34==1.1.6
+git+https://github.com/flywheel-io/gears.git@v0.1.4#egg=gears
+ipython
```
gsfr: I'm guessing you want IPython for debugging, but it's such a quick install and I'd rather not bake it into the image.

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
 jsonschema==2.6.0
-Markdown==2.6.5
+markdown==2.6.5
```
gsfr: The reason for the capitalization is to match the output of `pip freeze`, so that it can be copied verbatim.

requirements.txt
gsfr: Just seeing where IPython came from. I'm taking it out.

gsfr:
@ambrussimon nginx/unit#65 has been closed. We should now be able to drop your coverage endpoint workaround. However, it appears that Unit's introduction of HTTP Keep-Alive has broken our tests. The previous commit (https://github.com/nginx/unit/commit/497faf1b9abb188cab40c389a9e6221add5dd496) is the last one that seems to work.

ambrussimon:
@gsfr Just rebased (fixed conflicts), no functional change.

kofalt:
Ref flywheel-io/sdk#80 for current endpoint status!

gsfr:
Thanks, @kofalt.

@ambrussimon Please take a look at the failing SDK tests. Getting those to pass would be a great milestone for this PR.

kofalt:
If Ambrus or anyone else has a look at playing with those, note the docs to help you out:
https://github.com/flywheel-io/sdk#testing

In particular, the `-run` flag and `SdkTestDebug` env variable may be of interest.

gsfr:
@ambrussimon In trying to get flywheel-io/sdk#80 going, I added `upgrade_schema` to `dev+mongo.sh`. Unfortunately, that seems to break some tests. Is it ok/desirable to have this? Can we fix the tests?

PR #995, opened 2017-11-15
Quick fix for logging

### LINKED ISSUE #977: Python filenames must be unique
Our log format mentions a filename:

```
2017-10-31 20:48:18      scitran.api                  base.py   191:DEBU None
```

Unfortunately, we have a `dao/base.py` and a `web/base.py`. Might be others.

kofalt:
Closes #977. I keep running into this, so I hope nobody minds if this rename fixes things for now. Comments on the linked ticket suggest a better long term fix.

This was simply the smallest change, alternatives accepted. I would probably call several files in dao differently, but I didn't want to cause a lot of churn without a better reason to do so.


nagem:
It really doesn't make sense to call the module with the parent class `container` when the module with the child classes is `containerstorage`. I would be more willing to call it `basecontainerstorage`, but that is very long.

kofalt:
Agreed - I would probably have gone with `containerstorage` and then `containerstorage_implementations` or `cs_impls` or something, but then we're renaming 2 files and changing ~30 imports instead of 1 & 3, and that seemed a bit much for code that I don't own... etc.

I'll rename it to `basecontainerstorage` for now then :+1:

kofalt:
Updated & green

nagem:
Cool, go ahead 👍

PR #996, opened 2017-11-15
400 on empty file uid|label|reaper upload

### LINKED ISSUE #993: Disallow empty uploads
Allowing file-less uploads on reaper-focussed endpoints (`/upload/uid`, `/upload/label`, `/upload/reaper`) is unintended and needlessly complicates debugging of reaper tools. I suggest we stop allowing such *empty* uploads.

hkethi002:
Fixes #993 
### Review Checklist

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -18,14 +18,14 @@ from .jobs import rules
 from .jobs.jobs import Job
 from .types import Origin
 from .web import encoder
-
+from .web.errors import FileFormException
 
 class Placer(object):
     """
     Interface for a placer, which knows how to process files and place them where they belong - on disk and database.
     """
 
-    def __init__(self, container_type, container, id_, metadata, timestamp, origin, context, file_fields):
+    def __init__(self, container_type, container, id_, metadata, timestamp, origin, context):
         self.container_type = container_type
         self.container      = container
         self.id_            = id_
@@ -43,7 +43,6 @@ class Placer(object):
 
         # A list of files that have been saved via save_file() usually returned by finalize()
         self.saved          = []
-        self.file_fields    = file_fields
 
 
     def check(self):
@@ -76,15 +75,7 @@ class Placer(object):
         Helper function that throws unless metadata was provided.
         """
         if self.metadata == None:
-            raise files.FileFormException('Metadata required')
-
-    def requireFiles(self):
-        """
-        Helper function that throws unless metadata was provided.
-        """
-        if not self.file_fields or len(self.file_fields) < 1:
-            raise files.FileFormException('No files selected to upload')
-
+            raise FileFormException('Metadata required')
 
     def save_file(self, field=None, file_attrs=None):
         """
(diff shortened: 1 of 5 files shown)


nagem:
I think we would rather `400` (client messed up) than `500` (we messed up).

api/upload.py
```diff
@@ -95,6 +85,19 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
     # Ref docs from placer.TargetedPlacer for details.
     if strategy == Strategy.targeted and len(file_fields) > 1:
```
hkethi002: Should I switch this to a 400 as well?

api/upload.py
```diff
@@ -95,6 +85,19 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
...
         raise Exception("Targeted uploads can only send one file")
```
hkethi002: Should I switch this to a 400 as well

nagem:
We'd want to change the existing 500s as well. Inside `placer.py` there is a check state, and we have `requireTarget` and `requireMetadata`, would we also want to write one for `requireFile`? Not 100% sure that's the best way forward but would probably be cleaner.

api/placer.py
```diff
@@ -75,7 +77,15 @@ def requireMetadata(self):
         Helper function that throws unless metadata was provided.
         """
         if self.metadata == None:
-            raise Exception('Metadata required')
+            raise files.FileFormException('Metadata required')
+
+    def requireFiles(self):
+        """
+        Helper function that throws unless metadata was provided.
```
nagem: I think you meant this to say "unless no files were uploaded" or similar

api/placer.py
```diff
@@ -114,6 +124,16 @@ class TargetedPlacer(Placer):
     An exception is thrown in upload.process_upload() if you try. This could be fixed by making a better schema.
     """
 
+    # TODO: Change schemas to enabled targeted uploads of more than one file.
+    # Ref docs from placer.TargetedPlacer for details.
+    def requireFiles(self):
```
nagem: Maybe I'm missing it but it doesn't look like this is getting used in the `check()` phase for `TargetedPlacer`

kofalt:
Our of band, Harsha: we can't access `file_fields` that way because it breaks the ability of the Placer interface to incrementally process a metadata stream. Megan and I worked on some alternatives, check with us.

nagem:
@hkethi002 and I went through moving the check back into the `process_upload` path.

api/upload.py
```diff
@@ -95,7 +93,11 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
     # TODO: Change schemas to enabled targeted uploads of more than one file.
     # Ref docs from placer.TargetedPlacer for details.
     if strategy == Strategy.targeted and len(file_fields) > 1:
-        raise Exception("Targeted uploads can only send one file")
+        raise FileFormException("Targeted uploads can only send one file")
+    elif strategy in [Strategy.reaper, Strategy.uidupload, Strategy.labelupload, Strategy.uidmatch] and len(file_fields) < 1:
```
kofalt: If you want to keep the incremental processing window open for these upload strategies, then we cannot access `file_fields` here. The fix is to instead have the check during each strategy's `finalize` calls.

Megan and I discussed a possible fix for the targeted case as well, but probably not as part of this PR.

api/upload.py
```diff
@@ -84,8 +84,6 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
 
     placer_class = strategy.value
     placer = placer_class(container_type, container, id_, metadata, timestamp, origin, context)
-    placer.check()
```
kofalt: This line shouldn't have moved. Otherwise, this PR seems fine.

PR #1001, opened 2017-11-20
Moved exceptions to web/errors.py

### LINKED ISSUE #998: Centralize Exception declarations
- Put all the Exceptions and Errors in web/errors.py
- Add comments describing each Exception and Error and what it means

hkethi002:
Fixes #998 
### Review Checklist

### DIFF
--- a/api/web/errors.py
+++ b/api/web/errors.py
@@ -1,0 +1,62 @@
+
+class APIAuthProviderException(Exception):
+    pass
+
+# Creating mulitple containers with same id
+class APIConflictException(Exception):
+    pass
+
+# For checking database consistency
+class APIConsistencyException(Exception):
+    pass
+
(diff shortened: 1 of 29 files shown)


api/web/errors.py
```diff
@@ -0,0 +1,62 @@
...
+class APIUnknownUserException(Exception):
+    pass
+
+class APIValidationException(Exception):
+    def __init__(self, errors):
+
+        super(APIValidationException, self).__init__('Validation Error.')
+        self.errors = errors
+
+# Payload for a POST or PUT does not match mongo json schema
```
nagem: >json object prepared for storage layer does not match mongo json schema

That would probably be a better description of the error, just to be clear it's been edited since being pulled from the original request payload.

Issue #1005, opened 2017-11-21
Rewrite download logic


nagem:
The logic around creating unique paths for containers during download is difficult to maintain.

PR #1006, opened 2017-11-21
Fix download count for conflict paths

### LINKED ISSUE #1004: Download file path conflict resolution doesn't count past 0
If a session has 4 acquisitions named `effort1`, the download logic properly places the files in the first `effort1` in a folder titled `effort1`, but puts the rest in `effort1_0`, not creating the expected `effort1_1`, `effort1_2`, ...

nagem:
Fixes #1004 

A rewrite would make issues like this take a lot less time to fix (noted in #1005). 

### Review Checklist

### DIFF
--- a/tests/integration_tests/python/test_download.py
+++ b/tests/integration_tests/python/test_download.py
@@ -8,8 +8,10 @@ def test_download(data_builder, file_form, as_admin, api_db):
     project = data_builder.create_project(label='project1')
     session = data_builder.create_session(label='session1')
     session2 = data_builder.create_session(label='session1')
+    session3 = data_builder.create_session(label='session1')
     acquisition = data_builder.create_acquisition(session=session)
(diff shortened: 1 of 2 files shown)


tests/integration_tests/python/test_download.py
```diff
@@ -55,10 +60,13 @@ def test_download(data_builder, file_form, as_admin, api_db):
...
     # Verify a single file in tar with correct file name
     found_second_session = False
+    found_third_session = False
     for tarinfo in tar:
         assert os.path.basename(tarinfo.name) == file_name
         if 'session1_0' in str(tarinfo.name):
             found_second_session = True
+        if 'session1_1' in str(tarinfo.name):
+            found_third_session = True
     assert found_second_session
```
hkethi002: Should found_third_session be tested here after the loop?

nagem: Whoops - nice catch!

PR #1007, opened 2017-11-22
Search size uses filters

### LINKED ISSUE #1003: limit:-1 is failing on reasonable sized data sets
With a query that should return a reasonable size of results:
```
{
    "return_type":"file",
    "filters": [{
        "match":{"session._id":"58f177c11f4a0300130d5fe1"}}],
    "limit":-1
}
```

This error is returned: 
```
TransportError: TransportError(500, u'search_phase_execution_exception', u'Result window is too large, from + size must be less than or equal to: [10000] but was [155984]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting.')
```

It appears as if the initial count that sets the size for the actual query is returning a count of all files rather than a filtered count. Also knowing Elastic uses `max_result_window`, we should fail with a `400 - requested result set too large` (or similar) when that original returned count is larger than the setting.

hkethi002:
Fixes #1003 
### ToDo
 - [x] Tests
### Review Checklist

### DIFF
--- a/api/handlers/dataexplorerhandler.py
+++ b/api/handlers/dataexplorerhandler.py
@@ -483,7 +483,7 @@ class DataExplorerHandler(base.RequestHandler):
         aggs['by_session']['subject.age'] = age_node['subject.age']
         return {'facets': aggs}
 
-    def search_size(self, return_type):
+    def search_size(self, return_type, filter=None):
         body = {
             "size": 0,
             "aggs" : {
@@ -495,6 +495,13 @@ class DataExplorerHandler(base.RequestHandler):
                 }
             }
         }
+
+        if filter:
+            body["query"] = {
+                "bool": {
+                    "filter": filter
+                }
+            }
 
         size = config.es.search(
             index='data_explorer',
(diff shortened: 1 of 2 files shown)


api/handlers/dataexplorerhandler.py
```diff
@@ -496,11 +508,20 @@ def search_size(self, return_type):
...
+                }
+            }
+        try:
+            size = config.es.search(
+                index='data_explorer',
+                doc_type='flywheel',
+                body=body)['aggregations']['count']['value']
+            size = int(size*1.02)
+        except TransportError:
+            self.abort(400, "Reuest to large, filter your request")
```
nagem: A couple things:
1) This wouldn't be the failure point for large requests - it would just return 1 aggregation with a large count (since `size` is set to 0). 
2) A user could request more than 10,000 in their `size` param as well, so I would check size in `_parse_request` after resolving the size param. 
3) I would modify the error to say: 
> Request would return more than 10,000 results. Please add additional filters.

nagem:
Rethinking, `10,000` is just the default setting for `max_result_window`, but I think it's alright to be explicit with 10k. I'll update our config settings to make it explicit.

hkethi002:
Is that a change to flyfly or core?

nagem:
It is an elastic setting that FW will set in it's mongo-connector code.

PR #1015, opened 2017-11-29
Mark failed job output

### LINKED ISSUE #968: Mark files created via a job that fails
We need to clearly distinguish outputs of a failed gear from those of a succeeded gear. Even outputs from a "failed" gear may be valuable or even valid. Without them, debugging would likely be impeded, if not impossible.

Analysis gears
  - Analyses are directly tied to a job, so marking output files might be redundant

Utility Gears
  - For failed utility gears, output files will be added to acquisitions us usual, but those files will be marked in the DB
  - no gear rules will fire on those files until they are accepted (or deleted) by a user

Remaining questions
  - Metadata from failed gears should be handled in a similar fashion, but details TBD.

ambrussimon:
Resolves #968 

Notes/TBDs:
* Didn't add `/jobs/x/complete` endpoint as discussed, added feature to existing `/upload/engine` endpoint in a backwards-compatible way instead. Opinions?
* Without further context it seems like success info might as well be stored on the job. What's the reasoning behind keeping it in a separate ticket?
* Job ticket purging/cleanup not implemented!

### Review Checklist

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -9,9 +9,10 @@ from urlparse import urlparse
 
 from .. import upload
 from .. import util
-from ..auth import require_login, has_access
+from ..auth import require_drone, require_login, has_access
+from ..dao import hierarchy
 from ..dao.containerstorage import ProjectStorage, SessionStorage, AcquisitionStorage
-from ..dao.containerutil import ContainerReference
+from ..dao.containerutil import ContainerReference, pluralize
 from ..web import base
 from ..web.encoder import pseudo_consistent_json_encode
 from ..web.errors import APIPermissionException, APINotFoundException, InputValidationException
@@ -458,7 +459,7 @@ class JobHandler(base.RequestHandler):
         try:
             Job.get(_id)
         except Exception: # pylint: disable=broad-except
-            self.abort(404, 'Job not found')
+            raise APINotFoundException('Job not found')
 
         return Logs.add(_id, doc)
 
@@ -478,6 +479,57 @@ class JobHandler(base.RequestHandler):
 
         new_id = Queue.retry(j, force=True)
         return { "_id": new_id }
+
+
+    @require_drone
+    def prepare_complete(self, _id):
+        try:
+            j = Job.get(_id)
+        except Exception: # pylint: disable=broad-except
+            # TBD maybe raise APINotFound from Job.get?
+            raise APINotFoundException('Job not found')
+        payload = self.request.json
+        ticket = {
+            'job': j.id_,
+            'success': payload.get('success', False),
+        }
+        # TBD why not store success on the job itself?
+        return {'ticket': config.db.job_tickets.insert_one(ticket).inserted_id}
+
+
+    @require_login
+    def accept_failed_output(self, _id):
+        try:
+            j = Job.get(_id)
+        except Exception: # pylint: disable=broad-except
+            raise APINotFoundException('Job not found')
+
+        # Permission check
+        if not self.superuser_request:
+            j.destination.check_access(self.uid, 'rw')
+
+        if j.state != 'failed':
+            self.abort(400, 'Can only accept failed output of a job that failed')
+
+        # TBD recording user/timestamp? logging?
+        # TBD ticket cleanup/purging?
+
+        # Remove flag from files
+        container = j.destination.get()
+        for f in container.get('files'):
+            if f['origin'] == {'type': 'job', 'id': _id}:
+                del f['from_failed_job']
+        collection = pluralize(j.destination.type)
+        query = {'_id': container['_id']}
+        updates = {'$set': {'files': container['files']}}
+        config.db[collection].update_one(query, updates)
+
+        # Apply metadata
+        hierarchy.update_container_hierarchy(j.produced_metadata, container['_id'], j.destination.type)
+
+        j.failed_output_accepted = True
+        j.save()
+
 
 class BatchHandler(base.RequestHandler):
 
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -2,6 +2,7 @@
 API request handlers for the jobs module
 """
 import bson
+import copy
 import os
 import StringIO
 from jsonschema import ValidationError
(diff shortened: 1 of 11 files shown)


kofalt:
To be clear, would the engine call `/prepare-complete`, get a ticket, and then call `/upload/engine` with that ticket ID? In what scenario would someone call `/accept-failed-output`?

api/jobs/handlers.py
```diff
@@ -482,6 +483,57 @@ def retry(self, _id):
         new_id = Queue.retry(j, force=True)
         return { "_id": new_id }
 
+
+    @require_drone
+    def prepare_complete(self, _id):
+        try:
+            j = Job.get(_id)
+        except Exception: # pylint: disable=broad-except
+            # TBD maybe raise APINotFound from Job.get?
```
nagem: Yes I think that would be preferred. Core should hopefully only throw a generic `Exception` when it has no better error to throw, or perhaps via developer error (aka to alert a developer they misused a function).

nagem: That should remove the redundant error checking below as well.

api/jobs/handlers.py
```diff
@@ -482,6 +483,57 @@ def retry(self, _id):
...
+
+        # Permission check
+        if not self.superuser_request:
+            j.destination.check_access(self.uid, 'rw')
+
+        if j.state != 'failed':
+            self.abort(400, 'Can only accept failed output of a job that failed')
+
+        # TBD recording user/timestamp? logging?
+        # TBD ticket cleanup/purging?
```
nagem: We could always place a ttl index on the mongo collection.

ambrussimon:
@kofalt Correct. Formalizing just a bit more (current implementation):
* `POST /jobs/x/prepare-complete {"success": true}` => create & return `{"ticket": "y"}`
* `POST /upload/engine?level=L&id=ID&job_id=x&job_ticket_id=y` FileForm

On `/accept-failed-output`: my naive assumption was that there might be algs that have meta/files in their failure output that users still would like to store/further analyze. This shouldn't be the default though, hence the marking, but leaving the option.

Another TBD I forgot about was whether to require the `job_ticket_id` URL param at all during `/engine/upload`, since I can find the relevant ticket just from the job_id... But then we are ever closer to just storing the success/failure on the job itself.

kofalt:
Is there a purpose for the  `level`, `id`, and `job_id` fields if there is already `job_ticket_id`?
I would remove all but the later.

Other than that, I think this sounds good. I will have to think about it again later when I do a proper code review & test extensively with the current engine.

> Another TBD I forgot about was whether to require the job_ticket_id URL param at all during /engine/upload, since I can find the relevant ticket just from the job_id

I would not do that; it is not valid to assume that there is a 1:1 ratio of jobs to job-completion-tickets.

kofalt:
Also, do you have any plans for disabling the old upload behavior in a later release? Probably just flipping a switch that rejects `/engine/upload` if there is no `job_ticket_id`, and removing any related code.

ambrussimon:
@nagem @kofalt Fixed/added everything discussed + rebased. PR now ready for final review.

api/jobs/handlers.py
kofalt: Do we want to address @nagem's comment before merge?

api/jobs/handlers.py
kofalt: @ambrussimon and I discussed this earlier, I proposed 300 seconds.

api/jobs/handlers.py
kofalt: Retcon: I used github incorrectly and reviewed the wrong diff :+1:

kofalt:
Code looks pretty good, I made some comments above that ended up in collapsed sections, please check them. Some code moved around so it's harder to tell which comments still apply.

I'm going to run at test job load on this API branch and report back on my results.

api/jobs/handlers.py
```diff
@@ -479,6 +480,46 @@ def retry(self, _id):
...
+
+        # Permission check
+        if not self.superuser_request:
+            j.destination.check_access(self.uid, 'rw')
+
+        if j.state != 'failed':
+            self.abort(400, 'Can only accept failed output of a job that failed')
+
+        # Remove flag from files
+        container = j.destination.get()
```
nagem: Unfortunately it looks like `ContainerReference.get` 500s when the object doesn't exist. We can either catch that generic exception here and 404 or modify ContainerReference to properly throw an APINotFound or something.

nagem:
One requirement that I realized was missed in the writeup on #968 is that jobs should not spawn until file results are accepted. The function signature for calling jobs shouldn't be too difficult to reproduce, since we are also applying the metadata here:

`rules.create_jobs(db, container_before, container_after, container_type)`

We would also need to modify rule evaluation to skip files that have the failed marking.

Sorry I didn't notice this until now, let me know if anything is unclear in my description above and we can sort it out. After that change and the one comment I added about ContainerReferencing 500ing, this is good to go in my opinion. @kofalt looks like he had a check he'd like to run before his approval, though.

kofalt:
Correct. I've been out for a few days and been unable to run my tests, but it sounds like one more modification needs to be made. Once the final changes are ready, I'll hit this branch with the engine properly to make sure it's good to go.

ambrussimon:
@kofalt @nagem Fixed. Please go ahead and hit it with the engine.

nagem:
Final changes look good from my end 👍. Nice work on this ticket!

api/jobs/handlers.py
```diff
@@ -479,6 +481,51 @@ def retry(self, _id):
...
+        container_before = copy.deepcopy(container)
+        for f in container.get('files'):
+            if f['origin'] == {'type': 'job', 'id': _id}:
+                del f['from_failed_job']
```
kofalt: I notice there's no access logging for this change. @nagem, do you think we should record who decided to go ahead & mark failed job data as accepted? Seems like an important decision that should be recorded.

api/jobs/jobs.py
```diff
@@ -198,6 +204,8 @@ def map(self):
             d.pop('request')
         if d['now'] is False:
             d.pop('now')
+        if d['failed_output_accepted'] is False:
+            d.pop('failed_output_accepted')
```
kofalt: Not sure I agree with this, but sure.

kofalt:
Did a quick rebase to make sure branch is still good, I hope you don't mind.
Feature checks out on current engine from my side of things.

Feature looks good to go pending a question above.

nagem:
Access log is added, feel free to merge when CI is green @ambrussimon.

ambrussimon:
Verified locally that no rebase is needed. Merging.

PR #1017, opened 2017-11-29
get_all analyses

### LINKED ISSUE #1024: Get All Analyses
- Parameter `children` on whether to return analyses of child containers as well
- Move tree logic to containerutil.py
- unsupport group and then we won't need to have `permissions.id` in the query for the children

hkethi002:
Fixes #1024 
- Returns all analyses given a group/project/session/acquisition
- 3 types of calls
  - `/{cont_name}/{cid}/analyses`
    - It will return all analyses of the container
    - example, `/Project/{ProjectId}/analyses` will return any analyses that have that project as a parent 
  - `/{cont_name}/{cid}/{sub_cont_name}/analyses`
    - This will return any analyses that have any of the sub_containers under the container as a parent
    -  example, `projects/{ProjectId}/acquisitions/analyses` will return any analyses that have an acquisition that is under that project as a parent
  - `/{cont_name}/{cid}/all/analyses`
    - This will return any analyses that have a parent which is a container under the container in the url
    - example, `projects/{ProjectId}/all/analyses` will return any analyses that have any session or acquisition or the project itself as a parent
- Must have permission to the top level container and (only applies when making calls at the group level) will only return analyses for which the user has permission to its parent
### Review Checklist

### DIFF
--- a/api/handlers/refererhandler.py
+++ b/api/handlers/refererhandler.py
@@ -139,20 +139,40 @@ class AnalysesHandler(RefererHandler):
         self.log_user_access(AccessType.view_container, cont_name=analysis['parent']['type'], cont_id=analysis['parent']['id'])
         return analysis
 
-    def get_all(self, cont_name, cid):
-        cid = bson.ObjectId(cid)
+    def get_all(self, cont_name, cid, **kwargs):
+        """
+        The possible endpoints for this method are:
+        /{cont_name}/{id}/analyses
+        /{cont_name}/{id}/{sub_cont_name}/analyses
+        /{cont_name}/{id}/all/analyses
+        """
+        if not cont_name == 'groups':
+            cid = bson.ObjectId(cid)
         # Check that user has permission to container
         container = ContainerStorage.factory(cont_name).get_container(cid)
         if not container:
             self.abort(404, 'Element not found in container {} {}'.format(cont_name, cid))
         permchecker = self.get_permchecker(container)
         permchecker(noop)('GET')
+        # cont_level is the sub_container name for which the analysis.parent.type should be
+        cont_level = kwargs.get('sub_cont_name')
+        cont_names = {'groups':0, 'projects':1, 'sessions':2, 'acquisitions':3}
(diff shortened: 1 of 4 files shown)


api/handlers/refererhandler.py
```diff
@@ -136,6 +136,54 @@ def get(self, **kwargs):
...
+        storages = {
+            'groups':       containerstorage.GroupStorage(),
+            'projects':     containerstorage.ProjectStorage(),
+            'sessions':     containerstorage.SessionStorage(),
+            'acquisitions': containerstorage.AcquisitionStorage()
+        }
+        parent_names = ['groups','projects','sessions', 'acquisitions']
+
+        # Only support groups/projects/sessions/acquisitions
+        if cont_name not in parent_names:
```
nagem: This can be achieved with filters on the routes table as well.

api/handlers/refererhandler.py
```diff
@@ -136,6 +136,54 @@ def get(self, **kwargs):
...
+            # For each parent id, find all of its children and add them to the list of child ids in the parent tree
+            for parent_id in parent_tree[parent_name]:
+                parent_tree[child_name] = parent_tree[child_name] + [cont["_id"] for cont in storage.get_children(parent_id, projection={'_id':1}, uid=self.uid)]
+
+            parent_name = child_name
+        # We only need a list of all the ids, no need for the tree anymore
+        parents = [pid for parent in parent_tree.keys() for pid in parent_tree[parent]]
+
+        # We set User to None because we check for permission when finding the parents
+        analyses = containerstorage.AnalysisStorage().get_all_el({'parent.id':{'$in':parents}},None,{'info': 0, 'files.info': 0})
```
nagem: Because I think about this every time we rely on uniqueness of ObjectIds across collections (even though they are not guaranteed), I thought I would post it for reference: https://stackoverflow.com/questions/4677237/possibility-of-duplicate-mongo-objectids-being-generated-in-two-different-colle

ryansanford:
Awaiting final review by @nagem

nagem:
LGTM, this came together really nicely - thanks @hkethi002 !

PR #1023, opened 2017-12-04
Simplified download path logic

hkethi002:
Ref #1005 
### Changes
Not a huge improvement and may not be worth the extra risk, but simplified _path_from_container by using the prefix and ids
### Review Checklist

### DIFF
--- a/api/download.py
+++ b/api/download.py
@@ -32,7 +32,7 @@ def _filter_check(property_filter, property_values):
 
 class Download(base.RequestHandler):
 
-    def _append_targets(self, targets, cont_name, container, prefix, total_size, total_cnt, optional, data_path, filters):
+    def _append_targets(self, targets, cont_name, container, prefix, total_size, total_cnt, data_path, filters):
         for f in container.get('files', []):
             if filters:
                 filtered = True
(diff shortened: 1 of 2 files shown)


nagem:
Another thing we could add to this is removing the `optional` flag from the request. It seems as if there was a time where files would be marked "optional" and the download endpoint would not download them until the flag `optional` was true: https://github.com/scitran/core/pull/1023/files#diff-5f6b428f8ab6dc5134ff7b9cbcad2062R49

It is safe to remove this option, as there are no "optional" files.

nagem:
One more thing - we skip missing files if we cannot resolve their filepath. We should add a warning log message to that skip to make it a little less silent. @ryansanford

hkethi002:
For removing the optional, should I remove it from the schemas, making this a breaking change?

nagem:
We can try coordinating with the clients (frontend, I don't think the SDK and CLI use that option), or remove it's reference in the download logic, wait for clients to make the change, and then remove it from the schemas. I think the second option will allow us to more forward more quickly so that sounds alright with me.

api/download.py
```diff
@@ -46,15 +46,14 @@ def _append_targets(self, targets, cont_name, container, prefix, total_size, tot
...
-                filepath = os.path.join(data_path, util.path_from_hash(f['hash']))
-                if os.path.exists(filepath): # silently skip missing files
-                    if cont_name == 'analyses':
-                        targets.append((filepath, prefix + '/' + ('input' if f.get('input') else 'output') + '/' + f['name'], cont_name, str(container.get('_id')),f['size']))
-                    else:
-                        targets.append((filepath, prefix + '/' + f['name'], cont_name, str(container.get('_id')),f['size']))
-                    total_size += f['size']
-                    total_cnt += 1
+            filepath = os.path.join(data_path, util.path_from_hash(f['hash']))
+            if os.path.exists(filepath): # silently skip missing files
```
nagem: This is the location of the skipped missing filepaths I mentioned. I would write something similar to 
> Expected filepath <filepath> to exist but it is missing. File will be skipped in download.

nagem:
This fixes an issue that is being experienced on master by a few users. I'm going to write one more test for the specific error that is being experienced and then I'd say it's ready to merge.

nagem:
I loaded a mock dataset similar to the one the customer is seeing problems with and tested its download against this branch - everything looks great. Feel free to merge this as soon as CI is green.

PR #1026, opened 2017-12-05
Fix regex project match for reaper upload

nagem:
A regex is required to do a case insensitive lookup of Mongo string fields. The existing regex did not use proper start-of-string and end-of-string tokens so it matched more generously than was expected. This change requires a case-insensitive exact match rather than the existence of the search string. 

FYI @gsfr 
### Review Checklist


ryansanford:
LGTM

Tested end-to-end and confirmed match is now case-insensitive, and anchored at both ends. Nice work!

PR #1030, opened 2017-12-06
Delete functionality - phase 1 (tagging)

### LINKED ISSUE #1012: Delete functionality
* Allow deletion of individual files, as well as of Projects, Sessions, and Acquisitions, and Analyses
* Project deletion requires Admin privileges, all other deletes require RW privileges
* Project and Session deletes include deleting of all contained Analyses
* Delete confirmation dialog offers option to "purge immediately" (meaning within 24 hours); Otherwise, deleted data will be purged after 90 days.
* Until purged, data can be restored by Flywheel support only
* Files and acquisitions, which are referenced by Analyses, may not be deleted by the user; appropriate UI/CLI/SDK warning dialogs needed
* Remove deleted items from Elastic (search)
* Tag all currently hidden files and containers as "hidden" and do not mark them as deleted.  (We will inform customers that hidden files are being restored, but are tagged as "hidden" so they can search on them and delete if wanted.)

Remaining questions:
* Do we remove jobs that only reference deleted containers?
* Do we allow deleting utility gear input files? Breaking provenance, but freeing space.

ambrussimon:
Resolves #1012 

Minor TODOs left, but bulk is ready for 1st pass/quick review for discussion.

TODO/TBD:
- [x] TODO container delete permissions still need to be adjusted
- [x] TBD containers getting the delete tag don't apply it on their files - no explicit need though. Opinions?
- [x] TBD effort could go into unifying querying so that `{'deleted': {'$exists': False}}` can be used less
- [x] TBD in hindsight, would we rather want `{'deleted': {'$ne': True}}`?

### Review Checklist

### DIFF
--- a/api/auth/containerauth.py
+++ b/api/auth/containerauth.py
@@ -26,10 +26,8 @@ def default_container(handler, container=None, target_parent_container=None):
                     required_perm = 'admin'
                 has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
             elif method == 'DELETE':
-                required_perm = 'rw'
-                if target_parent_container.get('cont_name') == 'group':
-                    # Project deletion requires admin
-                    required_perm = 'admin'
+                # Container deletion always requires admin
+                required_perm = 'admin'
                 has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
             elif method == 'PUT' and target_parent_container is not None:
                 has_access = (
--- a/api/auth/containerauth.py
+++ b/api/auth/containerauth.py
@@ -27,14 +27,10 @@ def default_container(handler, container=None, target_parent_container=None):
                 has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
             elif method == 'DELETE':
                 required_perm = 'rw'
-                if container.get('has_children'):
-                    # If the container has children or files, admin is required to delete
+                if target_parent_container.get('cont_name') == 'group':
+                    # Project deletion requires admin
                     required_perm = 'admin'
-                    additional_error_msg = 'Container is not empty.'
-                if target_parent_container:
-                    has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
-                else:
-                    has_access = _get_access(handler.uid, container) >= INTEGER_PERMISSIONS[required_perm]
+                has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
             elif method == 'PUT' and target_parent_container is not None:
                 has_access = (
                     _get_access(handler.uid, container) >= INTEGER_PERMISSIONS['admin'] and
--- a/api/dao/containerutil.py
+++ b/api/dao/containerutil.py
@@ -20,47 +20,33 @@ SINGULAR_TO_PLURAL = {
 }
 PLURAL_TO_SINGULAR = {p: s for s, p in SINGULAR_TO_PLURAL.iteritems()}
 
-def propagate_changes(cont_name, _id, query, update):
+
+def propagate_changes(cont_name, cont_ids, query, update, include_refs=False):
     """
-    Propagates changes down the heirarchy tree.
+    Propagates changes down the heirarchy tree recursively.
 
-    cont_name and _id refer to top level container (which will not be modified here)
+    cont_name and cont_ids refer to top level containers (which will not be modified here)
     """
 
+    containers = ['groups', 'projects', 'sessions', 'acquisitions']
+    if not isinstance(cont_ids, list):
+        cont_ids = [cont_ids]
+    if query is None:
+        query = {}
 
-    if cont_name == 'groups':
-        project_ids = [p['_id'] for p in config.db.projects.find({'group': _id}, [])]
-        session_ids = [s['_id'] for s in config.db.sessions.find({'project': {'$in': project_ids}}, [])]
+    if include_refs:
+        analysis_query = {'parent.type': singularize(cont_name), 'parent.id': {'$in': cont_ids}}
+        config.db.analyses.update_many(analysis_query, update)
 
-        project_q = copy.deepcopy(query)
-        project_q['_id'] = {'$in': project_ids}
-        session_q = copy.deepcopy(query)
-        session_q['_id'] = {'$in': session_ids}
-        acquisition_q = copy.deepcopy(query)
-        acquisition_q['session'] = {'$in': session_ids}
+    if cont_name in ('groups', 'projects', 'sessions'):
+        child_cont = containers[containers.index(cont_name) + 1]
+        child_ids = [c['_id'] for c in config.db[child_cont].find({singularize(cont_name): {'$in': cont_ids}}, [])]
+        child_query = copy.deepcopy(query)
+        child_query['_id'] = {'$in': child_ids}
+        config.db[child_cont].update_many(child_query, update)
 
-        config.db.projects.update_many(project_q, update)
-        config.db.sessions.update_many(session_q, update)
-        config.db.acquisitions.update_many(acquisition_q, update)
-
-
-    # Apply change to projects
-    elif cont_name == 'projects':
-        session_ids = [s['_id'] for s in config.db.sessions.find({'project': _id}, [])]
-
-        session_q = copy.deepcopy(query)
-        session_q['project'] = _id
-        acquisition_q = copy.deepcopy(query)
-        acquisition_q['session'] = {'$in': session_ids}
-
-        config.db.sessions.update_many(session_q, update)
-        config.db.acquisitions.update_many(acquisition_q, update)
-
-    elif cont_name == 'sessions':
-        query['session'] = _id
-        config.db.acquisitions.update_many(query, update)
-    else:
-        raise ValueError('changes can only be propagated from group, project or session level')
+        # Recurse to the next hierarchy level
+        propagate_changes(child_cont, child_ids, query, update, include_refs=include_refs)
 
 
 def add_id_to_subject(subject, pid):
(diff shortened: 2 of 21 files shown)


nagem:
>TBD containers getting the delete tag don't apply it on their files - no explicit need though. 
I don't see a problem until we move them to their own collection. Search will properly remove them.

nagem:
>in hindsight, would we rather want {'deleted': {'$ne': True}}

I think we would remove the key if we "un-delete" something, but curious if anyone else has an opinion on the matter.

ambrussimon:
@nagem Thanks for all the help and the input. Implemented and tested all features requested for the tagging phase (including permission handling and analysis ref checks). Sticking with `{'exists': ...}` because it's not just binary, I'm actually setting it to a `datetime`.

Also freshly rebased.

api/dao/containerutil.py
```diff
@@ -133,6 +133,16 @@ def get_stats(cont, cont_type):
     return cont
 
 
+def get_analyses(acquisition_id, filename=None):
```
nagem: Mind giving this a bit more of a descriptive name or description? And allow it to take container_name and container_id? I can see it being useful for more than just this example.

api/handlers/containerhandler.py
```diff
@@ -534,6 +534,12 @@ def delete(self, cont_name, **kwargs):
             container['has_children'] = False
         if container.get('files') or container.get('analyses'):
             container['has_children'] = True
+
+        if cont_name == 'acquisitions':
+            analysis_ids = containerutil.get_analyses(_id)
+            if analysis_ids:
+                self.abort(400, 'Cannot delete acquisition {} referenced by analyses {}'.format(_id, analysis_ids))
```
nagem: 👍

api/handlers/listhandler.py
```diff
@@ -547,6 +547,12 @@ def delete(self, cont_name, list_name, **kwargs):
         permchecker, storage, _, _, keycheck = self._initialize_request(cont_name, list_name, _id, query_params=kwargs)
 
         permchecker(noop)('DELETE', _id=_id, query_params=kwargs)
+
+        if cont_name == 'acquisitions':
+            analysis_ids = containerutil.get_analyses(_id, kwargs['name'])
```
nagem: This wouldn't allow a user to delete a file in an acquisition if a different file in that acquisition was used as an input to an analysis. Could we modify `containerutil.get_analyses()` to take an optional filename and match on that if provided?

api/auth/containerauth.py
```diff
@@ -27,14 +27,10 @@ def f(method, _id=None, payload=None, unset_payload=None, recursive=False, r_pay
                 has_access = _get_access(handler.uid, target_parent_container) >= INTEGER_PERMISSIONS[required_perm]
             elif method == 'DELETE':
                 required_perm = 'rw'
```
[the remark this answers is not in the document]
ambrussimon: Fixed for containers, but not for individual files. Is that desired?

api/dao/containerutil.py
ambrussimon: Done.

api/handlers/listhandler.py
ambrussimon: `get_analyses()` (now `get_referring_analyses()`) worked exactly like that. I now added a test for it.

ambrussimon:
@nagem @ryansanford Rebased, patched, ready for review.

api/handlers/listhandler.py
nagem: Ah my apologies I misread the function. Thanks for adding a test.

nagem:
Changes look good, holding on merge for Ryan's review of "restore" script.

ambrussimon:
@ryansanford added `undelete.py` and manually (and very limitedly) tested that container undel/recursion/file undel work.

bin/undelete.py
```diff
@@ -0,0 +1,62 @@
...
+        log.info('Removing "deleted" tag from child containers recursively...')
+        propagate_changes(args.cont_name, args.cont_id, None, update, include_refs=True)
+    else:
+        log.info('Removing "deleted" tag from file {}/{}/{}...'.format(args.cont_name, args.cont_id, args.filename))
+        for f in container.get('files', []):
+            if f['name'] == args.filename:
+                del f['deleted']
+                break
+            else:
+                raise RuntimeError('Cannot find {}/{}/{}'.format(args.cont_name, args.cont_id, args.filename))
```
ryansanford: @ambrussimon Bug here. When container contains many files, and the first isn't a match, the script ends up bailing. I'll probably take a stab at fixing myself, since it appears trivial.

bin/undelete.py
```diff
@@ -48,8 +48,8 @@ def main():
             if f['name'] == args.filename:
                 del f['deleted']
                 break
-            else:
-                raise RuntimeError('Cannot find {}/{}/{}'.format(args.cont_name, args.cont_id, args.filename))
+        else:
```
nagem: Nice, Ryan! I've never used a for/else before, I'm used to using a "found" bool. 

http://book.pythontips.com/en/latest/for_-_else.html#else-clause

nagem: I'll also give a "nice" to Ambrus because I assume that's what he intended to do :)

nagem:
I'm going to add a new access log type (delete_container) and add it where appropriate here.

ryansanford:
I suggest undelete.py recursively undelete parent containers if they are deleted. Currently it does not, resulting in the restored container referencing parent containers that are not accessible by the API.

Scenario...
1. User deletes project, which recursively deletes all nested sessions and acquisitions.
2. User needs a couple specific sessions or acquisitions restored.

ryansanford:
Another scenario that may trump the one I just posted above, or make the approach to it more nuanced...

1. A week ago user deletes a single acquisition within a project.
2. Yesterday a user deletes that project by mistake, and wants that project restored.

I think it's reasonable for a user to expect that restoring a project is scoped to the single delete event in no.2, and would not end up restoring the acquisition deleted in no.1.

Undoubtably, it's a powerful capability to make it easy to indiscriminately undelete everything under a project. I'm questioning if that fits the administrative use case we were initially targeting.

@tcbtcb @lmperry I think your perspectives are important here.

ryansanford:
Trying to undelete a file that exists, but not currently deleted is resulting in an exception. Instead, exit code should be zero with no error message, since the state of the system matches the intended state of running undelete.py.

```
2017-12-21 15:42:45 scitran.undelete              undelete.py    46:INFO Removing "deleted" tag from file acquisitions/5a39375e7053b30020c9e2f7/4374_1_1b.nii.gz...
  File "bin/undelete.py", line 59, in <module>
    main()
  File "bin/undelete.py", line 49, in main
    del f['deleted']
KeyError: 'deleted'
```

ryansanford:
Scitran Collections, nor their attached files can be undeleted with the current implementation.
- Delete of the Scitran Collection results in the removal of the corresponding mongo doc. Unclear if this was intended from description in this PR or referenced issue.
- Deletion of Scitran Collection file attachments marks those files as deleted, but undelete.py does not accept `collections` as a valid container type for undeleting of its attachments. This seems like a bug.

ryansanford:
I'm done with my review.

PSA: Search indexing implementations should be updated as appropriate before this change sees production.

nagem:
@ryansanford if we end up needing to support that scenario (files deleted in the past don't get restored), I have a few ideas we could use (timestamp set once for all containers, delete timestamp not overwritten).

ryansanford:
@ambrussimon 
I spoke with Thad, and he does prefer the option to make the undelete act on only the containers that were part of the delete in question.  A blindly recursive undelete that I initially suggested is not desired.

ambrussimon:
Phew, I addressed everything discussed, @ryansanford.

* Fixed rebase conflicts.
* Now only propagating undelete to containers deleted in the "same scope/request" (ie. deleted timestamp is the same)
* Now handling (skipping) non-deleted files and containers without error.
* If any parent container(s) are deleted, raising descriptive error.
   * `--include-parents` suppresses the error, and creates any deleted parents.
   * Undeletion is *NOT* propagated for these parents!
   * Added `--always-propagate` for handling parents that got restored via `--include-parents`
* Now handling collections. TBD: collection refs from before deletion. See [diff1](https://github.com/scitran/core/pull/1030/files#diff-83d2b08159658cfe13794fe5e54eb197R108) [diff2](https://github.com/scitran/core/pull/1030/files#diff-c4f6cb53a1bdcc2f1afd777499535cb1R105)
* Added tests...

nagem:
The additional logging changes/tests are added.

nagem:
I wanted to mention something Ambrus and I discussed offline:

Although collections can be "recovered" by the script, only the collection and it's attachments (files, notes, etc) will be recovered. Sessions and acquisitions will need to be re-added manually. 
1) Files, info and notes are the only unique data in a collection, so recovering these items via the collection would be the only way to recover them in case of an unintentional deletion. 
2) There are technically restraints in how we reference which acquisitions/sessions belong to a collection that would make it complicated to preserve this in a deleted state. Since no unique data is lost in this kind of deletion, we deemed it unnecessary to support this kind of recovery action.

nagem:
>2.) When deleting files within any container, the access log should not have the file name url encoded so it can be easily used to source the undelete command.

I'll tackle this now by adding the (unencoded) filename to the access log.

ambrussimon:
@ryansanford Good catch. `cont_name` pluralization at fault. Added test revealed that analysis deletion (propagation) and undeletion scope was off - fixed. (+as always, fresh rebase)

bin/undelete.py
```diff
@@ -60,6 +60,11 @@ def undelete(cont_name, cont_id, filename=None, include_parents=False, always_pr
     if container is None:
         raise RuntimeError('Cannot find {}'.format(cont_str))
 
+    if cont_name == 'collections':
+        log.warning('Undeleting collections is limited such that any acquisitions or sessions'
+                    'will have to be re-added to the collection again. The files and notes of'
+                    'the collection are fully restored.')
+
```
nagem: Thanks for adding this!

nagem:
Additional file context on access log added.

PR #1038, opened 2017-12-18
Create list of formalized top-level info keys that will always be returned

nagem:
Some top-level keys have special meaning in the info dictionary on objects and files. This PR allows a static list of info keys through on list endpoints. Kept simple until more nuance (project/site level options) is necessary. 

### Review Checklist

### DIFF
--- a/api/dao/basecontainerstorage.py
+++ b/api/dao/basecontainerstorage.py
@@ -1,3 +1,4 @@
+import copy
 import bson
 import datetime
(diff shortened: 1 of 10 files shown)


nagem:
Speaking offline with creator of content in BIDS keys, they will never contain PHI and there are too many for it to be reasonable to filter. I would prefer if new keys would be an "all or nothing" top level key as well.

PR #1051, opened 2018-01-12
Fix upgrade_to_41() typo

ambrussimon:
Stumbled into while adding archived->hidden upgrader. PRing separately.

### Review Checklist


nagem:
Thanks for fixing those changes... I unfortunately tested the query in a local mongo instance rather than through the update db logic. 

@ryansanford If any site has taken an update with this update in the release, we should rerun the 41st upgrade on those systems.

nagem:
FYI: @ambrussimon has a added a new integration test file for testing db upgrades that should be used in the future. It can be found here:

https://github.com/scitran/core/pull/1052/files#diff-bda0ecad35db3f24f177a0603d1adae7

Issue #969, opened 2017-10-27
Add Object Storage


nagem:
A central ticket to discuss all changes required to support object storage other than local file systems. Feel free to edit requirements and add tickets to represent requirements listed below. 

## First pass
  -  file input will still stream through API via file forms
  - [x] remove CAS from existing file storage
    - #970
    - [x] write migration script to create unique file for each file in mongo
  - [x] abstract internal file I/O for local and S3 storage (use python library: <>)
  - [x] write migration script to move file objects from local storage to object storage
    - support accessing files in either location during transition
  - [x] add ability to view/access zip members
  - [x] serve files stored in Object Storage directly through OS provider rather than stream through API using redirection (local storage will stream through API as expected)
  - [x] support range reads on individual files (no need to support individual zip member range reads)

## Second pass
  - [ ] Add service that mimics S3 signed URLs for upload, serves files for download
  - [ ] Update API endpoints to use signed URLs for upload, can be done strategy by strategy
    - Format will change into 3 steps:
      - Request permission to upload from API, specify file information for intended upload, saved in mongo collection
      - Use signed URL to upload directly to OS service
      - Complete upload process, adding file objects and updating containers as necessary
    - [x] reaper
    - [ ] packfile
    - [x] label
    - [x] uid
    - [x] uid-match
    - [x] engine
    - [x] targetted
    - More (?)

## Future work
  - [ ] Break files into separate Mongo container
    - Consider constraining file name uniqueness per container to support download
    - Will get unique Mongo (or other) id 
    - Many places in db use a “FileReference” that uses a filename and container id/type. Mostly job inputs. Consider how breaking out files into their own collection will affect existing jobs.
  - [ ] Redesign `/download` to stream from OS service (unknowns about implementation)
  - [ ] Cache zip member list
    - Cache after access rather than async add member list as zip files are added
    - 2 options:
		1. Store .meta file for each zip document uploaded to S3
		2. Functool/other memory caching

Issue #878, opened 2017-07-28
Possibly update created date, replace all file fields on file replace


nagem:
When a file with the same name as an existing file in a container is uploaded, it replaces the file itself and updates the file attributes stored in Mongo, preserving some of the old values like `created`.  It is possible we'd want to replace all fields and treat it as a new file. Further thought/discussion required.

kofalt:
:+1: to this. Replacing a file blob should definitely not keep old attributes around.

We should strive to be immutable wherever possible; I treat it as an unfortunate concession that we allow for file replacement at all. Not to mention it's a serious race condition for gears :/

Pinging @gsfr for his thoughts.

gsfr:
Agreed. If the new file is different, it should be treated like a new file, whereas "replacement" with an identical file should be treated as a no-op.

kofalt:
Ah, so this?

```python
def replace(new, old):
  if new.hash == old.hash:
    # do literally nothing, including no modified update; return 304 Not Modified
  else:
    # delete all classification / info / whatever fields; store new; return 200
```

gsfr:
@kofalt Yup.

@nagem That reminds me, do we still have that "identical zip file" check in our code? I recently verified that the reaper produces bit-identical zip files on repeated reaps.

nagem:
@gsfr I haven't seen anything like that in our code before, and doing a quick check in the likely places didn't turn up anything.

kofalt:
Yeah, I think that bit of code died a glorious death awhile back :wink:

gsfr:
Glad to hear it

PR #1059, opened 2018-01-25
Generic storage

davidfarkas:
#969 Switch to PyFilesystem

### Review Checklist


api/resolver.py
```diff
@@ -1,4 +1,4 @@
-"""
+"""1
```
nagem: Looks like an accidental addition.

bin/oneoffs/migrate_storage.py
```diff
@@ -0,0 +1,216 @@
...
+from api import config, util
+
+log = logging.getLogger('migrate_storage')
+log.setLevel(logging.INFO)
+
+COLLECTIONS_PREFIXES = [('projects', 'files'),
+                            ('acquisitions', 'files'),
+                            ('analyses', 'files'),
+                            ('sessions', 'files'),
+                            ('sessions', 'subject.files'),
```
nagem: I don't think we have any actual users using subject files to be clear, but it is good to include them just in case. It is an undocumented feature of project upload.

bin/oneoffs/migrate_storage.py
```diff
@@ -0,0 +1,216 @@
...
+    log.info('Migrate collection files...')
+
+    _files = get_collections_files()
+
+    for i, f in enumerate(_files):
+        try:
+            file_id = f['fileinfo'].get('_id', '')
+            if file_id:
+                file_path = util.path_from_uuid(file_id)
+                if not config.fs.isfile(file_path):
```
nagem: I assume here is where files that have been created since starting the conversion would be ignored? Also does this mean the conversion script can run many times and will just be a noop when no files need to be migrated?

api/util.py
```diff
@@ -262,3 +245,33 @@ def create_nonce():
     randrange = random.SystemRandom().randrange
 
     return ''.join([NONCE_CHARS[randrange(x)] for _ in range(NONCE_LENGTH)])
+
+
+def path_from_uuid(uuid_):
+    """
```
nagem: Comment below should be updated.

nagem:
I started adding comments to the PR as I went through it even though work is still being done on the branch. Address anything I mentioned now or wait until later when we move into a more "official" review state.

api/config.py
```diff
@@ -323,3 +326,11 @@ def mongo_pipeline(table, pipeline):
 
 def get_auth(auth_type):
     return get_config()['auth'][auth_type]
+
+
+# Storage configuration
+fs = open_fs(__config['persistent']['fs_url'])
+signed_url_available = ['GCSFS']
```
ambrussimon: During the design phase we discussed that storages should be able to tell about themselves, whether they support signed URLs or not. It's not a hard requirement, especially if it's not easy to solve, but I would prefer that to having this config variable.

api/files.py
```diff
@@ -165,3 +158,48 @@ def guess_type_from_filename(filename):
...
+    try:
+        if type(file_system).__name__ in config.signed_url_available:
+            return file_system.geturl(file_path, filename=filename)
+    except fs.errors.NoURL:
+        return None
+
+
+def get_fs_by_file_path(file_path):
+    if config.support_legacy_fs:
+        if config.legacy_fs.isfile(file_path):
```
ambrussimon: Both if's test `config.legacy_fs.isfile(file_path)`...?

api/handlers/listhandler.py
```diff
@@ -480,13 +480,69 @@ def get(self, cont_name, list_name, **kwargs):
...
-            self.response.headers['Content-Length'] = str(fileinfo['size']) # must be set after setting app_iter
-            if self.is_true('view'):
-                self.response.headers['Content-Type'] = str(fileinfo.get('mimetype', 'application/octet-stream'))
+            signed_url = files.get_signed_url(file_path, file_system, filename=filename)
+            if signed_url:
+                self.redirect(signed_url)
             else:
-                self.response.headers['Content-Type'] = 'application/octet-stream'
-                self.response.headers['Content-Disposition'] = 'attachment; filename="' + filename + '"'
+                range_header = self.request.headers.get('Range', '')
```
ambrussimon: If it's not too much effort, could you please hide these changes from the PR by tracking range-reads as it's target branch (at least until that's merged).

api/files.py
davidfarkas: thanks, I forgot to remove that

bin/oneoffs/migrate_storage.py
davidfarkas: Yes, exactly

bin/oneoffs/migrate_storage.py
```diff
@@ -0,0 +1,216 @@
...
+
+                update_set = {
+                    f['prefix'] + '.$.modified': datetime.datetime.utcnow(),
+                    f['prefix'] + '.$._id': file_id
+                }
+                log.debug('update file in mongo: %s' % update_set)
+                # Update the file with the newly generated UUID
+                config.db[f['collection']].find_one_and_update(
+                    {'_id': f['collection_id'], f['prefix'] + '.name': f['fileinfo']['name']},
+                    {'$set': update_set}
```
nagem: We'll need to adjust this a bit to handle when a file is replaced while the migration is underway. An example:

While iterating over all acquisition files, any outputs from gears added to new or existing acquisitions will be stored in the new storage format and location. The migration script in it's current state handles that without issue. A "feature" of our current file placers is that they _replace_ existing files in a container if the uploaded file has the same name. If that process happens after the load of all files from the acquisition collection, the migration script will move the old file from the old path to a new (and different) path and overwrite the information in the file object. This will leave the DB in a state where it looks like a dicom -> nifti gear had been rerun, generating a new file, but the nifti referenced in the DB will be the old nifti. The new nifti will still exist, but will be an unreferenced object. 

To solve this, I think we'll need to update the `query` portion of the `find_one_and_update` to not match if the object the file document points to has changed since the initial load. Maybe matching on `hash`? We wouldn't want to use `modified` because we don't care if someone updated `info` in the meantime. We could also make adjustments to updated `created` when files are replaced. 

Note: if the `modified_count` is 0, the file was replaced or removed. We'll want to clean up the file object in the new path that will go unreferenced.

nagem: https://github.com/scitran/core/issues/878 is relevant when discussing a reset of `created` on replace.

api/files.py
```diff
@@ -165,3 +158,45 @@ def guess_type_from_filename(filename):
...
+
+    if file_uuid_path and config.fs.isfile(file_uuid_path):
+        return file_uuid_path, config.fs
+    else:
+        raise fs.errors.ResourceNotFound('File not found: %s', file_info['name'])
+
+
+def get_signed_url(file_path, file_system, filename=None):
+    try:
+        if hasattr(file_system, 'get_signed_url'):
```
nagem: I made a change here from checking if `config.fs` has the method `get_signed_url` to checking if `file_system` does. I believe that's what we'll want to do while supporting legacy file system in the transition state, let me know if that change is incorrect. Ryan was seeing this issue before the fix:
```
(most recent call last):
  File "./api/handlers/listhandler.py", line 482, in get
    signed_url = files.get_signed_url(file_path, file_system, filename=filename)
  File "./api/files.py", line 190, in get_signed_url
    return file_system.get_signed_url(file_path, filename=filename)
AttributeError: 'OSFS' object has no attribute 'get_signed_url'
 request_id=e5c40c00-1518721379
```

nagem: That was when attempting to access an unmigrated file when config.fs was set to a gcs file system.

api/files.py
```diff
@@ -138,15 +137,18 @@ def _FieldStorage__write(self, line):
                 # Always write fields of type "file" to disk for consistent renaming behavior
                 if self.filename:
                     self.file = self.make_file('')
-
                     self.file.write(self._FieldStorage__file.getvalue())
+                    self.hasher.update(self._FieldStorage__file.getvalue())
                 self._FieldStorage__file = None
+
             self.file.write(line)
+            if hasattr(self, 'hasher'):
```
ambrussimon: Were you able to track down how this can get into a state where `self.hasher` is not set at this point?

api/config.py
```diff
@@ -321,3 +324,9 @@ def mongo_pipeline(table, pipeline):
 
 def get_auth(auth_type):
     return get_config()['auth'][auth_type]
+
+
+# Storage configuration
+fs = open_fs(__config['persistent']['fs_url'])
+legacy_fs = open_fs('osfs://' + __config['persistent']['data_path'])
+support_legacy_fs = True
```
ambrussimon: Can this be toggled without code change/redeploy?
I assumed we want that, and definitely desirable if not too hard.

api/handlers/listhandler.py
```diff
@@ -360,24 +357,25 @@ def _check_ticket(self, ticket_id, _id, filename):
...
-                info['members'].append(m)
-
-            return info
+        with file_system.open(file_path, 'rb') as f:
+            with zipfile.ZipFile(f) as zf:
+                info = {}
+                info['comment'] = zf.comment
+                info['members'] = []
+                for zi in zf.infolist():
+                    m = {}
```
ambrussimon: Why not
```
info['members'].append({
    'path': zi.filename,
    ...
})
```
instead?

api/placer.py
```diff
@@ -85,10 +85,12 @@ def save_file(self, field=None, file_attrs=None):
...
         Requires an augmented file field; see process_upload() for details.
         """
-
         # Save file
-        if field is not None:
-            files.move_form_file_field_into_cas(field)
+        if field is not None and self.file_processor is not None:
+            self.file_processor.store_temp_file(field.path, util.path_from_uuid(field.uuid))
+
+        # if field is not None:
```
ambrussimon: Code comment leftovers.

api/placer.py
```diff
@@ -461,47 +463,41 @@ def check(self):
...
     def finalize(self):
 
-        paths = os.listdir(self.folder)
+        paths = self.file_processor.persistent_fs.listdir(self.folder)
         total = len(paths)
 
         # Write all files to zip
         complete = 0
         for path in paths:
-            p = os.path.join(self.folder, path)
```
ambrussimon: I suggest using `relpath`/`abspath`, or maybe `full_path`, but definitely something different to just `p`.

ambrussimon:
Did more thorough review. I left a couple minor comments, and there's also the merge conflict, but LGTM!

api/files.py
davidfarkas: In case of the metadata we don't have a file name so we won't create the file and hasher. I'll add a comment to clarify this.

nagem:
My final testing of the migration script edge cases (Analysis inputs and files changing while the upgrade is taking place) passed as expected. Awesome work!!

After a rebase I'd consider this PR approved and ready to merge, but we should hold off on merging until `file-browser` goes in first. That will also be after the switch over to the new repo location.

PR #1062, opened 2018-01-31
Menagerie of job handler improvements

kofalt:
1. Adding logs to a job will now unconditionally heartbeat that job first. A job consumer now has half the requests over time required when a job is running.
1. Creating a job-completion ticket will prevent that job from being orphaned. A job consumer no longer has to keep heartbeating while an upload is in progress.
1. For this reason, tickets now last an hour instead of 5 minutes.
1. Job tickets are now their own class in `jobs.py`.
1. A bunch of handlers were not using `@directives` or had ineffective `return` statements. Corrected.
1. Moved some repetitive download logic to `util.set_for_download`.
1. Moved some dumb humanizing logic to `util.humanize_validation_error`.
1. Harmonized some, but not all, uses of `self.response.json` variable naming.
1. Slightly reduced the log spam when jobs-next has nothing to return.
1. Fixed a bug in the new job-ticket handler that was not marking jobs as complete / failed.
1. Removed meaningless tempdir warnings. Congrats, CI is 288 lines shorter :tada:

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -8,24 +8,24 @@ import StringIO
 from jsonschema import ValidationError
 from urlparse import urlparse
 
+from . import batch
+from .. import config
 from .. import upload
 from .. import util
-from ..auth import require_drone, require_login, has_access
+from ..auth import require_drone, require_login, require_admin, has_access
+from ..auth.apikeys import JobApiKey
 from ..dao import hierarchy
 from ..dao.containerstorage import ProjectStorage, SessionStorage, AcquisitionStorage
 from ..dao.containerutil import ContainerReference, pluralize
+from ..util import humanize_validation_error, set_for_download
+from ..validators import validate_data, verify_payload_exists
 from ..web import base
 from ..web.encoder import pseudo_consistent_json_encode
 from ..web.errors import APIPermissionException, APINotFoundException, InputValidationException
 from ..web.request import AccessType
-from .. import config
-from . import batch
-from ..validators import validate_data, verify_payload_exists
-
-from ..auth.apikeys import JobApiKey
 
 from .gears import validate_gear_config, get_gears, get_gear, get_invocation_schema, remove_gear, upsert_gear, suggest_container, get_gear_by_name, check_for_gear_insertion
-from .jobs import Job, Logs
+from .jobs import Job, JobTicket, Logs
 from .batch import check_state, update
 from .queue import Queue
 from .rules import create_jobs, validate_regexes
@@ -35,11 +35,9 @@ class GearsHandler(base.RequestHandler):
 
     """Provide /gears API routes."""
 
+    @require_login
     def get(self):
         """List all gears."""
-
-        if self.public_request:
-            self.abort(403, 'Request requires login')
 
         gears   = get_gears()
         filters = self.request.GET.getall('filter')
@@ -49,13 +47,9 @@ class GearsHandler(base.RequestHandler):
 
         return gears
 
+    @require_login
     def check(self):
-        """
-        Check if a gear upload is likely to succeed.
-        """
-
-        if self.public_request:
-            self.abort(403, 'Request requires login')
+        """Check if a gear upload is likely to succeed."""
 
         check_for_gear_insertion(self.request.json)
         return None
@@ -63,27 +57,16 @@ class GearsHandler(base.RequestHandler):
 class GearHandler(base.RequestHandler):
     """Provide /gears/x API routes."""
 
+    @require_login
     def get(self, _id):
-        """Detail a gear."""
-
-        if self.public_request:
-            self.abort(403, 'Request requires login')
-
         return get_gear(_id)
 
+    @require_login
     def get_invocation(self, _id):
+        return get_invocation_schema(get_gear(_id))
 
-        if self.public_request:
-            self.abort(403, 'Request requires login')
-
-        gear = get_gear(_id)
-        return get_invocation_schema(gear)
-
+    @require_login
     def suggest(self, _id, cont_name, cid):
-
-        if self.public_request:
-            self.abort(403, 'Request requires login')
-
         cr = ContainerReference(cont_name, cid)
         if not self.superuser_request:
             cr.check_access(self.uid, 'ro')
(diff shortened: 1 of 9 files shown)


nagem:
>Creating a job-completion ticket will prevent that job from being orphaned. A job consumer no longer has to stop heartbeating while an upload is in progress.

Did you mean a job consumer will no longer _have to keep heartbeating_?

nagem:
>A bunch of handlers were not using @directives

👍 👍 👍

api/jobs/handlers.py
```diff
@@ -328,40 +280,30 @@ def next(self):
...
         c = j.config
         if c is None:
             c = {}
 
-        # Serve config as formatted json file
-        self.response.headers['Content-Type'] = 'application/octet-stream'
-        self.response.headers['Content-Disposition'] = 'attachment; filename="config.json"'
-
         # Detect if config is old- or new-style.
         # TODO: remove this logic with a DB upgrade, ref database.py's reserved upgrade section.
```
nagem: I believe this db upgrade has been completed, can we remove the check below?

api/jobs/handlers.py
```diff
@@ -474,18 +414,12 @@ def get_logs_html(self, _id):
...
-
-
         doc = self.request.json
 
-        try:
-            Job.get(_id)
-        except Exception: # pylint: disable=broad-except
-            raise APINotFoundException('Job not found')
+        j = Job.get(_id)
+        Queue.mutate(j, {}) # Unconditionally heartbeat
```
nagem: The removal of the exception handling causes this endpoint to `500` when the job does not exist rather than `404`. IMO we should be moving towards a state where the API only `500`s on server error, not user/client error. You could switch the `Jobs.get` error to an `APINotFoundExeception` to avoid having to catch Exceptions in handler layers, like we do in container storage classes.

api/jobs/queue.py
```diff
@@ -417,13 +417,26 @@ def scan_for_orphans():
...
+        }
 
         while True:
+            orphan_candidate = config.db.jobs.find_one(query)
+            if orphan_candidate is None:
+                break
+
+            # If the job is currently attempting to complete, do not orphan.
+            ticket = JobTicket.find(orphan_candidate['_id'])
+            if ticket is not None and len(ticket) > 0:
```
nagem: You could write `if ticket:` and get the same result, but it's alright if you were going for clarity.

api/jobs/queue.py
```diff
@@ -417,13 +417,26 @@ def scan_for_orphans():
...
+
+            # CAS this job, since it does not have a ticket
+            select = { '_id': orphan_candidate['_id'] }
+
             doc = config.db.jobs.find_one_and_update(
-                {
-                    'state': 'running',
-                    'modified': {'$lt': datetime.datetime.utcnow() - datetime.timedelta(seconds=100)},
-                },
+                dict(query, **select),
```
nagem: Pretty sneaky sis.

api/util.py
```diff
@@ -174,6 +174,18 @@ def sanitize_string_to_filename(value):
...
+    """
+    Takes a jsonschema.ValidationError, returns a human-friendly string
+    """
+
+    key = 'none'
+    if len(val_err.relative_path) > 0:
+        key = val_err.relative_path[0]
+    message = val_err.message.replace("u'", "'")
+
+    return 'Gear manifest does not match schema on key ' + key + ': ' + message
```
nagem: This should be useful in more than just gear validation errors. If its going to live in general util can we adjust it a bit to not mention gear manifests or take an optional string param that can be inserted into the message response?

api/jobs/handlers.py
kofalt: I don't know. The instructions in both locations were very clear, so I'm assuming not.

api/jobs/handlers.py
kofalt: I checked down the call stack, and `APINotFoundExeception` is what is thrown in this case. Unless I misread.

api/jobs/queue.py
kofalt: Yup, I really dislike random-thruthiness checks

api/jobs/queue.py
kofalt: :roll_eyes:

api/jobs/handlers.py
nagem: You were right, I misread. I believe it used to be an Exception and I didn't look close enough.

api/util.py
kofalt: Mmm, fair, I'll drop the gear specific part

kofalt:
Comments addressed, and small log tweak added.

9) Slightly reduced the log spam when jobs-next has nothing to return.

api/jobs/handlers.py
```diff
@@ -301,67 +259,51 @@ def stats(self):
...
-
         tags = self.request.GET.getall('tags')
         if len(tags) <= 0:
             tags = None
 
         job = Queue.start_job(tags=tags)
 
         if job is None:
-            self.abort(400, 'No jobs to process')
+            raise InputValidationException('No jobs to process')
```
nagem: Since we're adjusting this anyway, could we use an `APINotFoundException`/`404`? The client didn't make an improperly formatted request.

kofalt: Uh, good question... The route exists and is valid, so a 404 does not feel appropriate here. I wanted to use something that kept the 400, hence my choice. I could make a new error class if you wanted, but it just seemed so trivial I didn't bother initially.

nagem: 404 as in a "next job" is missing. IMO an empty response would be the best way to respond but easier to write around a non-200. Not a huge deal we change it.

kofalt: Yeah, the trick here is I didn't want to change behavior for the existing clients. I could go check or change our consumers to not mind it, but it doesn't seem worth a breaking change.

kofalt:
10. Fixed a bug in the new job-ticket handler that was not marking jobs as complete / failed.
11. Removed meaningless tempdir warnings. Congrats, CI is 288 lines shorter :tada:

kofalt:
10. Fixed a bug in the new job-ticket handler that was not marking jobs as complete / failed.
11. Removed meaningless tempdir warnings. Congrats, CI is 288 lines shorter :tada:

ryansanford:
@kofalt @nagem Any final changes or review blocking the merge of this? Correct that there are still no breaking changes here?

I understand this PR to be a blocker to wrapping up the new Flywheel Engine.

nagem:
@ryansanford Looks like more changes were added again so I’ll need to review those before it can merge.

kofalt:
FWIW, this branch does not need to be a blocker; I have configured e2 to use the new job-complete ticketing that this branch modifies, but I made sure it is very easy to disable. Six character diff.

I can even make that configurable if desired... maybe I should, just in case. :thinking:

ryansanford:
I'd prefer this PR to have final review and merge. Thanks for the flexibility though @kofalt

nagem:
@ambrussimon do you mind taking a look at the last two commits if you have a chance? They reference your work with job completion tickets.

PR #1066, opened 2018-02-06
On packfile upload, use acquisition timestamp and subject code to find containers

nagem:
If acquisition timestamp and/or subject code are provided in the request payload metadata, use this information to be more specific in the query to find an existing container. 

This fixes an issue where files with the same acquisition name but different timestamps ended up in the same container when they should be separate containers.

This also fixes an issue if two sessions with the same label but different subject codes are uploaded.  

### Review Checklist


nagem:
Writing tests for situations noted above, look for those soon.

PR #1069, opened 2018-02-07
Propose batch from pre-constructed jobs

### LINKED ISSUE #940: Propose a batch via a set of already-constructed jobs
In addition to 

```go
ProposeBatch(gearId string, config map[string]interface{}, tags []string, targets []*ContainerReference)
```

We would support:

````go
ProposeBatchJobs(gearId string, config map[string]interface{}, tags []string, jobs []*Job)
````

### PR DESCRIPTION
**Pull request by hkethi002**
Fixes #940 
### Changes
- New Endpoint: `POST /batch/jobs` with a list of jobs, `jobs: [{...}, {...}, ...]`, in the payload
- The job maps from the payload are stored in `batch.proposal.preconstructed_jobs` in mongo
- Jobs are validated on proposal for gear manifest, inputs, config, and permissions
- On batch run, the jobs then follow the normal job cycle
### Review Checklist

### DIFF
--- a/api/jobs/batch.py
+++ b/api/jobs/batch.py
@@ -134,84 +134,100 @@ def run(batch_job):
     proposal = batch_job.get('proposal')
     if not proposal:
         raise APIStorageException('The batch job is not formatted correctly.')
-    proposed_inputs = proposal.get('inputs', [])
-    proposed_destinations = proposal.get('destinations', [])
+    preconstructed_jobs = proposal.get('preconstructed_jobs')
 
-    gear_id = batch_job['gear_id']
-    gear = gears.get_gear(gear_id)
-    gear_name = gear['gear']['name']
+    # If Running a batch from already-constructed jobs
+    if preconstructed_jobs:
+        origin = batch_job.get('origin')
+        jobs = []
+        job_ids = []
 
-    config_ = batch_job.get('config')
-    origin = batch_job.get('origin')
-    tags = proposal.get('tags', [])
-    tags.append('batch')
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by AshleyAncona** [comment]
@nagem and @kofalt this is ready for review

File: api/jobs/batch.py [hunk 1]
```diff
@@ -134,84 +134,100 @@ def run(batch_job):
...
-        'config':   config_,
-        'gear_id':  gear_id,
-        'tags':     tags,
-        'batch':    str(batch_job.get('_id')),
-        'inputs':   {}
-    }
+    preconstructed_jobs = proposal.get('preconstructed_jobs')
+
+    # If Running a batch from already-constructed jobs
+    if preconstructed_jobs:
```
kofalt: Break these two branches out into different functions, and call them from here.

api/jobs/queue.py
```diff
@@ -128,11 +128,10 @@ def retry(job, force=False):
 
         return new_id
 
-
     @staticmethod
-    def enqueue_job(job_map, origin, perm_check_uid=None):
+    def validate_job(job_map, origin, create_job=False, perm_check_uid=None):
```
kofalt: Please do not change this function signature. Instead, revert changes to this function, then remove `job.insert()` to make it consistent with the original documentation. Then check usage of `enqueue_job` to see who will need `job.insert()` added after calling.

kofalt (commented):
-

kofalt:
Notably, this PR is a lot easier to read when using `w=1`.

PR #1078, opened 2018-02-14
Fix handling of multi-value file fields

ehlertjd:
Some clients (e.g. Swagger), rather than specifying `file1`, `file2`, etc... will
specify multiple values for `file` when uploading multiple files.
This code change will handle lists of values for individual fields.

### Review Checklist


kofalt:
This change is confusing to me, because I know for a fact that multiple `file` fields has worked with this logic in the past. Do you have a curl that will fail on master and succeed with this PR?

ehlertjd:
@kofalt The modified integration test fails on master and succeeds on this PR.

nagem:
LGTM!

PR #1079, opened 2018-02-16
Analysis endpoints update

### LINKED ISSUE #1058: Formalize "adhoc" analyses
### Current Functionality
The API allows users create an analysis by uploading files, marking them as inputs and outputs. The `POST` body includes all information about the files and analysis the user wishes to set and the action is started and completed in one request. 

The default and expected analysis type is currently job-based analyses. The user specifies information about the analysis and the job to be created. The inputs on the job are considered inputs to the analysis and the results of the job are uploaded as outputs on the analysis. 

### Proposed Functionality
The existing "adhoc" analysis endpoint will be replaced with new endpoints: 
  - The first, where a user creates the analysis and references all *internal* inputs as ContainerReferences. External inputs uploaded by the user will not be accepted
  -  The second, where a user can upload the outputs resulting from their offline analysis work

Clients (for example SDKs and CLIs) would support a workflow allowing users to pull the inputs referenced into a local workspace, process and analyze these files, and push the outputs to core after an arbitrary amount of time. This would require a way to retrieve the input files for the analysis after the initializing endpoint has been used. 

### Things to consider
  - there are a few users depended on the existing "adhoc" analysis endpoint, this would be a breaking change that affects their workflow. It may be necessary to allow a transition period for these users.
  - an analysis in the incomplete state would still need to be returned in lists like analyses whose jobs have not completed
  - permissions around who can upload output results (Anyone with POST permissions? Only the user that created the analysis?)
  - where/if to store file-like objects involved in the analysis computation like code/scripts/etc
  - moving this object to a different name/container type if difference is notable
  - whether or not to continue the practice of copying the current state of file objects to the analysis' file list. Would become external references when files become separate container

ambrussimon:
Resolves #1058 

#### Old - _fails schema validation_
```
POST /api/<cont_name>/<cont_id>/analyses?job=true

{"analysis": {"label": "with-job"},
 "job": {"gear_id": gearid,
         "inputs": {"csv": {"type": "acquisition", "id": acqid, "name": "file.csv"}}}
```
#### New - no URL param, analysis payload contains job
```
POST /api/<cont_name>/<cont_id>/analyses

{"label": "with-job",
 "job": {"gear_id": gearid,
         "inputs": {"csv": {"type": "acquisition", "id": acqid, "name": "file.csv"}}}
```
#### Old - _accepted as legacy method_
```
POST /api/<cont_name>/<cont_id>/analyses

// File form with both input and output files, plus metadata below
{"label": "no-job",
 "inputs": [{"name": "input.txt"}],
 "outputs": [{"name": "output.txt"}]}
```
#### New - JSON payload instead of file form, inputs only, inputs are filerefs
```
POST /api/<cont_name>/<cont_id>/analyses

{"label": "no-job",
 "inputs": [{"type": "acquisition", "id": acqid, "name": "input.txt"}]}
```
### Get analysis
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>
GET /api/analyses/<analysis_id>
```
#### Old response - inputs and outputs under `files`, tagged
```
{ ...
 "files": [{"name": "input.txt", ... , "input": true},
           {"name": "output.txt", ... , "output": true}],
... }
```
#### New response - inputs under `inputs`, no `input`/`output` tag
```
{ ...
 "inputs": [{"name": "input.txt", ... }],
 "files": [{"name": "output.txt", ... }],
... }
```
### Download analysis file(s)
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>/files[/filename]
GET /api/analyses/<analysis_id>/files[/filename]
```
### Download analysis input(s) - works exactly like outputs (`/files`)
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>/inputs[/filename]
GET /api/analyses/<analysis_id>/inputs[/filename]
```
Note that creating an inputs only download ticket via `/downloads` is *NOT* supported.
### Upload analysis outputs - only allowed when it's without a job
```
POST /api/<cont_name>/<cont_id>/analyses/<analysis_id>/files
POST /api/analyses/<analysis_id>/files

// File form with output files, plus metadata below
[{"name": "output1.txt", ... },
 {"name": "output2.txt", ... }]
```

### Review Checklist


swagger/templates/analysis-inputs.yaml
```diff
@@ -0,0 +1,45 @@
...
+        in: query
+        name: ticket
+    responses:
+      '200':
+        description: ''
+        schema:
+          $ref: schemas/output/analysis-inputs-create-ticket.json
+        examples:
+          response:
+            $ref: examples/output/analysis-inputs-create-ticket.json
```
ehlertjd: Is there any significant difference between this and `analysis-files.yaml`? Could the differences be paramaterized so that we don't need separate templates for `analysis-inputs.yaml` and `analysis-files.yaml`? Similarly, I'm guessing that the missing `analysis-inputs-create-ticket-filename.yaml` is closely related to `analysis-files-create-ticket-filename.yaml`.

In either case, can we update `analysis-files.yaml` to clarify that these now only include output files?

ehlertjd:
This is a breaking change for [scitran/client](https://github.com/scitran/client/blob/2d00804b9cf3f1f517a2f3c21216c317b3ccc9b2/%40scitran/putAnalysis.m#L103) and [scitran/python-client](https://github.com/scitran/python-client/blob/35ce6409737d570d8e054a5a69a9948f4b56fa66/scitran_client/flywheel_analyzer.py#L176).

ambrussimon:
@nagem Please see summary below.

## Breaking changes
### Create analysis _with job_
#### Old - _fails schema validation_
```
POST /api/<cont_name>/<cont_id>/analyses?job=true

{"analysis": {"label": "with-job"},
 "job": {"gear_id": gearid,
         "inputs": {"csv": {"type": "acquisition", "id": acqid, "name": "file.csv"}}}
```
#### New - no URL param, analysis payload contains job
```
POST /api/<cont_name>/<cont_id>/analyses

{"label": "with-job",
 "job": {"gear_id": gearid,
         "inputs": {"csv": {"type": "acquisition", "id": acqid, "name": "file.csv"}}}
```
### Create analysis _without job_
#### Old - _accepted as legacy method_
```
POST /api/<cont_name>/<cont_id>/analyses

// File form with both input and output files, plus metadata below
{"label": "no-job",
 "inputs": [{"name": "input.txt"}],
 "outputs": [{"name": "output.txt"}]}
```
#### New - JSON payload instead of file form, inputs only, inputs are filerefs
```
POST /api/<cont_name>/<cont_id>/analyses

{"label": "no-job",
 "inputs": [{"type": "acquisition", "id": acqid, "name": "input.txt"}]}
```
### Get analysis
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>
GET /api/analyses/<analysis_id>
```
#### Old response - inputs and outputs under `files`, tagged
```
{ ...
 "files": [{"name": "input.txt", ... , "input": true},
           {"name": "output.txt", ... , "output": true}],
... }
```
#### New response - inputs under `inputs`, no `input`/`output` tag
```
{ ...
 "inputs": [{"name": "input.txt", ... }],
 "files": [{"name": "output.txt", ... }],
... }
```
### Download analysis file(s)
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>/files[/filename]
GET /api/analyses/<analysis_id>/files[/filename]
GET /api/download?ticket= {"nodes": [{"level": "analysis", "_id": <analysis_id>}]}
```
#### Old response - for inputs and outputs
#### New response - for outputs only
## Additive changes
### Download analysis input(s) - works exactly like outputs (`/files`)
```
GET /api/<cont_name>/<cont_id>/analyses/<analysis_id>/inputs[/filename]
GET /api/analyses/<analysis_id>/inputs[/filename]
```
Note creating an analysis input download ticket via `/api/download` is not supported!
### Upload analysis outputs - only allowed when it's without a job
```
POST /api/<cont_name>/<cont_id>/analyses/<analysis_id>/files
POST /api/analyses/<analysis_id>/files

// File form with output files, plus metadata below
[{"name": "output1.txt", ... },
 {"name": "output2.txt", ... }]
```

nagem:
@lmperry mind taking a look at the changes described above and let us know if anything isn't how you would expect it?

api/handlers/refererhandler.py
```diff
@@ -211,9 +189,23 @@ def delete(self, cont_name, cid, _id):
...
+    def upload(self, **kwargs):
+        """Upload ad-hoc analysis outputs generated offline."""
+        _id = kwargs.get('_id')
+        analysis = self.storage.get_container(_id)
+        parent = self.storage.get_parent(analysis['parent']['type'], analysis['parent']['id'])
+        permchecker = self.get_permchecker(parent)
+        permchecker(noop)('POST')
+
+        if analysis.get('job'):
+            raise InputValidationException('Analysis created via a job does not allow file uploads')
```
nagem: Maybe I missed it but we might want to add a check to ensure the analysis doesn't already have inputs. I believe this action should only be completed once.

nagem:
LGTM other than the comment I mentioned above. Let's hold off on a merge until @lmperry gets a chance to take a look, though.

lmperry:
@nagem, thanks for pinging me on this. LGTM.

nagem:
@ambrussimon and I talked about this offline but we'll also need a db update to bring existing analysis into the new format (separate input and output arrays).

nagem:
@ambrussimon I made a small change to be kind to requests that don't specify the correct filegroup type when trying to download an individual file. Makes it less of a breaking change for the frontend client.

kofalt:
Overall this looks like a positive change, nice work @ambrussimon. Quick questions:

1. I don't see a database migration, so the DB format is staying the same, correct? If so, does the change to `gears.py` still make sense?
1. Is it worth fixing `TargetedPlacer` to take multiple files, rather than splitting off a fixed version into `TargetedMultiPlacer`?

nagem:
>I don't see a database migration, so the DB format is staying the same, correct? If so, does the change to gears.py still make sense?

The DB migration was referenced in a comment above, it's currently under construction.

kofalt:
3. I notice from the SDK PR that we're using a `user` string on `AdhocAnalysis`. Any chance that could just be a normal origin struct?
4. Sanity check, it's now entirely up to the analysis creator if a job should be created, right? There's no field you have to set to prevent that from happening?

kofalt:
@nagem solid, let's double-check that change after the upgrade. I'd be a fan of outputs on `files` which wouldn't require any further change, but anything works :)

nagem:
>3 
Nice catch, that should be set by core based on request origin, not specified by the user. I think that was old logic from the original legacy style that ended up sticking around. 

>4
Correct

ambrussimon:
> 2

Schema-wise that's non-reconcilable (cannot make endpoints accept `{one}` OR `[{more}, {files}]`), or at least not easily/I failed at it. I also imagine it's non-trivial to handle such a schema in the SDK. I'm already sort of pushing the limits of what can/should be handled in a single endpoint with analyses.

Instead, I'd suggest a breaking change in the future, where `TargetedPlacer := TargetedMultiPlacer`, eg. we use a list of files even if `length == 1`. Unless anyone's aware of clients or something heavily relying on that not breaking.

nagem:
Upgrade looks good 👍

PR #1080, opened 2018-02-16
Support <id:xyz> syntax in resolver

kofalt:
When resolving a human-meaningful path, you can now send `<id:xyz>` syntax to unambiguously resolve a path segment. This is useful when an ambiguity exists amongst a container's children, such as two sessions with the same label.

Para exemplar:

```
POST /api/resolve HTTP/1.1

{"path":["scitran","Testdata","foo","<id:5a871c5f3871ca01205e78a2>"]}
```

### DIFF
--- a/api/resolver.py
+++ b/api/resolver.py
@@ -34,7 +34,7 @@ class Node(object):
         raise NotImplementedError() # pragma: no cover
 
     @staticmethod
-    def filter(children, criterion, id=False):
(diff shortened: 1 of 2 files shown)


nagem:
The change looks innocuous enough. "Filtering" by `id` in code shines a light on the fact that we shouldn't be doing any label filtering in code either and let Mongo handle that. That change can be addressed outside this PR.

PR #1082, opened 2018-02-21
Properly handle replacing of files, rerun jobs when file is replaced

### LINKED ISSUE #878: Possibly update created date, replace all file fields on file replace

### LINKED ISSUE #1081: Replacement of a file should trigger gear rule
A file was replaced with updated contents via the upload/reaper endpoint, resulting in the measurements and info fields being cleared. This replacement happened after the dicom-mr-classifier gear completed. The result was that the info and measurements fields were left empty, and the gear rule in place did not fire a new job to reconstitute their contents. Also, the generated nifti files are not in sync with the latest dicom.

nagem:
Closes #878 
Closes #1081 

Test fixes and additions yet to come

### DIFF
--- a/api/dao/hierarchy.py
+++ b/api/dao/hierarchy.py
@@ -229,47 +229,45 @@ def upsert_fileinfo(cont_name, _id, fileinfo):
     _id = bson.ObjectId(_id)
 
     container_before = config.db[cont_name].find_one({'_id': _id})
-    container_after, file_before = None, None
+    container_after = None
+    saved_state = 'saved'
 
+
+    # Look to see if file with the same name already exists in the container
     for f in container_before.get('files',[]):
-        # Fine file in result and set to file_after
+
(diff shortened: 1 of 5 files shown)


nagem:
@gsfr How this affects the reaper:

If a reaper (or anything) sends a file to a container with the same filename:
 - If the file object is the same (via hash), the existing file object in the file system and in mongo is unchanged. The file update is ignored.
 - If the file object has a different hash, the existing file object is removed from Mongo and a new file is added to Mongo with metadata specified on that upload. All new jobs resulting from rules are run, as well as all old jobs that had this file as input. 

Let me know if this is contrary to what you'd expect.

PR #1091, opened 2018-03-01
Add elapsed to job completion ticket

kofalt:
Also provide a more robust check for job state at the beginning of both placers, tests were failing without it and it's smart to do here.


nagem:
LGTM

PR #1092, opened 2018-03-02
Add queue peek support

kofalt:
Allows a consumer to peek from the queue.
Not guaranteed to return a job, even if there are matching candidates.

So is it peek, or maybe-peek? We report, you decide.

### Review Checklist

### DIFF
--- a/api/jobs/handlers.py
+++ b/api/jobs/handlers.py
@@ -269,11 +269,12 @@ class JobsHandler(base.RequestHandler):
 
     @require_admin
     def next(self):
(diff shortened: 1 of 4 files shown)


kofalt:
Also drops support for `job.now`, which was an unused preemptive feature.

nagem:
LGTM, I'll spend ~3 minutes seeing if there is a more obvious way to do a noop mongo update but with the comment it's fine. After a test is added to hit the `peek` branches, feel free to merge at your convenience.

PR #1093, opened 2018-03-02
Use (now available) pypi tempfile backport

ambrussimon:
Found 3rd party "official" backport of py3 tempfile.

### Review Checklist


kofalt:
This brings back the log spam :pensive: 
https://travis-ci.org/scitran/core/builds/348324143#L1304

But whatever lol.

gsfr:
Would be good to silence those warnings. @ambrussimon Worth subclassing?

ambrussimon:
@kofalt Thanks for catching that regression. Fixed by disabling warnings with that message.

Also, polling for opinions here about further decreasing test output noise;
What do you think about getting rid of the log lines reporting `Initialized request ...`, at least for tests, using something like `if os.environ.get('SCITRAN_RUNTIME_COVERAGE') == 'true'`? (Alternative suggestions welcome.)

kofalt:
I very much would love for that log line be disabled by default.

gsfr:
I'm definitely in favor of reducing test output noise.

PR #1097, opened 2018-03-08
Add file and tree browsing to gear suggest endpoint

nagem:
Tests incoming, functionality not 100% complete.

### Review Checklist

### DIFF
--- a/api/dao/basecontainerstorage.py
+++ b/api/dao/basecontainerstorage.py
@@ -63,7 +63,6 @@ class ContainerStorage(object):
         """
         cont_storage_name = containerutil.singularize(cont_name).capitalize() + 'Storage'
         for subclass in cls.__subclasses__():
-            config.log.warning('we are comparing {} and {}'.format(subclass.__name__, cont_storage_name))
             if subclass.__name__ == cont_storage_name:
                 return subclass()
         return cls(containerutil.pluralize(cont_name))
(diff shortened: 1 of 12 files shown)


ryansanford:
@nagem Looks like the last couple commits are related to tests. Does that suggest you believe the functionality to be complete?

nagem:
@ryansanford it is ready for review as well as testing with the clients. The format of the response _should_ be the same as the former mocked endpoint but there might be some discrepancies that would be very easy to fix.

api/dao/basecontainerstorage.py
```diff
@@ -124,6 +130,70 @@ def get_children(self, _id, projection=None, uid=None):
...
+            query = {containerutil.singularize(self.cont_name): bson.ObjectId(_id)}
+
+        if uid:
+            query['permissions'] = {'$elemMatch': {'_id': uid}}
+        if not projection:
+            projection = {'info': 0, 'files.info': 0, 'subject': 0, 'tags': 0}
+        return ContainerStorage.factory(child_name).get_all_el(query, None, projection)
+
+
+    def get_parent_tree(self, _id, cont=None, projection=None, add_self=False):
```
nagem: Are there much better ways I could have done this tree walking? Absolutely. 

I plan on rewriting it to be a bit more recursive when we formalize subjects, currently scheduled as the next big project.

api/jobs/gears.py
```diff
@@ -54,38 +53,24 @@ def get_gear_by_name(name):
...
-        files[:] = [x for x in files if x.get('output')]
-        for f in files:
-            f['suggested'] = {}
-            for x in schemas:
-                f['suggested'][x] = schemas[x].is_valid(f)
-        analysis['files'] = files
+    for f in files:
+        f['suggested'] = {}
+        for x in schemas:
+            f['suggested'][x] = schemas[x].is_valid(f)
```
kofalt: This is a weird change, but ok.

api/jobs/handlers.py
```diff
@@ -67,12 +67,59 @@ def get_invocation(self, _id):
...
+            children = storage.get_children(cid, projection={'files': 0})
+            response['children'][pluralize(storage.child_cont_name)] = [{'cont_name': storage.child_cont_name, '_id': c['_id'], 'label': c['label']} for c in children]
+
+        parents = storage.get_parent_tree(cid, cont=container)
+        response['parents'] = [{'cont_name': p['cont_type'], '_id': p['_id'], 'label': p.get('label', 'unknown')} for p in parents]
+
+        files = add_suggest_info_to_files(gear, container.get('files', []))
+        response['files'] = [{'name': f['name'], 'suggested': f['suggested']} for f in files]
+
+        return response
```
kofalt: I would prefer this be hidden away in more function calls than logic in the handler, but I don't see any obvious fast way to do that.

api/jobs/gears.py
nagem: We only need to provide one container's file suggestions at a time so it was modified to take an array of files rather than several named arrays of containers containing files.

api/jobs/handlers.py
nagem: Yeah it should get broken into it's own method like `build_container_tree` or something and then run `suggest` on the resulting file array. I'll make sure that gets adjusted in the next major change.

PR #1099, opened 2018-03-13
Signed url upload

davidfarkas:
With this PR the clients will be able to upload files using signed URLs. The only limitation is that only one file can be uploaded per request.
The packfile upload is not supported because it is not possible to create the zip file without the API downloads the files again.
It is not a braking change, the upload endpoints are backward compatible, so the clients can use the old way. 

### Review Checklist

### DIFF
--- a/api/upload.py
+++ b/api/upload.py
@@ -31,7 +31,7 @@ Strategy = util.Enum('Strategy', {
 
 
 def process_upload(request, strategy, container_type=None, id_=None, origin=None, context=None, response=None,
-                   metadata=None):
+                   metadata=None, file_fields=None, tempdir=None):
     """
     Universal file upload entrypoint.
 
@@ -78,14 +78,17 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
 
     # The vast majority of this function's wall-clock time is spent here.
     # Tempdir is deleted off disk once out of scope, so let's hold onto this reference.
-    file_processor = files.FileProcessor(config.get_item('persistent', 'data_path'), config.fs)
-    form = file_processor.process_form(request)
+    file_processor = files.FileProcessor(config.get_item('persistent', 'data_path'), config.fs, tempdir_name=tempdir)
+    if not file_fields:
+        form = file_processor.process_form(request)
+        # Non-file form fields may have an empty string as filename, check for 'falsy' values
+        file_fields = extract_file_fields(form)
 
-    if 'metadata' in form:
-        try:
-            metadata = json.loads(form['metadata'].value)
-        except Exception:
-            raise FileStoreException('wrong format for field "metadata"')
+        if 'metadata' in form:
+            try:
+                metadata = json.loads(form['metadata'].value)
+            except Exception:
+                raise FileStoreException('wrong format for field "metadata"')
 
     placer_class = strategy.value
     placer = placer_class(container_type, container, id_, metadata, timestamp, origin, context, file_processor)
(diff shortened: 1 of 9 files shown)


api/upload.py
```diff
@@ -204,14 +255,42 @@ def engine(self):
...
         }
-        if level == 'analysis':
-            return process_upload(self.request, Strategy.analysis_job, origin=self.origin, container_type=level, id_=cid, context=context)
+
+        # Request for upload ticket
+        if self.get_param('ticket') == '':
+            return self._create_ticket()
+
+        # Check ticket id and skip permissions check if it clears
+        ticket_id = self.get_param('ticket')
```
ambrussimon: Having URL params `'ticket'` _and_ `'job_ticket'` on the same endpoint seems very confusing. I would suggest renaming this one to `'upload_ticket'` unless there's a problem with that I don't see yet.

api/upload.py
```diff
@@ -204,14 +255,42 @@ def engine(self):
...
+                # If we don't have an origin with this request, use the ticket's origin
+                self.origin = ticket.get('origin')
+
+            file_fields = [
+                util.dotdict({
+                    'filename': ticket['filename']
+                })
+            ]
+
+            if level is not 'analysis':
```
ambrussimon: How can analysis job results be uploaded directly to storage (using a ticket)?

api/upload.py
```diff
@@ -204,14 +255,42 @@ def engine(self):
...
+        # Request for upload ticket
+        if self.get_param('ticket') == '':
+            return self._create_ticket()
+
+        # Check ticket id and skip permissions check if it clears
+        ticket_id = self.get_param('ticket')
+        if ticket_id:
+            ticket = self._check_ticket(ticket_id)
+            if not self.origin.get('id'):
+                # If we don't have an origin with this request, use the ticket's origin
```
nagem: In what situation would you expect us to not have an origin? Above we check to make sure the request has admin rights, so we only allow authenticated requests for this endpoint.

nagem:
I'm not sure what kind of Swagger coverage we have for the engine and reaper upload endpoints, but could you amend and add information about how to use the signed URL ticket method?

api/upload.py
```diff
@@ -166,6 +174,33 @@ def process_upload(request, strategy, container_type=None, id_=None, origin=None
...
+        ticket = util.upload_ticket(self.request.client_addr, self.origin, tempdir, filename, metadata)
+        return {'ticket': config.db.uploads.insert_one(ticket).inserted_id,
+                'upload_url': signed_url}
+
+    def _check_ticket(self, ticket_id):
+        ticket = config.db.uploads.find_one({'_id': ticket_id})
+        if not ticket:
+            self.abort(404, 'no such ticket')
+        if ticket['ip'] != self.request.client_addr:
+            self.abort(400, 'ticket not for this resource or source IP')
```
nagem: You could argue this is a 403 because ip is our rough permission check (other than ownership of the ticket).

api/upload.py
```diff
@@ -174,20 +209,36 @@ def upload(self, strategy):
             if not user:
                 self.abort(403, 'Uploading requires login')
 
+        if strategy in ['label', 'uid', 'uid-match', 'reaper']:
+            strategy = strategy.replace('-', '')
+            strategy = getattr(Strategy, strategy)
+        else:
+            self.abort(500, 'strategy {} not implemented'.format(strategy))
```
nagem: I don't think we need the 500 here because something not accounted for won't make it past the routing logic in `api.py`.

api/upload.py
```diff
@@ -174,20 +209,36 @@ def upload(self, strategy):
...
+        # Request for upload ticket
+        if self.get_param('ticket') == '':
+            return self._create_ticket()
+
+        # Check ticket id and skip permissions check if it clears
+        ticket_id = self.get_param('ticket')
+        if ticket_id:
+            ticket = self._check_ticket(ticket_id)
+            if not self.origin.get('id'):
+                # If we don't have an origin with this request, use the ticket's origin
```
nagem: Same question as before - I don't think it's possible to have an empty origin with an authenticated request.

PR #1103, opened 2018-03-19
Require all path characters to be ascii

nagem:
Downloads failed when container or file names contained Unicode characters. For now, strip all non ascii characters. Encoding in utf-8 is also an option that can be explored in the future if users would like to maintain their allowed unicode container labels in download paths. 

### Review Checklist


api/download.py
```diff
@@ -266,6 +265,9 @@ def _find_new_path(path, ids_of_paths, _id):
             path = container['uid']
         if not path and container.get('code'):
             path = container['code']
+
+        path = path.encode('ascii', 'ignore')
```
ambrussimon: I did the same in reaper lately, I think `errors='ignore'` is a bit easier on the eyes.

PR #1104, opened 2018-03-20
Analysis fix

ambrussimon:
@ehlertjd @nagem 

> We were doing some tests for the new analysis endpoints this afternoon and I realized that inputs from SDK analyses are getting saved as a list of file references rather than a copy of the file object. Also, the API doesn’t check to ensure the inputs exist in the system

I looked into both and couldn't recreate the behavior:
* Tests were already in place for checking whether [filerefs are turned into fileinfos](https://github.com/scitran/core/blob/master/tests/integration_tests/python/test_analyses.py#L123)
* Added new tests that show filerefs are also verified to [ensure they exist](https://github.com/scitran/core/compare/analysis-fix?expand=1#diff-126d08ae5967d12347692a35180c33e4R14) in the system
* My only thin guess is that the ["fallback to legacy"](https://github.com/scitran/core/blob/master/api/handlers/refererhandler.py#L75) mechanism is a bit loose and that legacy placer was triggered during SDK tests, which in turn only checked `anyOf`
* ^ Fixed by separating out legacy input schema
* Also found that analysis notes input schemas were pointing to `notes-list-output`, fixed

### Review Checklist

### DIFF
--- a/swagger/schemas/definitions/analysis.json
+++ b/swagger/schemas/definitions/analysis.json
@@ -46,8 +46,7 @@
 		"analysis-input-any": {
 			"anyOf": [
 				{"$ref":"#/definitions/analysis-input-adhoc"},
-				{"$ref":"#/definitions/analysis-input-job"},
-				{"$ref":"#/definitions/analysis-input-legacy"}
+				{"$ref":"#/definitions/analysis-input-job"}
 			],
 			"x-sdk-schema": {"$ref":"#/definitions/analysis-input-adhoc"}
 		},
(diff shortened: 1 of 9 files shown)


ambrussimon:
Please provide repro steps if the problem wasn't solved by this.

ehlertjd:
@ambrussimon Added an integration test to `test_analyses.py` demonstrating the problem.

ambrussimon:
@ehlertjd Perfect, thank you! Fixed.

swagger/schemas/definitions/analysis.json
```diff
@@ -8,7 +8,7 @@
 					"type": "array",
 					"items": {"$ref":"file.json#/definitions/file-reference"}
 				},
-				"notes":       {"$ref":"note.json#/definitions/notes-list-output"},
+				"notes":       {"$ref":"note.json#/definitions/notes-list-input"},
```
ehlertjd: Good catch

PR #1105, opened 2018-03-20
Fix ticket-based job state transition for analysis jobs

### LINKED ISSUE #1084: Analysis jobs do not transition to complete state when using completion tickets
Non-analysis jobs complete fine. Only analysis jobs are impacted.

ambrussimon:
Fixes #1084 

* Mostly copied relevant parts from EnginePlacer (DRY suggestions welcome)
* Added tests

### Review Checklist

### DIFF
--- a/api/placer.py
+++ b/api/placer.py
@@ -331,6 +331,8 @@ class EnginePlacer(Placer):
             job_ticket = JobTicket.get(self.context.get('job_ticket_id'))
             job = Job.get(job_ticket['job'])
             success = job_ticket['success']
+        elif self.context.get('job_id'):
+            job = Job.get(self.context.get('job_id'))
 
         if self.metadata is not None:
             bid = bson.ObjectId(self.id_)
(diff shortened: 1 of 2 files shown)


ryansanford:
@nagem @ambrussimon Please confirm if this adds any risk for endpoints in use, so we can decide whether to hold this merge for after the imminent release. Thanks.

ambrussimon:
@ryansanford Change scope is limited to `POST /upload/engine?level=analysis` and should be only additive in terms of functionality. Personally I think the risk is minimal.

api/placer.py
```diff
@@ -350,23 +352,11 @@ def finalize(self):
...
-            else:
-                Queue.mutate(job, {
-                    'state': 'failed',
-                    'profile': {
-                        'elapsed': job_ticket['elapsed']
-                    }
-                })
+            Queue.mutate(job, {'state': 'complete' if success else 'failed',
+                               'profile': {'elapsed': job_ticket['elapsed']}})
+            job = Job.get(job.id_)
```
nagem: This will raise an `APINotFoundException`, not return `None`. 

It looks like we call it above too so I'm curious why you're checking for `None` below. Race condition that it gets deleted between the `Queue.mutate` call and `save`?

ambrussimon: `Queue.mutate` does _not_ update the loaded job's attributes (except partially in this special case: https://github.com/scitran/core/blob/master/api/jobs/queue.py#L82)
Therefore calling `job.save()` once more (when setting `saved_files` below) would try to revert the state transition.
Was thinking of updating `mutate` instead, but wanted to limit the scope of the change and the possible errors introduced.

api/placer.py
```diff
@@ -743,27 +732,45 @@ def process_file_field(self, field, file_attrs):
...
+            if not job_ticket['success']:
+                file_attrs['from_failed_job'] = True
+
         file_attrs['created'] = file_attrs['modified']
         self.save_file(field)
         self.saved.append(file_attrs)
 
     def finalize(self):
-        # Search the sessions table for analysis, replace file field
-        if self.saved:
```
kofalt: Sanity check: you're sure we don't need the `self.saved` conditional? I'm not super familiar here.

kofalt (approved):
LGTM, one quick question.

PR #1108, opened 2018-03-23
Note move to flywheel-io/core

gsfr:


kofalt (approved):
I would have skipped the bold, but all good :+1:

